PR #1, opened 2023-05-15
More repo setup boilerplate

drevell:
Following the internal checklist doc.

The workflow .yml file is a copy of https://github.com/abcxyz/bet-in-a-box/blob/main/.github/workflows/test-unit.yml.


.github/workflows/lint-test.yml
```diff
@@ -0,0 +1,26 @@
+name: 'Lint and test'
+
+on:
+  push:
+    branches:
+      - 'main'
+  pull_request:
+    branches:
+      - 'main'
+  workflow_dispatch:
```
chloechien, line 10: If CI will guard every PR, curious the reason we still need workflow_dispatch here.

chloechien (approved):

.github/workflows/lint-test.yml
drevell: Not sure, but maybe somebody will want to lint a branch that's not part of a PR, and I think this will let them do that.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,132 @@
+# Contributor Covenant Code of Conduct
```
sethvargo, line 1: FYI, we don't need this because it's in this repo: https://github.com/abcxyz/.github. `.github` is a special repo where we can host things like contribution guides, codes of conduct, and issue templates. I would suggest removing from here, since it's inherited from .github (and centrally managed).

PR #5, opened 2023-05-24
YAML model and validation for spec file

drevell:


templates/model/spec.go
```diff
@@ -0,0 +1,285 @@
...
+	Kind       String `yaml:"kind"`
+
+	Desc   String   `yaml:"desc"`
+	Inputs []*Input `yaml:"inputs"`
+	Steps  []*Step  `yaml:"steps"`
+}
+
+// UnmarshalYAML implements yaml.Unmarshaler.
+func (s *Spec) UnmarshalYAML(n *yaml.Node) error {
+	if err := extraFields(n, []string{"apiVersion", "kind", "desc", "inputs", "steps"}); err != nil {
```
sethvargo, line 77: Should these keys be constants or should the list be a variable? It's not entire clear where they are coming from.

templates/model/spec.go
```diff
@@ -0,0 +1,285 @@
...
+		return err
+	}
+
+	type shadowType Spec
+	shadow := &shadowType{} // unmarshal into a type that doesn't have UnmarshalYAML
```
sethvargo, line 82: I see the comment, but I don't fully understand _why_ we need to do this - enlighten me?

templates/model/spec.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	*s = Spec(*shadow)
+	s.Pos = yamlPos(n)
+	return nil
+}
+
+// Validate implements Validator.
+func (s *Spec) Validate() error {
+	return errors.Join(
```
sethvargo, line 94: I really love how `errors.Join` ignores nil errors - it makes these kinds of things so much nicer.

templates/model/spec_test.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	for _, tc := range cases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			got := &Spec{}
+			dec := NewDecoder(strings.NewReader(tc.in))
+			if err := dec.Decode(got); tc.wantUnmarshalErr != "" || err != nil {
+				if diff := testutil.DiffErrString(err, tc.wantUnmarshalErr); diff != "" {
```
sethvargo, line 125: I might be misunderstanding the branching, but I think this misses a case where we want to assert where the marshall error is expected by the error is nil? I think the differrstring helper is smart enough to handle this if you just pass the values directly:

```go
err := dec.Decode(got)
if diff := testutil.DiffErrString(err, tc.wantUnmarshalErr); diff != nil {
  // ...
}
```

(throughout)

templates/model/validate.go
```diff
@@ -0,0 +1,99 @@
...
+
+func notZero[T comparable](pos ConfigPos, x valWithPos[T], fieldName string) error {
+	var zero T
+	if x.Val == zero {
+		return pos.AnnotateErr(fmt.Errorf("field %q is required", fieldName))
+	}
+	return nil
+}
+
+func nonemptySlice[T any](pos ConfigPos, s []T, fieldName string) error {
```
sethvargo, line 34: Nit: 
```suggestion
func nonEmptySlice[T any](pos ConfigPos, s []T, fieldName string) error {
```

templates/model/validate.go
```diff
@@ -0,0 +1,99 @@
...
+
+// Fail if any unexpected fields are seen. The input must be a mapping/object; anything else will
+// result in an error.
+// This is a workaround for the brokenness of KnownFields in the upstream yaml lib
+// (https://github.com/go-yaml/yaml/issues/460).
+func extraFields(n *yaml.Node, knownFields []string) error {
+	if n.Kind != yaml.MappingNode {
+		return fmt.Errorf("got yaml node of kind %d, expected %d", n.Kind, yaml.MappingNode)
+	}
+	m := map[string]any{}
```
sethvargo, line 49: I don't know if we can guarantee that the response is a `map[string]any`? It could be a `[]any` too.

templates/model/validate.go
```diff
@@ -0,0 +1,99 @@
...
+	if err := n.Decode(m); err != nil {
+		return err //nolint:wrapcheck
+	}
+
+	var unknownField string
+	for k := range m {
+		if !slices.Contains(knownFields, k) {
+			unknownField = k
```
sethvargo, line 57:
```suggestion
			unknownField = k
			break
```

templates/model/validate.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	return pos.AnnotateErr(fmt.Errorf("unknown field name %q", unknownField))
+}
+
+type validator interface {
+	Validate() error
+}
+
+func validateIfNotNil(v validator) error {
```
sethvargo, line 83: Even though this is private, can you document the use case?

templates/model/validate.go
```diff
@@ -0,0 +1,99 @@
...
+		return nil
+	}
+	return v.Validate() //nolint:wrapcheck
+}
+
+func validateEach[T validator](s []T) error {
+	var out []error
+	for _, v := range s {
+		if err := v.Validate(); err != nil {
+			out = append(out, err)
```
sethvargo, line 94: Nit: why not use `merr = errors.Join(merr, err)` instead? It avoids double allocing the `[]error`.

templates/model/pos.go
```diff
@@ -0,0 +1,56 @@
...
+func yamlPos(n *yaml.Node) ConfigPos {
+	return ConfigPos{
+		Line:   n.Line,
+		Column: n.Column,
+	}
+}
+
+// AnnotateErr prepends the config file location of a parsed value to an error. If the input err is
+// nil, then nil is returned.
+func (c ConfigPos) AnnotateErr(err error) error {
```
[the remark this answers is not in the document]
drevell: Done.

templates/model/spec.go
```diff
@@ -0,0 +1,285 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package model contains the structs for unmarshaled YAML files.
+//
+//nolint:wrapcheck // We don't want to excessively wrap errors, like "yaml error: yaml error: ..."
+package model
+
+// Notes for maintainers explaining the confusing stuff in this file:
```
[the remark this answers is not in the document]
drevell: 👍👍

templates/model/spec.go
drevell: Done, I made it a list with a descriptive name, which I think clarifies it.

templates/model/spec.go
drevell: Rewrote to clarify, lmk.

templates/model/spec.go
drevell: Yeah, it's like an "error monad" or something. Kinda DSL-ish.

templates/model/validate.go
drevell: Done

templates/model/validate.go
drevell: Hmm indeed. Done.

templates/model/validate.go
drevell: The check for MappingNode three lines up actually guarantees this.

templates/model/validate.go
drevell: Done.

templates/model/validate.go
drevell: Sure, done.

templates/model/spec_test.go
drevell: But I want to skip the rest of the rest if there was an error, or if there was an expected error, or both. I don't want to try to Validate() if unmarshaling failed, regardless of whether hte failure was expected.

yolocs (approved):

templates/model/spec_test.go
yolocs: What I tend to do is to return on err. I find it slightly cleaner

```go
err := dec.Decode(got)
if diff := testutil.DiffErrString(err, tc.wantUnmarshalErr); diff != "" {
  // ...
}
if err != nil {
  return
}
```

templates/model/pos.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// AnnotateErr prepends the config file location of a parsed value to an error. If the input err is
+// nil, then nil is returned.
+func (c *ConfigPos) AnnotateErr(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	pos := "(position unknown)" // This can happen when field values are defaults, rather that coming from the config file
```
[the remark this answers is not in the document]
drevell: ctrl-f "default in spec.go

templates/model/spec.go
```diff
@@ -0,0 +1,285 @@
...
+// Q. What's up with the weird "shadow" unmarshalling pattern (ctrl-f "shadow")
+// A. We often want to "unmarshal all the fields of my struct, but also run
+//    our own logic before and after unmarshaling. To do this, we have to create
+//    a separate type that *doesn't* implement yaml.Unmarshaler, and unmarshal
+//    into that first. If we're inside UnmarshalYAML() and we try to unmarshal
+//    into the receiver, there will be infinite recursion into UnmarshalYAML().
+//    See e.g. https://github.com/go-yaml/yaml/issues/107#issuecomment-524681153.
+//
+// Q. Why is validation done as a separate pass instead of in UnmarshalYAML()?
+// A. Because there's a very specific edge case that we need to avoid.
```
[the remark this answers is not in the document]
drevell: That's not actually relevant to the point being made here. In this paragraph, I'm explaining why calling Validate() from UnmarshalYAML() doesn't work. Cross-field validation can be implemented with either approach, so it doesn't affect the decision being described bere.

templates/model/validate.go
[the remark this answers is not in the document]
drevell: Hmm... that actually seems less clear to me. How about `validateUnlessNil`? Code updated.

templates/model/spec_test.go
drevell: That sounds good to me. Switched to Chen's suggestion.

PR #6, opened 2023-05-24
Make the template source into a positional argument

drevell:
This has two benefits:
 - Avoids the awkwardness of a "required flag"
 - Makes the command easier to type by skipping the "-s"

### DIFF
--- a/templates/commands/render.go
+++ b/templates/commands/render.go
@@ -18,6 +18,7 @@ package commands
 import (
 	"context"
 	"errors"
+	"flag"
 	"fmt"
 	"io/fs"
 	"os"
@@ -30,7 +31,7 @@ type Render struct {
 
 	testFS fs.StatFS
 
-	flagSource         string
+	source             string
 	flagSpec           string
 	flagDest           string
 	flagGitProtocol    string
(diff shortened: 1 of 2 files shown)


yolocs (approved):
I like it

chloechien (approved):

templates/commands/render.go
```diff
@@ -47,28 +48,29 @@ func (r *Render) Desc() string {
...
-		Aliases: []string{"s"},
-		Example: "helloworld@v1",
-		Target:  &r.flagSource,
-		Usage: `Required. The location of the template to be instantiated. Many forms are accepted. ` +
-			`"helloworld@v1" means "github.com/abcxyz/helloworld repo at revision v1; this is for a template owned by abcxyz. ` +
-			`"myorg/helloworld@v1" means github.com/myorg/helloworld repo at revision v1; this is for a template not owned by abcxyz but still on GitHub. ` +
-			`"mygithost.com/mygitrepo/helloworld@v1" is for a template in a remote git repo but not owned by abcxyz and not on GitHub. ` +
-			`"mylocaltarball.tgz" is for a template not in git but present on the local filesystem. ` +
-			`"http://example.com/myremotetarball.tgz" os for a non-Git template in a remote tarball.`,
-	})
```
sethvargo, line 71: If you upgrade to the latest CLI package and add a `PredictArgs()` function, you can get tab completion: https://github.com/abcxyz/pkg/blob/bd3d55380569e64a7c5a3acf2ff1bd81f502f8dd/cli/command_completions_doc_test.go#L47-L51

templates/commands/render_test.go
```diff
@@ -86,7 +87,7 @@ func TestParseFlags(t *testing.T) {
 			t.Parallel()
 			r := &Render{}
 			err := r.parseFlags(tc.args)
-			if tc.wantErr != "" {
+			if err != nil || tc.wantErr != "" {
 				if diff := testutil.DiffErrString(err, tc.wantErr); diff != "" {
 					t.Fatal(diff)
 				}
```
sethvargo, line 93: Sorry, I still don't understand why we need the preflight check on `err` or `tc.wantErr`. DiffErrString should be taking care of all of that.

drevell: Consider the case where there's an err, and tc.wantErr matches it. Under your suggestion, DiffErrString would return "" and the test would continue on. It would call cmp.Diff to check the values of the flags, checks which we don't want (in this case).

Now you might be saying "yes we DO want to continue because we should then assert that the Render object is in a cleanly failed state, like just assert that it's a zero object or nil or something." But no, flag parsing will have left the the flags in a half-populated state that we should not predict or rely on.

sethvargo: Ah I see. I think it's usually more obvious to do something like:

```go
if diff := testutils.DiffErrString(err, tc.wantErr); err != nil {
  t.Fatal(diff) // which will stop tests
}
```

or

```go
if diff := testutils.DiffErrString(err, tc.wantErr); err != nil {
  t.Error(diff)
}
if err != nil {
  return
}
```

drevell: The first one looks not quite right (it will fail the test if there's an error, regardless of whether it was expected). I'll adopt the second one though.

PR #7, opened 2023-05-25
Add more action types

drevell:
This PR adds the remaining action types from the design doc: `regex_replace`, `string_replace`, and `go_template`.

### DIFF
--- a/templates/model/primitive.go
+++ b/templates/model/primitive.go
@@ -25,6 +25,9 @@ type String = valWithPos[string]
 // Bool represents a boolean field in a model, together with its location in the input file.
 type Bool = valWithPos[bool]
 
+// Bool represents an integer field in a model, together with its location in the input file.
+type Int = valWithPos[int]
+
 // valWithPos unmarshals a type T from YAML, and adds on the location in the YAML doc that it came
 // from. This allows helpful error messages.
 type valWithPos[T any] struct {
--- a/templates/model/primitive.go
+++ b/templates/model/primitive.go
@@ -25,7 +25,7 @@ type String = valWithPos[string]
 // Bool represents a boolean field in a model, together with its location in the input file.
 type Bool = valWithPos[bool]
 
-// Bool represents an integer field in a model, together with its location in the input file.
+// Int represents an integer field in a model, together with its location in the input file.
 type Int = valWithPos[int]
 
 // valWithPos unmarshals a type T from YAML, and adds on the location in the YAML doc that it came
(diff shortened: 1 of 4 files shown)


templates/model/spec.go
```diff
@@ -215,10 +225,9 @@ func (s *Step) Validate() error {
 		notZero(s.Pos, s.Desc, "desc"),
 		validateUnlessNil(s.Print),
 		validateUnlessNil(s.Include),
-		// TODO: add more action types:
-		// validateIfNotNil(s.RegexReplace),
-		// validateIfNotNil(s.StringReplace),
-		// validateIfNotNil(s.GoTemplate),
+		validateUnlessNil(s.RegexReplace),
+		validateUnlessNil(s.StringReplace),
+		validateUnlessNil(s.GoTemplate),
```
yolocs, line 230: Is it worth validating only one of action can be specified? Or it isn't?

templates/model/spec.go
```diff
@@ -288,3 +297,118 @@ func (i *Include) Validate() error {
...
+
+// RegexReplace is an action that replaces a regex match (or a subgroup of it) with a
+// template expression.
+type RegexReplace struct {
+	// Pos is the YAML file location where this object started.
+	Pos *ConfigPos `yaml:"-"`
+
+	Regex    String `yaml:"regex"`
+	Subgroup Int    `yaml:"subgroup"`
+	With     String `yaml:"with"`
```
yolocs, line 309: Comments (throughout)? Given this file is the source of truth of the spec.

Also, should we add go example later?

templates/model/primitive.go
```diff
@@ -25,6 +25,9 @@ type String = valWithPos[string]
 // Bool represents a boolean field in a model, together with its location in the input file.
 type Bool = valWithPos[bool]
 
+// Bool represents an integer field in a model, together with its location in the input file.
```
yolocs, line 28:
```suggestion
// Int represents an integer field in a model, together with its location in the input file.
```

yolocs (approved):

templates/model/spec.go
drevell: That's guaranteed by Spec.UnmarshalYAML(), take a look there. We choose a single field to parse into, then call Decode() just once, so only one of them can be set.

templates/model/primitive.go
drevell: Oops, done.

templates/model/spec.go
yolocs: You mean `Step.UnmarshalYAML()`?  I see it now.

templates/model/spec.go
drevell: Yeah, we should do that. Let's come back and do documentation once we get something running.

PR #8, opened 2023-06-01
integrate with go-getter for fetching templates

drevell:


templates/commands/render.go
```diff
@@ -17,19 +17,27 @@ package commands
...
 	"io/fs"
 	"os"
+	"path/filepath"
 
 	"github.com/abcxyz/pkg/cli"
+	"github.com/abcxyz/pkg/logging"
+	"github.com/hashicorp/go-getter/v2"
 )
 
+var logger = logging.NewFromEnv("ABC_")
```
sethvargo, line 34: I'm not a fan of making this global, since overriding it for tests is then harder. Can we embed it in the command or context instead?

drevell: Sure, moved to a field on the Render struct.

templates/commands/render.go
```diff
@@ -141,25 +161,155 @@ func (r *Render) parseFlags(args []string) error {
...
+	if gg == nil {
+		gg = &getter.Client{
+			Getters:       getter.Getters,
+			Decompressors: getter.Decompressors,
+		}
+	}
+
+	if r.logger == nil {
+		r.logger = logging.NewFromEnv("ABC_")
+	}
```
yolocs, line 204: I don't think you need to make the logger a field. Once you have a logger, you can put it in context with https://github.com/abcxyz/pkg/blob/main/logging/logger.go#L86. During test, when you call `readRun`, you can put a test logger (https://github.com/abcxyz/pkg/blob/main/logging/logger.go#L109) in the context instead.

templates/commands/render.go
```diff
@@ -141,25 +161,155 @@ func (r *Render) parseFlags(args []string) error {
...
+	var merr error
+	for _, p := range tempDirs {
+		merr = errors.Join(merr, fs.RemoveAll(p))
+	}
+	return merr
+}
+
+// Generate the name for a temporary directory, without creating it. debugID is
+// an optional name that can be included to help understand which directory is
+// which during debugging.
```
yolocs, line 289: Not sure I understand the use of debug id. It seems it's a fixed string in real run and not used in test?

templates/commands/render_test.go
```diff
@@ -155,11 +160,177 @@ func TestDestOK(t *testing.T) {
...
+	err        error
+}
+
+func (f *fakeGetter) Get(ctx context.Context, req *getter.Request) (*getter.GetResult, error) {
+	if f.err != nil {
+		return nil, f.err
+	}
+	if req.Src != f.wantSource {
+		return nil, fmt.Errorf("got template source %s but expected %s", req.Src, f.wantSource)
+	}
```
yolocs, line 324: I'd prefer comparing got/want in test instead since it's more explicit.

templates/commands/render.go
sethvargo: I think this was my fault. I misled Dave with my comment "Can we embed it in the command or context instead?".

templates/commands/render.go
drevell: Now moved to context per the other thread.

templates/commands/render.go
drevell: Added to the context.

templates/commands/render.go
drevell: I renamed the variable and rewrote the comment to make this more clear, PTAL.

templates/commands/render_test.go
drevell: I think you're looking for the word "expected" to be "want" instead? I made that change.

sethvargo (approved):

PR #9, opened 2023-06-02
Fix fake assertion style per Chen's offline feedback

drevell:


yolocs (approved):

templates/commands/render_test.go
```diff
@@ -232,6 +232,10 @@ func TestRealRun(t *testing.T) {
 				t.Error(diff)
 			}
 
+			if fg.gotSource != r.source {
+				t.Errorf("fake getter got template source %s but wanted %s", fg.gotSource, r.source)
+			}
```
sethvargo, line 237: I _think_ Chen might have wanted this:

```suggestion
			if got, want := fg.gotSource, r.source {
				t.Errorf("fake getter got template source %s but wanted %s", got want)
			}
```

PR #10, opened 2023-06-05
Implement the "include" and "print" actions

drevell:
The "include" action copies files from the template into the scratch directory. This is analogous to a docker "COPY".

The "print" command just prints to stdout.

In both of these actions, template variable substitution is allowed.

This PR is big, but it's mostly tests.

Coming next: implementing more action types and committing the scratch dir to the output directory.

### DIFF
--- a/templates/commands/render.go
+++ b/templates/commands/render.go
@@ -42,6 +42,9 @@ const (
 	// make them identifiable.
 	templateDirNamePart = "template-copy"
 	scratchDirNamePart  = "scratch"
+
+	// Permission bits: rwx------
+	mkdirPerms = 0o700
 )
 
 type Render struct {
@@ -287,7 +290,7 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
 	if err != nil {
 		return err
 	}
-	if err := rp.fs.MkdirAll(scratchDir, 0o700); err != nil {
+	if err := rp.fs.MkdirAll(scratchDir, mkdirPerms); err != nil {
 		return fmt.Errorf("failed to create scratch directory: MkdirAll(): %w", err)
 	}
 
--- a/templates/commands/render.go
+++ b/templates/commands/render.go
@@ -12,8 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package render implements the "templates render" subcommand for installing a template.
+// Package commands implements the template-related subcommands.
 package commands
+
+// This file implements the "templates render" subcommand for installing a template.
 
 import (
 	"context"
--- a/templates/commands/render.go
+++ b/templates/commands/render.go
@@ -411,18 +411,6 @@ func copyRecursive(pos *model.ConfigPos, from, to string, rfs renderFS) error {
 	})
 }
 
-func actionRegexReplace(ctx context.Context, p *model.RegexReplace, sp *stepParams) error {
-	return fmt.Errorf("not implemented")
-}
-
-func actionStringReplace(ctx context.Context, p *model.StringReplace, sp *stepParams) error {
-	return fmt.Errorf("not implemented")
-}
-
-func actionGoTemplate(ctx context.Context, p *model.GoTemplate, sp *stepParams) error {
-	return fmt.Errorf("not implemented")
-}
-
 func loadSpecFile(fs renderFS, templateDir, flagSpec string) (*model.Spec, error) {
 	f, err := fs.Open(filepath.Join(templateDir, flagSpec))
 	if err != nil {
--- a/templates/commands/render.go
+++ b/templates/commands/render.go
@@ -41,6 +41,12 @@ const (
 	// make them identifiable.
 	templateDirNamePart = "template-copy"
 	scratchDirNamePart  = "scratch"
+
+	// Bitmasks for bitwise logic with read/write/execute permission in a
+	// FileMode.
+	ownerRead  = 0o400
+	ownerWrite = 0o200
+	ownerExec  = 0o100
 )
 
 type Render struct {
@@ -185,11 +191,11 @@ func (r *Render) parseFlags(args []string) error {
 type realFS struct{}
 
 func (r *realFS) Mkdir(name string, perm os.FileMode) error {
-	return os.Mkdir(name, perm)
+	return os.Mkdir(name, perm) //nolint:wrapcheck
 }
 
 func (r *realFS) MkdirAll(name string, perm os.FileMode) error {
-	return os.MkdirAll(name, perm)
+	return os.MkdirAll(name, perm) //nolint:wrapcheck
 }
 
 func (r *realFS) Open(name string) (fs.File, error) {
@@ -197,7 +203,7 @@ func (r *realFS) Open(name string) (fs.File, error) {
 }
 
 func (r *realFS) ReadFile(name string) ([]byte, error) {
-	return os.ReadFile(name)
+	return os.ReadFile(name) //nolint:wrapcheck
 }
 
 func (r *realFS) RemoveAll(name string) error {
@@ -209,7 +215,7 @@ func (r *realFS) Stat(name string) (fs.FileInfo, error) {
 }
 
 func (r *realFS) WriteFile(name string, data []byte, perm os.FileMode) error {
-	return os.WriteFile(name, data, perm)
+	return os.WriteFile(name, data, perm) //nolint:wrapcheck
 }
 
 func (r *Render) Run(ctx context.Context, args []string) error {
@@ -280,7 +286,7 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
 		return err
 	}
 
-	scratchDir, err := rp.tempDirNamer("scratch")
+	scratchDir, err := rp.tempDirNamer(scratchDirNamePart)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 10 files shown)


templates/commands/render.go
```diff
@@ -235,24 +275,237 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+	inputs      map[string]string
+	scratchDir  string
+	stdout      io.Writer
+	templateDir string
+}
+
+func executeOneStep(ctx context.Context, step *model.Step, sp *stepParams) error {
+	switch {
+	case step.Print != nil:
+		return actionPrint(ctx, step.Print, sp)
```
yolocs, line 324: This is my personal taste. I find it'd be a little less cluttered if we have actions in separate files, e.g. `action_print.go` and `action_print_test.go`

templates/commands/render.go
```diff
@@ -47,10 +59,17 @@ type Render struct {
...
+//
+// We can't use os.DirFS or fs.StatFS because they lack some methods we need. So
+// we created our own interface.
 type renderFS interface {
 	fs.StatFS
+	Mkdir(name string, perm os.FileMode) error
+	MkdirAll(string, os.FileMode) error
+	ReadFile(string) ([]byte, error)
 	RemoveAll(string) error
+	WriteFile(name string, data []byte, perm os.FileMode) error
```
sethvargo, line 72: Nit: I find it weird that some of the params are named and some are not. Since this is an interface, I would suggest we name them or add comments (or both).

templates/commands/render.go
```diff
@@ -235,24 +275,237 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+	if err != nil {
+		return err
+	}
+
+	scratchDir, err := rp.tempDirNamer("scratch")
+	if err != nil {
+		return err
+	}
+	if err := rp.fs.MkdirAll(scratchDir, 0o700); err != nil {
+		return fmt.Errorf("MkdirAll(): %w", err)
```
sethvargo, line 288: "failed to create scratch directory" or something more human

templates/commands/render.go
sethvargo: I agree with Chen FWIW. It makes parsing and finding things a lot easier.

templates/commands/render.go
```diff
@@ -235,24 +275,237 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+	case step.Include != nil:
+		return actionInclude(ctx, step.Include, sp)
+	case step.RegexReplace != nil:
+		return actionRegexReplace(ctx, step.RegexReplace, sp)
+	case step.StringReplace != nil:
+		return actionStringReplace(ctx, step.StringReplace, sp)
+	case step.GoTemplate != nil:
+		return actionGoTemplate(ctx, step.GoTemplate, sp)
+	default:
+		return fmt.Errorf("internal error: unknown step action type %q", step.Action.Val)
```
sethvargo, line 334: Is this an _internal_ error? Or is this is a user error?

templates/commands/render.go
```diff
@@ -235,24 +275,237 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+	}
+}
+
+func errWithPos(pos *model.ConfigPos, fmtStr string, args ...any) error {
```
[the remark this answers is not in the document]
sethvargo: I wonder if we want to dangle off `ConfigPos`:

```go
pos.NewError("expected %s", thing)
```

templates/commands/render.go
```diff
@@ -235,24 +275,237 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+	err := fmt.Errorf(fmtStr, args...)
+	lineStr := fmt.Sprintf("line %d", pos.Line)
+	if pos.Line == 0 {
+		lineStr = "(unknown line number)"
+	}
+	return fmt.Errorf("failed executing template spec file at %s: %w", lineStr, err)
+}
+
+func parseGoTmpl(tpl string) (*template.Template, error) {
```
sethvargo, line 347: What is the value of this helper?

templates/commands/render.go
```diff
@@ -235,24 +275,237 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+	return template.New("").Option("missingkey=error").Parse(tpl)
+}
+
+func parseAndExecuteGoTmpl(m model.String, inputs map[string]string) (string, error) {
+	goTmpl, err := parseGoTmpl(m.Val)
```
sethvargo, line 352: Does it make sense to cache any of these templates so we don't keep reparsing?

templates/commands/render.go
```diff
@@ -235,24 +275,237 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+	if err != nil {
+		return "", errWithPos(m.Pos, `error compiling as go-template: %w`, err)
+	}
+
+	sb := strings.Builder{}
```
sethvargo, line 357: Nit: 

```go
var sb strings.Builder
```

templates/commands/render.go
```diff
@@ -235,24 +275,237 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+		lineStr = "(unknown line number)"
+	}
+	return fmt.Errorf("failed executing template spec file at %s: %w", lineStr, err)
+}
+
+func parseGoTmpl(tpl string) (*template.Template, error) {
+	return template.New("").Option("missingkey=error").Parse(tpl)
+}
+
+func parseAndExecuteGoTmpl(m model.String, inputs map[string]string) (string, error) {
```
sethvargo, line 351: It feels a little weird that this returns a string instead of accepting an `io.Writer`. What do you think?

templates/commands/render.go
```diff
@@ -235,24 +275,237 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+
+		if err := safeRelPath(p.Val); err != nil {
+			return errWithPos(p.Pos, "invalid path: %w", err)
+		}
+
+		absSrc := filepath.Join(sp.templateDir, walkRelPath)
+		absDst := filepath.Join(sp.scratchDir, walkRelPath)
+
+		if _, err := sp.fs.Stat(absSrc); err != nil {
+			if errors.Is(err, os.ErrNotExist) {
```
sethvargo, line 399: What if it's a different error? I think we're missing a different return?

templates/commands/render.go
```diff
@@ -235,24 +275,237 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+		// set conservatively to either 600 (u=rw) or 700 (u=rwx), depending on
+		// the owner-executable permission bit of the source file. These
+		// permissions will be further limited by the user's umask (see `man
+		// umask` if unfamiliar), so the actual permissions on disk might be
+		// more restrictive than we ask for.
+		info, err := rfs.Stat(path)
+		if err != nil {
+			return fmt.Errorf("Stat(): %w", err)
+		}
+		ownerExecuteBit := info.Mode() & 0o100
```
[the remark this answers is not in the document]
sethvargo: Bit banging!

templates/commands/render.go
```diff
@@ -235,24 +275,237 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+		dstPerms := 0o600 | ownerExecuteBit
+
+		if err := rfs.WriteFile(dst, buf, dstPerms); err != nil {
```
sethvargo, line 461: It can be a follow-up, but I think you're going to want an atomic write here. I've done this before, it's not fun: https://github.com/hashicorp/consul-template/blob/e97a929318455f1ff0d11360a746412d633a004c/renderer/renderer.go#L117-L214

templates/commands/render.go
sethvargo, line 458: This locks us to Linux-only...

templates/commands/render.go
```diff
@@ -235,24 +275,237 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+	return fmt.Errorf("not implemented")
+}
+
+func loadSpecFile(fs renderFS, templateDir, flagSpec string) (*model.Spec, error) {
+	buf, err := fs.ReadFile(filepath.Join(templateDir, flagSpec))
+	if err != nil {
+		return nil, fmt.Errorf("error reading template spec: ReadFile(): %w", err)
+	}
+
+	decoder := model.NewDecoder(bytes.NewReader(buf))
```
sethvargo, line 487:
```go
f, err := fs.Open(filepath.Join(templateDir, flagSpec))
if err != nil {
  // ...
}
defer f.Close()

decoder := model.NewDecoder(f)
```

? Avoids reading the entire thing into a buffer.

templates/commands/render.go
```diff
@@ -235,24 +275,237 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+	spec := model.Spec{}
```
sethvargo, line 488: Nit: 

```go
var spec model.Spec
```

templates/commands/render.go
```diff
@@ -322,3 +575,13 @@ func destOK(fs fs.StatFS, dest string) error {
 
 	return nil
 }
+
+func safeRelPath(p string) error {
```
sethvargo, line 579: I'd like to see some unit tests for this please.

templates/commands/render.go
```diff
@@ -235,24 +275,237 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
 		return err
 	}
 
-	_ = templateDir // TODO: add template rendering logic
+	spec, err := loadSpecFile(rp.fs, templateDir, r.flagSpec)
+	if err != nil {
+		return err
+	}
+
+	scratchDir, err := rp.tempDirNamer("scratch")
```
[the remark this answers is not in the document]
drevell: Thanks for catching that.

templates/commands/render.go
drevell: Sure, done.

templates/commands/render.go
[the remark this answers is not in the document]
drevell: Makes sense. Done.

templates/commands/render.go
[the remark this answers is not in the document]
drevell: It's a bitwise "and" to find the value of the "owner execute" bit in the file permissions. I rewrote this a bit to try to make it more clear.

templates/commands/render_test.go
```diff
@@ -236,30 +310,396 @@ func TestRealRun(t *testing.T) {
 				t.Errorf("fake getter got template source %s but wanted %s", fg.gotSource, r.source)
 			}
 
-			wantFiles := tc.templateContents
-			if !tc.keepTempDirs {
-				wantFiles = nil
+			if diff := cmp.Diff(stdoutBuf.String(), tc.wantStdout); diff != "" {
```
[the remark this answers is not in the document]
drevell: I'd rather not; I want to print a diff rather than make the user diff it with their eyeballs.

templates/commands/render.go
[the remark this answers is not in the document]
drevell: > I wonder if we want to dangle off `ConfigPos`:
> 
> ```go
> pos.NewError("expected %s", thing)
> ```

I'd rather not. Sometimes we want a nil or zero configpos, and it would be weird and awkward to have a method on a nil receiver.

templates/commands/render.go
drevell: Added a comment and removed the unncecessary param names.

templates/commands/render.go
drevell: It's an internal error. If the user screwed up, we would have caught it in the parsing step. This error means that the parser knows about an action type but the renderer doesn't. So it will "never" happen.

templates/commands/render.go
drevell: I'm thinking maybe this whole thing was overengineered, and we should just let the umask be in charge of permissions. WDYT?

sethvargo: I am agreeable

templates/commands/render.go
drevell: Done.

templates/commands/render.go
drevell: Yeah, that is better. Done.

templates/commands/render.go
drevell: SGTM, I think this can wait. Added a TODO.

templates/commands/render.go
drevell: It avoids this appear in several places: `template.New("").Option("missingkey=error")`

templates/commands/render.go
drevell: Done.

templates/commands/render.go
drevell: Done, PTAL

templates/commands/render.go
drevell: Added tests.

templates/commands/render.go
drevell: That will come soon. In the current version of the code, there's very little template reuse. Once we add the regex_replace action we'll want to cache both parsed templates and parsed regexes, since they'll be reused in a loop.

This exists in my prototype, it's just not time yet for this code.

templates/commands/render.go
drevell: Eh, I think that's a little worse. It's going to be in memory regardless, and that would force the caller to go through the boilerplate of declaring a Buffer and calling String() on it, to no end.

templates/commands/render.go
```diff
@@ -235,24 +276,182 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+		if err != nil {
+			return model.ErrWithPos(pos, "ReadFile(): %w", err) //nolint:wrapcheck
+		}
+
+		info, err := rfs.Stat(path)
+		if err != nil {
+			return fmt.Errorf("Stat(): %w", err)
+		}
+
+		// TODO: this arguable should be an atomic write so we don't end up with
```
yolocs, line 401: nit: if we won't get to this soon, create an issue.

templates/commands/render.go
```diff
@@ -235,24 +276,182 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+	return fmt.Errorf("not implemented")
+}
+
+func actionStringReplace(ctx context.Context, p *model.StringReplace, sp *stepParams) error {
+	return fmt.Errorf("not implemented")
+}
+
+func actionGoTemplate(ctx context.Context, p *model.GoTemplate, sp *stepParams) error {
+	return fmt.Errorf("not implemented")
+}
```
yolocs, line 424: nit: might as well creating separate files for them?

yolocs (approved):

sethvargo (dismissed):

templates/commands/render.go
drevell: Done.

templates/commands/render.go
drevell: Right now I have a private gdoc with a list of things to address. As soon as I finish porting the prototype repo to this repo, I'll create a GitHub issue for each item in that list. I'll remove this TODO comment since it's tracked elsewhere.

templates/commands/render.go
```diff
@@ -235,24 +278,167 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+	spec, err := loadSpecFile(rp.fs, templateDir, r.flagSpec)
+	if err != nil {
+		return err
+	}
 
+	scratchDir, err := rp.tempDirNamer(scratchDirNamePart)
+	if err != nil {
+		return err
+	}
+	if err := rp.fs.MkdirAll(scratchDir, 0o700); err != nil {
```
chloechien, line 290: Wondering should file permissions also be constant?

chloechien (approved):

templates/commands/render.go
drevell: Done.

templates/commands/render.go
drevell: I also noticed and fixed this :)

PR #11, opened 2023-06-06
More detailed logic for when overwriting is OK

drevell:
The logic in this PR is to use `overwrite=true` when writing to the scratch directory. This allows multiple include actions to touch the same set of files, if the template author wishes.

This PR also improves error messages in some edge cases.

This PR also forbids overwriting files with directories and vice versa, because that would be confusing for both template developers and abc code maintainers.

In an upcoming PR, we will set `overwrite=$(value_of_--force_overwrite_flag)` when writing to the final output directory.

### DIFF
--- a/templates/commands/render.go
+++ b/templates/commands/render.go
@@ -362,7 +362,7 @@ func parseAndExecuteGoTmpl(m model.String, inputs map[string]string) (string, er
 }
 
 // "from" may be a file or directory. "pos" is only used for error messages.
-func copyRecursive(pos *model.ConfigPos, from, to string, rfs renderFS) error {
+func copyRecursive(pos *model.ConfigPos, from, to string, rfs renderFS, overwrite bool) error {
 	return fs.WalkDir(rfs, from, func(path string, d fs.DirEntry, err error) error { //nolint:wrapcheck
 		if err != nil {
 			return err // There was some filesystem error. Give up.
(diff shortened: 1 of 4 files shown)


yolocs (commented):
From the description, it sounds like we'd default to not allow overwrite? I'm not clear when we want to preserve (the content in) the scratch folder.

drevell:
> From the description, it sounds like we'd default to not allow overwrite? I'm not clear when we want to preserve (the content in) the scratch folder.

We never want to preserve the content of the scratch folder. Therefore, we use overwrite=true in actionInclude(). In the future, we'll want to use overwrite=$(value_of_overwrite_flag) when writing to the final destination directory (not the scratch directory). I'll update the PR description to try to make this more clear.

templates/commands/render.go
```diff
@@ -391,6 +391,15 @@ func copyRecursive(pos *model.ConfigPos, from, to string, rfs renderFS) error {
 		if !d.IsDir() {
 			dirToCreate = filepath.Dir(dst)
 		}
+
+		dirInfo, err := rfs.Stat(dirToCreate)
+		if err != nil {
+			if !os.IsNotExist(err) {
+				return model.ErrWithPos(pos, "Stat(): %w", err) //nolint:wrapcheck
+			}
+		} else if !dirInfo.Mode().IsDir() {
```
sethvargo, line 400: Can `.Mode`() ever return nil?

sethvargo (approved):

templates/commands/render_action_include.go
```diff
@@ -46,7 +46,12 @@ func actionInclude(ctx context.Context, i *model.Include, sp *stepParams) error
 			return fmt.Errorf("Stat(): %w", err)
 		}
 
-		if err := copyRecursive(p.Pos, absSrc, absDst, sp.fs); err != nil {
+		// Allow later includes to replace earlier includes in the scratch
+		// directory. This doesn't affect whether files in the final destination
+		// directory will be overwritten; that comes later.
+		const overwrite = true
```
yolocs, line 52: Thank you!

yolocs (approved):

PR #12, opened 2023-06-06
Fix a bug causing scratch dir not to be cleaned up

drevell:
The desired behavior is that the scratch dir and template dir are both removed unless the user specifies --keep-temp-dirs. We failed to actually clean up the scratch dir.

There was also an incorrect test where this was supposed to be tested; we incorrectly asserted the oppposite of what we should have. This has been fixed here.

### DIFF
--- a/templates/commands/render.go
+++ b/templates/commands/render.go
@@ -271,11 +271,15 @@ type runParams struct {
 
 // realRun is for testability; it's Run() with fakeable interfaces.
 func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
+	var tempDirs []string
+	defer func() {
+		err := r.maybeRemoveTempDirs(ctx, rp.fs, tempDirs...)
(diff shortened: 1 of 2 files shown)


yolocs (approved):

templates/commands/render.go
```diff
@@ -271,11 +271,15 @@ type runParams struct {
 
 // realRun is for testability; it's Run() with fakeable interfaces.
 func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
+	var tempDirs []string
+	defer func() {
+		err := r.maybeRemoveTempDirs(ctx, rp.fs, tempDirs...)
+		outErr = errors.Join(outErr, err)
```
sethvargo, line 277: Nit: I think you can simplify this:

```go
outErr = errors.Join(outErr, r.maybeRemoveTempDirs(ctx, rp.fs, tempDirs...))
```

PR #13, opened 2023-06-07
Copy the scratch dir contents to the dest dir

drevell:
This is the final phase of the the rendering process: taking each file and directory in the scratch dir and copying it into the destination.

We want to make this as atomic as possible. If there are any errors, we want to leave the destination directory untouched. We don't want half of the output files to succeed and half of them to be missing; this would be annoying for users to clean up. To achieve this atomicity, we first do a "dry run" copy that checks whether the copy *would* succeed, then we do a real copy. This could be made more efficient in the future, but it's good enough for now.

In this PR we also switch from ReadFile+WriteFile to io.Copy; it seemed silly to read a buffer then just write it again.

### DIFF
--- a/templates/commands/render.go
+++ b/templates/commands/render.go
@@ -366,8 +366,8 @@ func parseAndExecuteGoTmpl(m model.String, inputs map[string]string) (string, er
 }
 
 // "from" may be a file or directory. "pos" is only used for error messages.
-func copyRecursive(pos *model.ConfigPos, from, to string, rfs renderFS, overwrite, dryRun bool) (outErr error) {
-	return fs.WalkDir(rfs, from, func(path string, d fs.DirEntry, err error) error { //nolint:wrapcheck
+func copyRecursive(pos *model.ConfigPos, fromRoot, toRoot string, rfs renderFS, overwrite, dryRun bool) error {
+	return fs.WalkDir(rfs, fromRoot, func(path string, de fs.DirEntry, err error) error { //nolint:wrapcheck
 		if err != nil {
 			return err // There was some filesystem error. Give up.
 		}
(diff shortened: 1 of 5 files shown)


templates/model/pos.go
```diff
@@ -62,15 +62,11 @@ func (c *ConfigPos) AnnotateErr(err error) error {
 //	ErrWithPos(pos, "Foo(): %w", err)
 //
 // Calling this function with a zero or nil ConfigPos is valid. The resulting
-// error will just say that the position is unknown.
+// error will just omit information about config location.
 func ErrWithPos(pos *ConfigPos, fmtStr string, args ...any) error {
```
yolocs, line 66: Just realized it won't hurt to add a unit test for `ErrWithPos`?

templates/commands/render.go
```diff
@@ -397,21 +396,16 @@ func copyRecursive(pos *model.ConfigPos, from, to string, rfs renderFS, overwrit
 			if !os.IsNotExist(err) {
 				return model.ErrWithPos(pos, "Stat(): %w", err) //nolint:wrapcheck
 			}
+			if err := rfs.MkdirAll(dirToCreate, mkdirPerms); err != nil {
+				return model.ErrWithPos(pos, "MkdirAll(): %w", err) //nolint:wrapcheck
+			}
```
yolocs, line 401: So to confirm this is intentional - we could still create empty dirs in the destination in dryrun mode?

templates/commands/render.go
```diff
@@ -429,10 +423,26 @@ func copyRecursive(pos *model.ConfigPos, from, to string, rfs renderFS, overwrit
...
-		if err := rfs.WriteFile(dst, buf, info.Mode().Perm()); err != nil {
-			return fmt.Errorf("failed writing to scratch file: WriteFile(): %w", err)
+		wf, err := rfs.OpenFile(dst, os.O_CREATE|os.O_WRONLY, info.Mode().Perm())
+		if err != nil {
+			return model.ErrWithPos(pos, "OpenFile(): %w", err) //nolint:wrapcheck
+		}
+		defer func() { outErr = errors.Join(outErr, wf.Close()) }()
+
+		if _, err := io.Copy(wf, rf); err != nil {
+			return fmt.Errorf("Copy(): %w", err)
```
yolocs, line 445: nit: it seems it's not consistent when we will add path or not. On line 418, we have `Stat(%s)`. On line 423 we have `Stat()`. Was that intentional or we should make them consistent?

drevell: Fixed, stat() includes its path already, so we don't need the %s.

templates/commands/render.go
drevell: That was actually a bug, thanks for catching it. I just rewrote this whole thing to make the control flow less complex and bug-prone.

drevell: And added a test.

templates/model/pos.go
drevell: Added.

templates/commands/render.go
```diff
@@ -367,76 +366,123 @@ func parseAndExecuteGoTmpl(m model.String, inputs map[string]string) (string, er
...
-			dirToCreate = filepath.Dir(dst)
-		}
+// A helper function that's called by the WalkDirFunc inside copyRecursive in
+// the case that the source path is a directory.
+func copyRecursiveDir(path string, de fs.DirEntry, pos *model.ConfigPos, fromRoot, toRoot string, rfs renderFS, dryRun bool) error {
+	relToSrc, err := filepath.Rel(fromRoot, path)
+	if err != nil {
+		return model.ErrWithPos(pos, "filepath.Rel(%s,%s): %w", fromRoot, path, err) //nolint:wrapcheck
+	}
+	dst := filepath.Join(toRoot, relToSrc)
```
yolocs, line 393: Tabs seem to be off on these lines.

templates/commands/render.go
```diff
@@ -367,76 +366,123 @@ func parseAndExecuteGoTmpl(m model.String, inputs map[string]string) (string, er
 }
 
 // "from" may be a file or directory. "pos" is only used for error messages.
-func copyRecursive(pos *model.ConfigPos, from, to string, rfs renderFS, overwrite bool) error {
-	return fs.WalkDir(rfs, from, func(path string, d fs.DirEntry, err error) error { //nolint:wrapcheck
+func copyRecursive(pos *model.ConfigPos, fromRoot, toRoot string, rfs renderFS, overwrite, dryRun bool) error {
```
yolocs, line 369: Another nit: sometimes we use `from` and `to` and sometimes we use `dst`. Does it make sense to consistently use `src` and `dst`? Like `srcRoot` and `dstRoot`. This is also consistent with the parameters naming in io.Copy

templates/commands/render.go
```diff
@@ -367,76 +366,123 @@ func parseAndExecuteGoTmpl(m model.String, inputs map[string]string) (string, er
...
-		// directory tree, without naming its parent directory. We can't
-		// rely on WalkDir having traversed the parent directory of $path,
-		// so we must create the target directory if it doesn't exist.
-		dirToCreate := dst
-		if !d.IsDir() {
-			dirToCreate = filepath.Dir(dst)
-		}
+// A helper function that's called by the WalkDirFunc inside copyRecursive in
+// the case that the source path is a directory.
+func copyRecursiveDir(path string, de fs.DirEntry, pos *model.ConfigPos, fromRoot, toRoot string, rfs renderFS, dryRun bool) error {
```
yolocs, line 388: Dumb question - since we already create parent dir when copying files, can we just skip copying dirs entirely? 
One obvious caveat is that we then can't template empty folders.

templates/commands/render_action_include.go
```diff
@@ -51,7 +51,7 @@ func actionInclude(ctx context.Context, i *model.Include, sp *stepParams) error
 		// directory will be overwritten; that comes later.
 		const overwrite = true
 
-		if err := copyRecursive(p.Pos, absSrc, absDst, sp.fs, overwrite); err != nil {
+		if err := copyRecursive(p.Pos, absSrc, absDst, sp.fs, overwrite, false); err != nil {
```
yolocs, line 54: Why this change? We should overwrite when copying to scratch, right?

templates/commands/render.go
yolocs: But considering git also ignore empty folders, maybe that's actually good.

templates/commands/render.go
```diff
@@ -309,6 +299,15 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
 		return err
 	}
 
+	// Commit the contents of the scratch directory to the output directory. We
+	// first do a dry-run to check that the copy will fully succeed, so we don't
+	// leave a half-done mess in the user's dest directory.
```
sethvargo, line 304: Can we actually guarantee that it will succeed on the second try? I'm wondering if this is overkill.

templates/commands/render.go
sethvargo: I like `src` and `dst`

templates/commands/render_action_include.go
drevell: I think you've mistaken the dryRun param for the overwrite param.

templates/commands/render.go
drevell: Updated with src and dst.

templates/commands/render.go
drevell: It's not a perfect guarantee, but at least we avoid obvious issues like "destination file already exists." It's much better than just yolo'ing the files over. I changed the wording to "likely to succeed." 

Let's not let perfect be the enemy of good here.

templates/commands/render.go
drevell: > since we already create parent dir when copying files,

This is only sometimes true. In the general case, we can't assume that the parent has been traversed. The user may do something like this:

```
- action: include
  params:
    paths: ['a/deeply/nested/file.txt']
```

In which case the directories "a", "deeply", and "nested" will need to be created by mkdirall, and will NOT be seen by the WalkDirFunc. So I think the logic is corrrect as written.

 I tried to make this clear in the comment at line 428. LMK if you think it needs to be reworded.

templates/commands/render.go
drevell: Can you be more specific? It looks fine to me, and "gofumpt" is fine with it.

templates/commands/render_action_include.go
yolocs: Ah, right!

yolocs (approved):

PR #14, opened 2023-06-08
Add the stringreplace action

drevell:
This is an action that performs a straightforward string replacement on a list of paths (either directories or files). Template variables can be used when specifying any of the parameters:

- paths in which to replace
- strings to be replaced
- strings to be used as replacements.

Most of the logic is in the walkAndModify() function, and the string replacement logic is a thin wrapper around it. This lays the foundation for actionRegexReplace() and actionGoTemplate() to also use walkAndModify().

### DIFF
--- a/templates/commands/render_action.go
+++ b/templates/commands/render_action.go
@@ -42,7 +42,7 @@ func walkAndModify(pos *model.ConfigPos, rfs renderFS, scratchDir, relPath strin
 		return model.ErrWithPos(pos, "Stat(): %w", err) //nolint:wrapcheck
 	}
 
-	return filepath.WalkDir(walkFrom, func(path string, d fs.DirEntry, err error) error {
+	return filepath.WalkDir(walkFrom, func(path string, d fs.DirEntry, err error) error { //nolint:wrapcheck
 		if err != nil {
 			// There was some filesystem error. Give up.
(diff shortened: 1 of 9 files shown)


sethvargo (dismissed):

templates/commands/render_action.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+// Called with the contents of a file, and returns the new contents of the file
+// to be written.
+type walkAndModifyVisitor func([]byte) ([]byte, error)
+
+// Recursively traverses the directory or file scratchDir/relPath, calling the
+// given visitor for each file. If the visitor returns modified file contents
+// for a given file, that file will be overwritten with the new contents.
+func walkAndModify(pos *model.ConfigPos, rfs renderFS, scratchDir, relPath string, v walkAndModifyVisitor) error {
```
yolocs, line 33: Recommend moving a few funcs over `parseAndExecuteGoTmpl` and related, and `copyRecursive` and related.

yolocs (approved):

templates/commands/render_action.go
drevell: Ok, will do in a followup PR so I don't lose the precious approval bit by adding new commits to this PR.

yolocs (approved):

PR #15, opened 2023-06-14
Add actions regex_replace and regex_name_lookup

drevell:
The regex_replace action replaces a regex match (or a subgroup thereof) with a given string. The replacement string can use subgroup references like $1 or ${groupname}, and can also use Go template expressions. An example is:

  - regex: '([a-zA-Z]+), (?P<adjective>[a-zA-Z]+) world!'
    subgroup: 1
    with: 'fresh'

This would transform a file containing "Hello, cool world!" to "Hello, fresh world!"

The regex_name_lookup action replaces every named subgroup with the template input variable of the same name. For example:

  - regex: 'Hello, (?P<adjective>[a-zA-Z]+) world!'

... and suppose the CLI user provided --var=adjective=fresh. This would also transform a file containing "Hello, cool world!" to "Hello, fresh world!"


templates/commands/render.go
```diff
@@ -361,15 +363,15 @@ func parseGoTmpl(tpl string) (*template.Template, error) {
 	return template.New("").Option("missingkey=error").Parse(tpl) //nolint:wrapcheck
 }
 
-func parseAndExecuteGoTmpl(m model.String, inputs map[string]string) (string, error) {
-	goTmpl, err := parseGoTmpl(m.Val)
+func parseAndExecuteGoTmpl(pos *model.ConfigPos, s string, inputs map[string]string) (string, error) {
```
sethvargo, line 366: Nit: it's not clear what `s` is unless you read the code - it looks like it's the Go Template?

templates/commands/render_action.go
```diff
@@ -78,3 +79,35 @@ func walkAndModify(pos *model.ConfigPos, rfs renderFS, scratchDir, relPath strin
...
+	for i, re := range regexes {
+		templated, err := parseAndExecuteGoTmpl(re.Pos, re.Val, inputs)
+		if err != nil {
+			return nil, err
+		}
+
+		compiled[i], err = regexp.Compile(templated)
+		if err != nil {
+			return nil, model.ErrWithPos(re.Pos, "failed compiling regex: %w", err) //nolint:wrapcheck
+		}
```
sethvargo, line 94: Do we want to return a multierror here instead? It could be useful to get _all_ your errors at once instead of "fix error, run again, fix different error, run again, fix different error ..."

```suggestion
		templated, err := parseAndExecuteGoTmpl(re.Pos, re.Val, inputs)
		if err != nil {
		  merr = multierror.Append(merr, err)
			continue
		}

		compiled[i], err = regexp.Compile(templated)
		if err != nil {
			merr = multierror.Append(merr, model.ErrWithPos(re.Pos, "failed compiling regex: %w", err))
			continue
		}
```

templates/commands/render_action.go
```diff
@@ -78,3 +79,35 @@ func walkAndModify(pos *model.ConfigPos, rfs renderFS, scratchDir, relPath strin
...
+	}
+
+	return compiled, nil
+}
+
+// Returns a slice where the order of the elements is reverse. Returns the same
+// slice as the input in the case where the length of the input is 0 or 1,
+// because the reverse operation is a no-op on these.
+func reverse[T any](in []T) []T {
```
sethvargo, line 103: Why would we do this vs iterating over the slice in reverse? It feels unnecessarily expensive.

templates/commands/render_action_regexnamelookup.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func replaceWithNameLookup(allMatches [][]int, b []byte, rn *model.RegexNameLookupEntry, re *regexp.Regexp, inputs map[string]string) ([]byte, error) {
+	for i := 1; i < len(re.SubexpNames()); i++ { // skip group 0, which is always unnamed because it's "the whole regex match"
+		if re.SubexpNames()[i] == "" {
+			return nil, model.ErrWithPos(rn.Regex.Pos, `all capturing groups in a regex_name_lookup must be named, like (?P<myinputvar>myregex), not like (myregex)`) //nolint:wrapcheck
+		}
+	}
+
+	for _, oneMatch := range allMatches {
```
sethvargo, line 78: Instead of reversing on L54 and then doing this, I'm not sure I understand why we can't walk the slice in reverse:

```go
for i := len(allMatches)-1; i >= 0; i-- {
  oneMatch := allMatches[i]
}
```

templates/commands/render_action_regexnamelookup_test.go
```diff
@@ -0,0 +1,171 @@
...
+					},
+				},
+			},
+			want: map[string]string{
+				"a.txt": "alpha beta gamma",
+			},
+			wantErr: `must be named`,
+		},
+		{
+			name: "template_expr_in_regex_and_groupname_should_work",
```
sethvargo, line 98: Super nit: I'm not sure "should_work" is helpful in the name.

templates/commands/render_action_regexreplace.go
```diff
@@ -17,10 +17,144 @@ package commands
...
+				// matches indices.
+				allMatches := reverse(cr.FindAllSubmatchIndex(b, -1))
+
+				var err error
+				b, err = replaceWithTemplate(allMatches, b, rr, cr, sp.inputs)
+				if err != nil {
+					return nil, err
+				}
+			}
+			return b, nil
```
sethvargo, line 74: I'm not sure I understand why we're returning the last replacement?

templates/commands/render.go
drevell: I did some variable renaming to make this more clear.

templates/commands/render_action.go
drevell: SGTM, done.

templates/commands/render_action.go
drevell: Because `reverse()` seemed more readable me. It lets us use `for _ oneMatch := range allMatches` and avoids creating more index variables. This code already has a confusing number of indices (FindAllSubmatchIndex returns a slice-of-slices-of-indices 😱 ), so avoiding more indices seemed desirable.

 I take it you disagree though; so I switched over to iterating in reverse. I don't have strong feelings on it.

templates/commands/render_action_regexnamelookup.go
drevell: Done

templates/commands/render_action_regexnamelookup_test.go
drevell: Let me send a follow PR to fix this, I've been doing this throughout this codebase.

templates/commands/render_action_regexreplace.go
drevell: The result of each transformation is pipelined into the following transformation, and the final result is returned. There's nothing special about the last replacement. Note the `=` rather than `:=` in `b, err = replaceWithTemplate(...`.

templates/commands/render_action_regexnamelookup.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+	}
+
+	return nil
+}
+
+func replaceWithNameLookup(allMatches [][]int, b []byte, rn *model.RegexNameLookupEntry, re *regexp.Regexp, inputs map[string]string) ([]byte, error) {
+	for i := 1; i < len(re.SubexpNames()); i++ { // skip group 0, which is always unnamed because it's "the whole regex match"
+		if re.SubexpNames()[i] == "" {
+			return nil, model.ErrWithPos(rn.Regex.Pos, `all capturing groups in a regex_name_lookup must be named, like (?P<myinputvar>myregex), not like (myregex)`) //nolint:wrapcheck
```
yolocs, line 69: Do we need `//nolint:wrapcheck` here?

templates/commands/render_action_regexreplace.go
sethvargo: Ah I missed the non-assignment. Thanks.

sethvargo (approved):

templates/commands/render_action_regexreplace.go
```diff
@@ -17,10 +17,144 @@ package commands
...
+	return b, nil
+}
+
+// A regular expression that matches regex subgroup references like "$5" or
+// "${5}" in a string that will be passed to Regexp.Expand().
+var subGroupExtractRegex = regexp.MustCompile(`\$+` + // some number of dollar signs
+	`[{]?` + // then optionally has a brace character
+	`([0-9]+)`) // and then has some number of decimal digits (as a capturing group)
+
+// Given a string that will be passed to Regexp.Expand(), try to find the
```
yolocs, line 115: This is some next level code complexity

yolocs (commented):
Rant first - As a maintainer, while I appreciate the thorough documentation in code, I doubt whether I can effectively debug the regex related code in case something goes wrong. As a potential template builder myself, I would refuse to use regex because it would take me more time to debug whether I'm writing regex correctly.

---

Future - I'm interested in finding how our future users will use this. Would they prefer regex or the other ways?

---

For this PR - I don't have much Go readability comments - LGTM. But due to my limited regex knowledge, it's better to have some extra eyes to take a look.

templates/commands/render_action_regexnamelookup.go
drevell: Yep. The linter will flag any superfluous `//nolint` comments, so you can know that it's necessary.

drevell:
> Rant first - As a maintainer, while I appreciate the thorough documentation in code, I doubt whether I can effectively debug the regex related code in case something goes wrong. As a potential template builder myself, I would refuse to use regex because it would take me more time to debug whether I'm writing regex correctly.
> 
> Future - I'm interested in finding how our future users will use this. Would they prefer regex or the other ways?
> 
> For this PR - I don't have much Go readability comments - LGTM. But due to my limited regex knowledge, it's better to have some extra eyes to take a look.

When you're back from OOO, let's get together and explore ways to simplify.

PR #17, opened 2023-06-15
Add go_template action, plus more

drevell:
The go_template action executes the given path(s) as go templates, which can access the user-provided template variables when doing substitutions. The resulting template output is written back to the file that it came from.

Also moved some functions from render.go into render_action.go per Chen's earlier request.

For friendly error messages, there's new code that wraps the error returned by template.Execute() to provide a friendlier error message in the case where the user's template references a nonexistent input.


templates/commands/render_action.go
```diff
@@ -100,3 +111,150 @@ func templateAndCompileRegexes(regexes []model.String, inputs map[string]string)
...
+// there's no reason to print out spec file location in an error message. If
+// template execution fails because of a missing input variable, the error will
+// be wrapped in a unknownTemplateKeyError.
+func parseAndExecuteGoTmpl(pos *model.ConfigPos, tmpl string, inputs map[string]string) (string, error) {
+	parsedTmpl, err := parseGoTmpl(tmpl)
+	if err != nil {
+		return "", model.ErrWithPos(pos, `error compiling as go-template: %w`, err) //nolint:wrapcheck
+	}
+
+	// As of go1.20, if the template references a nonexistent variable, then the
```
sethvargo, line 131: Optional side quest: do you want to file a bug upstream about this?

templates/commands/render_action_test.go
```diff
@@ -209,3 +213,337 @@ func TestWalkAndModify(t *testing.T) {
 		})
 	}
 }
+
+func TestCopyRecursive(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name                 string
+		fromDirContents      map[string]modeAndContents
```
sethvargo, line 222:
```suggestion
		fromDirContents      map[string]*modeAndContents
```

throughout

sethvargo (approved):

templates/commands/render_action.go
drevell: I just found https://github.com/golang/go/issues/57801 for this, and added my thumbs up emoji to it 😀

PR #20, opened 2023-06-15
Validate apiVersion and kind spec fields

drevell:
Currently each field has only one valid value, so this is simple.

### DIFF
--- a/templates/commands/render_test.go
+++ b/templates/commands/render_test.go
@@ -167,7 +167,7 @@ func TestRealRun(t *testing.T) {
 	t.Parallel()
 
 	specContents := `
-apiVersion: 'abc'
+apiVersion: 'abcxyz.dev/cli/v1alpha1'
 kind: 'Template'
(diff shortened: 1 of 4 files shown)


templates/commands/render_test.go
```diff
@@ -167,7 +167,7 @@ func TestRealRun(t *testing.T) {
 	t.Parallel()
 
 	specContents := `
-apiVersion: 'abc'
+apiVersion: 'abcxyz.dev/cli/v1alpha1'
```
sethvargo, line 170: This feels weird? I would expect `cli.abcxyz.dev/v1alpha1` or `dev.abcxyz.cli/v1alpha1`

sethvargo (approved):

templates/commands/render_test.go
drevell: SGTM, updated.

sethvargo (approved):

Issue #24, opened 2023-06-15
Command line autocompletion


drevell:
### TL;DR

The library that we use for parsing command line flags, https://pkg.go.dev/github.com/abcxyz/pkg/cli, now offers support for autocompletion. I don't know what's possible here, but here are some things that might perhaps be possible:

- `abc templates render ~/dir/`<tab> would autocomplete with all possible directories inside `~/dir`
- Maybe remote template URLs could autocomplete? `abc templates render github.com/abcxyz/examples`<tab> could suggest all of the available example templates in that directory
- The `--dest` flag could autocomplete as a local directory name

Originally suggested by @sethvargo.

Issue #29, opened 2023-06-15
Write user guide


drevell:
### TL;DR

We need documentation for template users about how to:

 - Install the CLI
 - Do the complete flow of rendering a template (git clone / git init, render, push, PR, merge)
   - Flags to specify
 - Dealing with common errors
 - Install shell autocompletions (`COMP_INSTALL=1 COMP_YES=1 ./abc`)
 - Everything else a user needs to know

This will be a markdown file in this git repo.

PR #34, opened 2023-06-16
Add basic predictions for paths

sethvargo:
Fixes GH-24


sethvargo:
This does everything but "complete remote templates". It's possible with the API, but we don't actually have a canonical list yet. I can add the stubbed infrastructure that just returns "github.com/abcxyz/examples" if that's helpful.

templates/commands/render_test.go
```diff
@@ -76,7 +76,7 @@ func TestParseFlags(t *testing.T) {
 				flagSpec:           "./spec.yaml",
 				flagDest:           ".",
 				flagGitProtocol:    "https",
-				flagInputs:         nil,
+				flagInputs:         map[string]string{},
```
sethvargo, line 79: This was actually a bug in the CLI. Chen fixed it in https://github.com/abcxyz/pkg/pull/128, and the update brought it in.

examples/templates/render/hello_jupiter/README.md
```diff
@@ -6,21 +6,27 @@ by doing a simple string replacement.
...
 
-1. First, compile the `abc` binary:
+1. Compile the `abc` binary:
 
     ```shell
     $ go build
     ```
 
-2. Now execute the template defined by the `spec.yaml` file in the example directory.
+1. (Optional) Install shell completions:
```
drevell, line 15: I'd rather leave this out of the examples, and put it in the (eventual) user quickstart docs. I think it's important for this example to be as utterly simple as possible. I added a note to #29 so we don't forget.

templates/commands/render.go
```diff
@@ -116,6 +117,7 @@ func (r *Render) Flags() *cli.FlagSet {
 		Example: "path/to/spec.yaml",
 		Target:  &r.flagSpec,
 		Default: "./spec.yaml",
+		Predict: predict.Files("*"),
```
drevell, line 120: I'm not sure about this. `--spec` is relative to the downloaded template directory, so I don't think this will autocomplete correctly. It might not be possible to do any cleverness here.

templates/commands/render_test.go
drevell: 👍

drevell (changes requested):

examples/templates/render/hello_jupiter/README.md
sethvargo: Done

templates/commands/render.go
sethvargo: Done

drevell (approved):

Issue #45, opened 2023-06-21
Require named subgroups in regexes, remove support for numbered subgroups


drevell:
### TL;DR

To make spec files easier to maintain, let's enforce a more readable style. Let's require regular expressions to use named subgroups, and require that those groups are referenced by name when expanded.

A good example of something we want:
```
action: 'regex_replace'
replacements:
  - regex: 'Hello, (?P<name>[a-zA-Z]+)'    # Good: capturing group is named
    with: 'Good day, ${name}'    # Good: expansion is referencing subgroup by name
```

A bad example of something we don't want:
```
action: 'regex_replace'
replacements:
  - regex: 'Hello, ([a-zA-Z]+)'    # Bad: capturing group is not named
    with: 'Good day, ${1}'    # Bad: expansion is referencing subgroup by number
```

To do this, we will:
 - check the regular expressions after compiling them to make sure that every subgroup has a name (all the strings in Regexp.SubexpNames() should be non-empty except index 0)
 - check the strings we pass to Regexp.Expand()  and reject anything that looks like a numbered subgroup reference, like `$3` or `$$$3`

We should also change the `subgroup` field of the RegexpReplace model to be a string instead of a number:

```
action: 'regex_replace'
replacements:
  - regex: 'Hello, (?P<name>[a-zA-Z]+)'    # Good: capturing group is named
    subgroup: "name"  # This is currently an int but should become a string
    with: '{{.person_name}'  
```

PR #48, opened 2023-06-22
Use named subgroups exclusively

drevell:
To make sure our template spec files are maintainable, we force the use of named subgroups (eg `(?P<mygroup>re`) rather than numbered subgroups (eg `(re)`). Correspondingly we force the replacement strings to use named subgroup expansion (eg `${mygroup}`) rather than numbered (eg `${1}`).

Also rename the "subgroup" field to "subgroup_to_replace" to make it more obvious how it works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,9 +291,9 @@ Params:
     readability. Use a non-capturing group (like `(?:abc)|(?:abc)`) if you need
     grouping without capturing.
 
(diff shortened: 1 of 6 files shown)


drevell:
Fixes #45

sethvargo (approved):

PR #56, opened 2023-06-23
Add t/rest_server

kevya-google:
This PR adds a template for a simple HTTP/JSON REST server that can be rendered, future work could include addition or replacement of different components of the server such as a database. Also included are some README adjustments.


drevell (dismissed):

t/rest_server/main.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/abcxyz/pkg/renderer"
+	"github.com/go-chi/chi/v5"
+)
+
+var bind = flag.String("bind", ":8080", "Specifies server ip address and port to listen on.")
+
+func handleHello(h *renderer.Renderer) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		h.RenderJSON(w, http.StatusOK, map[string]string{"message": "hello world"})
```
sethvargo, line 37: Is this an opportunity to use a replacement to show the power of the CLI?

t/rest_server/main.go
```diff
@@ -0,0 +1,108 @@
...
+//   - listening to incoming requests in a goroutine
+func realMain(ctx context.Context) error {
+	// Make a new renderer for rendering json.
+	// Don't provide filesystem as we don't have templates to render.
+	h, err := renderer.New(ctx, nil,
+		renderer.WithOnError(func(err error) {
+			log.Printf("failed to render: %v", err)
+		}))
+	if err != nil {
+		return fmt.Errorf("failed to create renderer for main server %w", err)
```
sethvargo, line 53:
```suggestion
		return fmt.Errorf("failed to create renderer for main server: %w", err)
```

t/rest_server/main.go
```diff
@@ -0,0 +1,108 @@
...
+		return fmt.Errorf("error from server listener: %w", err)
+	case <-ctx.Done():
+	}
+
+	shutdownCtx, done := context.WithTimeout(context.Background(), 5*time.Second)
+	defer done()
+	if err := s.Shutdown(shutdownCtx); err != nil {
+		return fmt.Errorf("failed to shutdown: %w", err)
+	}
+	return nil
```
sethvargo, line 95: It seems a little weird that we'd depend on pkg/renderer, but not pkg/serving (or pkg/logging).

t/rest_server/main.go
drevell: I actually advised against doing this; that seemed a bit like complexity for its own sake. We should limit the data we need from users to the minimum to streamline the user experience.

As we evolve this template and add features (auth, DB) then there will be plenty of opportunities to demonstrate our replacement powers.

t/rest_server/main.go
kevya-google: Ok, I'll try adding these (if that's what we're recommending for CLI users to use)

drevell (changes requested):
I just realized there are no tests here... let's add some

t/rest_server/main.go
kevya-google: Will leave as no substitution for now.

t/rest_server/main.go
kevya-google: Added pkg/serving, pkg/logging.

kevya-google:
> I just realized there are no tests here... let's add some

Added test.

pdewilde (dismissed):
lgtm

t/rest_server/main.go
```diff
@@ -0,0 +1,100 @@
...
+
+func handleHello(h *renderer.Renderer) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		logger := logging.FromContext(r.Context())
+		logger.Info("handling request")
+		h.RenderJSON(w, http.StatusOK, map[string]string{"message": "hello world"})
+	})
+}
+
+// RealMain creates an example backend HTTP server.
```
drevell, line 45: RealMain -> realMain to match the case of the function declaration

t/rest_server/main.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func main() {
+	ctx, done := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	defer done()
+	logger := logging.FromContext(ctx)
+
+	flag.Parse()
+	if err := realMain(logging.WithLogger(ctx, logger)); err != nil {
+		done()
```
drevell, line 96: Maybe add a comment here explaining that the call to done() is made because deferred function calls won't execute due to logger.Fatal() never returning.

t/rest_server/main_test.go
```diff
@@ -0,0 +1,79 @@
...
+		if err := realMain(ctx); err != nil {
+			select {
+			case errCh <- err:
+			default:
+			}
+		}
+	}()
+
+	cmd := exec.Command(
+		"curl",
```
drevell, line 48: Not sure we should take a dependency on curl. My suggestion would be:

 - Here in this function, use `http.Get()` instead of exec'ing curl, to make a real network request
 - And also create a separate test that uses httptest (or whatever the convention is for abcxyz, I don't recall) to just test the handlers without a listening on a real port

drevell (changes requested):

t/rest_server/main.go
kevya-google: Updated comment.

t/rest_server/main.go
kevya-google: Added comment. Do the t.Fatals in the test file also need done() calls in the error handling if?

t/rest_server/main_test.go
```diff
@@ -0,0 +1,81 @@
...
+	default:
+	}
+
+	// Wait for done
+	select {
+	case <-doneCh:
+	case <-time.After(time.Second):
+		t.Errorf("expected server to be stopped")
+	}
+}
```
drevell, line 81: I think this can be simplified a little:

```suggestion
func TestRealMain(t *testing.T) {
	ctx := logging.WithLogger(context.Background(), logging.TestLogger(t))
	ctx, done := context.WithCancel(ctx)
	defer done()

	var realMainErr error
	finishedCh := make(chan struct{}, 0)
	go func() {
		defer close(finishedCh)
		realMainErr = realMain(ctx)
	}()

	time.Sleep(100 * time.Millisecond) // wait for server startup
	resp, err := http.Get(fmt.Sprintf("http://localhost:%s", defaultPort))
	if err != nil {
		t.Fatal(err)
	}
	defer resp.Body.Close()

	b, err := io.ReadAll(resp.Body)
	if err != nil {
		t.Fatal(err)
	}

	want := "hello world"
	if !strings.Contains(string(b), want) {
		t.Errorf("unexpected response: (-got,+want)\n%s", cmp.Diff(string(b), want))
	}

	// stop server
	done()

	// Wait for done
	select {
	case <-finishedCh:
	case <-time.After(time.Second):
		t.Fatalf("expected server to be stopped")
	}
  
  if realMainErr != nil {
    t.Errorf("realMain(): %v", realMainErr)
  }
}
```

drevell (changes requested):

t/rest_server/main.go
kevya-google: Discussed offline, t.Fatals don't require this.

sethvargo (approved):

t/rest_server/main_test.go
kevya-google: Changed to http.get(), Added separate test for handle function.

t/rest_server/main.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// This will block until the provided context is cancelled.
+	if err := server.StartHTTP(ctx, httpServer); err != nil {
+		return fmt.Errorf("error starting server: %w", err)
+	}
+	return nil
+}
+
+func main() {
+	ctx, done := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
```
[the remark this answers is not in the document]
kevya-google: Added comment.

t/rest_server/main_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	var realMainErr error
+	finishedCh := make(chan struct{})
+	go func() {
+		defer close(finishedCh)
+		realMainErr = realMain(ctx)
+	}()
+
+	time.Sleep(100 * time.Millisecond) // wait for server startup
+	resp, err := http.Get(fmt.Sprintf("http://localhost:%s/", defaultPort))
```
drevell, line 45: Should `defaultPort` be `*port`? They'll probably be the same, since the user is unlikely to specify `--port=foo` when running test, but technically I think it woulld be more correct to use `*port`.

t/rest_server/main_test.go
```diff
@@ -0,0 +1,119 @@
...
+		},
+	}
+	for _, tc := range cases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			server := httptest.NewServer(handleHello(h))
+			t.Cleanup(func() { server.Close() })
```
drevell, line 97: I think we can just use `defer server.Close()` instead of t.Cleanup(). Both will work, but I think it's good to be consistent, and defer is a little simpler.

The main use case for t.Cleanup() is for helper functions that create test resources, and these can't use `defer` because the function returns before the test finishes.

drevell (approved):

t/rest_server/main_test.go
sethvargo: I generally always use `t.Cleanup` because I can explain the rule "if you're in _test.go, use t.Cleanup, otherwise use defer" pretty cleanly. However, "if you're in _test.go and the test isn't a parent test creating resources that need to live until all subtests complete, use t.Cleanup, otherwise use defer" always felt like more cognitive overhead to me.

Either way, let's pick something and put it in the style guide.

drevell: This rule would be pretty simple: "use defer when you can, if not, use t.Cleanup()"

sethvargo: But how does one know "when you can"? It feels like asking developers to take in a lot of context to make that decision.

pdewilde: t.Cleanup() also means its easier to refactor duplicated code into a helper.

drevell: I'm fine with t.Cleanup() since the consensus seems to be going that way.

kevya-google: Ok, will leave as-is.

t/rest_server/main_test.go
kevya-google: Yes, I agree. Changed to *port

drevell (dismissed):

drevell (approved):

PR #58, opened 2023-06-23
feat: add unknown input validation and input defaults

### LINKED ISSUE #28: Check template input variables
### TL;DR

Currently we just accept the `--input` variables provided by the CLI user. We should check that:

 - Each --input corresponds to a valid declared input in the template's spec.yaml file (the user mustn't provide extra unnecessary inputs)
 - Each required input in the template's spec.yaml file has a corresponding --input (the user mustn't forget any template inputs that are needed)

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

### LINKED ISSUE #40: Support optional inputs with defaults
### TL;DR

Currently, if the user doesn't provide a value for every input, the template won't render. Instead, we could have the spec.yaml declare an input as optional and provide a default value.

```
inputs:
  - name: 'concurrent_workers'
    required: false  # default is true
    default: '10' # must be a YAML string, because this is a string replacement tool, hence the quotes
    desc: 'the number of worker threads'
    
```

There is some existing code for this that declares a "required" field in the YAML, but the logic isn't implemented yet; Required" is useless without "default": https://github.com/abcxyz/abc/blob/c0122ffb6bd49d4c123e184539a22f9fa0b6559f/templates/model/spec.go#L128

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

verbanicm:
Closes #40 
Closes #28


templates/commands/render.go
```diff
@@ -322,6 +328,35 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+// validateAndDefaultInputs validates and sets the input values provided by the user.
+func validateAndDefaultInputs(flagInputs map[string]string, spec *model.Spec) error {
+	var result *multierror.Error
+
+	for _, i := range spec.Inputs {
+		val, ok := flagInputs[i.Name.Val]
+		if !ok || val == "" {
+			if i.Default.Val != "" {
+				flagInputs[i.Name.Val] = i.Default.Val
+			} else if i.Required.Val {
```
drevell, line 340: I'm thinking we should just remove the "Required" field and just say "a field is required if and only if it doesn't have a default." What do you think?

templates/commands/render.go
```diff
@@ -322,6 +328,35 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+	if result != nil {
+		result.ErrorFormat = func(errors []error) string {
+			out := strings.Builder{}
+
+			for _, e := range errors {
+				out.WriteString(fmt.Sprintf("\n  - %s", e))
+			}
+			return out.String()
+		}
+	}
```
drevell, line 355: Elsewhere in `abc`, we just do

```
var merr error

for ... {
  ...
  if err != nil {
    merr = errors.Join(merr,err)
  }
}
return merr
```

That seems to me to be preferable to this multierror scheme because it avoids this 10 lines of ErrorFormat definition.

templates/commands/render.go
```diff
@@ -322,6 +328,35 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
 	return nil
 }
 
+// validateAndDefaultInputs validates and sets the input values provided by the user.
+func validateAndDefaultInputs(flagInputs map[string]string, spec *model.Spec) error {
```
drevell, line 332: Can we also validate that no superfluous extra inputs were provided by the user?

templates/model/spec.go
```diff
@@ -125,12 +125,13 @@ type Input struct {
...
 	Name     String `yaml:"name"`
 	Desc     String `yaml:"desc"`
+	Default  String `yaml:"default"`
 	Required Bool   `yaml:"required"`
 }
 
 // UnmarshalYAML implements yaml.Unmarshaler.
 func (i *Input) UnmarshalYAML(n *yaml.Node) error {
-	knownYAMLFields := []string{"name", "desc", "required"}
+	knownYAMLFields := []string{"name", "desc", "default", "required"}
```
drevell, line 134: I think we need some tests for this in spec_test.go

templates/commands/render_test.go
```diff
@@ -185,6 +185,28 @@ steps:
     paths: ['file1.txt', 'dir1', 'dir2/file2.txt']
 `
 
+	inputsSpecContents := `
```
drevell, line 188: Can we add the inputs to `specContents` and avoid the need to create `inputSpecContents`?

templates/commands/render.go
```diff
@@ -322,6 +328,35 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+			}
+		}
+	}
+
+	if result != nil {
+		result.ErrorFormat = func(errors []error) string {
+			out := strings.Builder{}
+
+			for _, e := range errors {
+				out.WriteString(fmt.Sprintf("\n  - %s", e))
```
drevell, line 351: Maybe we should print full help text here: input name, description, and default. The template might not be documented and they may not know how to look at the spec file, so this may be the only way the user finds out what inputs are needed.

I guess we should also build a `abc templates describe <url>` command so people can find out about inputs, I'll file an issue for that.

drevell (changes requested):

templates/commands/render.go
verbanicm: I feel like there could be a use case, where a string is optional, and if given you render a certain part of a template. I'd say lets leave it for now.

verbanicm: Maybe we can default the default to empty string and remove the required flag, this could achieve the same functionality.

templates/commands/render_test.go
verbanicm: ack

templates/model/spec.go
verbanicm: ack

templates/commands/render.go
verbanicm: ack

templates/commands/render.go
verbanicm: +1

templates/commands/render.go
verbanicm: I like that...

templates/commands/render.go
verbanicm: Torn here, we dont want to overwhelm the user with a large amount of output. Maybe after we implement describe we can inform the user how to describe the template so they can learn the inputs from there.

verbanicm: Created https://github.com/abcxyz/abc/issues/59 to track, we can resolve later.

templates/commands/render.go
```diff
@@ -322,6 +343,92 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+		}
+	}
+
+	return merr
+}
+
+// promptForInput prompts the user for a value for a missing input. The prompt will continue
+// until a value is provided unless the input was configured with a default value, it will be
+// used for an empty input provided by the user.
+func (r *Render) promptForInput(i *model.Input) error {
```
verbanicm, line 391: Not sure this is the best way to do this but.......open to feedback.

verbanicm:
This PR is dependent on https://github.com/abcxyz/pkg/pull/143.

This code will need to update to the latest pkg before merging.

templates/commands/render.go
```diff
@@ -112,6 +115,12 @@ func (r *Render) Flags() *cli.FlagSet {
 	set := cli.NewFlagSet()
 
 	f := set.NewSection("RENDER OPTIONS")
+	f.BoolVar(&cli.BoolVar{
```
sethvargo, line 118: Long term, feels like this might want to be a sticky flag

templates/commands/render.go
```diff
@@ -291,6 +300,18 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+	if err := processUnknownInputs(r.flagInputs, spec); err != nil {
+		return fmt.Errorf("unknown input(s):\n%w", err)
+	}
+
+	if err := r.validateFlagInputs(spec); err != nil {
+		if r.flagQuiet {
+			return fmt.Errorf("missing required input(s):\n%w", err)
+		} else {
+			return fmt.Errorf("failed to validate input(s): %w", err)
+		}
```
sethvargo, line 312: What is this supposed to be doing?

templates/commands/render.go
```diff
@@ -322,6 +343,92 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+}
+
+// validateFlagInputs validates the input values.
+func (r *Render) validateFlagInputs(spec *model.Spec) error {
+	var merr error
+
+	for _, i := range spec.Inputs {
+		_, ok := r.flagInputs[i.Name.Val]
+		if !ok {
+			if !r.flagQuiet {
```
sethvargo, line 371: I see - is "quiet" the correct word for this? It feels like "automated" or "no-prompt" might be better

sethvargo: But also - we can detect whether there's a TTY.

templates/commands/render.go
sethvargo: https://github.com/abcxyz/pkg/blob/main/cli/command.go#L277

templates/commands/render.go
```diff
@@ -322,6 +343,92 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+	var merr error
+
+	specInputs := make(map[string]any)
+	for _, v := range spec.Inputs {
+		specInputs[v.Name.Val] = struct{}{}
+	}
+
+	for key, value := range flagInputs {
+		if _, ok := specInputs[key]; !ok {
+			merr = errors.Join(merr, fmt.Errorf("  - %s=%s", key, value))
```
drevell, line 357: The convention that we use in spec.go is that each error is fully specified, not just a bullet point. `an input with an unknown key of %q was provided, that is not an input accepted by this template` or something.

templates/commands/render.go
drevell: --no-prompt sgtm

templates/commands/render.go
```diff
@@ -322,6 +343,92 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+		}
+	}
+
+	return merr
+}
+
+// validateFlagInputs validates the input values.
+func (r *Render) validateFlagInputs(spec *model.Spec) error {
```
drevell, line 365: I'd suggest renaming this function. The current name `validateFlagInputs` makes it sound like this function is read-only, but it mutates the state of `r.inputs` by calling promptForInput.

templates/commands/render_test.go
```diff
@@ -198,13 +204,16 @@ steps:
...
 		wantDestContents     map[string]string
+		wantFlagInputs       map[string]string
 		wantStdout           string
 		wantErr              string
 	}{
 		{
 			name: "simple_success",
 			flagInputs: map[string]string{
-				"name_to_greet": "🐈",
+				"name_to_greet":   "name",
```
drevell, line 214: Can we make this `name` a real name ("Alice", "Bob", etc)? That will make its behavior a little more self-explanatory.

drevell (changes requested):

templates/commands/render_test.go
```diff
@@ -352,13 +491,21 @@ steps:
...
 				flagDest:           dest,
 				flagForceOverwrite: tc.flagForceOverwrite,
 				flagInputs:         tc.flagInputs,
 				flagKeepTempDirs:   tc.flagKeepTempDirs,
 				flagSpec:           "spec.yaml",
 				source:             "github.com/myorg/myrepo",
 			}
+
+			stdin, _, _ := r.Pipe()
+			stdin.WriteString(tc.stdinInput)
```
verbanicm, line 507: Couldn't figure out a way to send multiple sets of inputs. I believe since scanner.Scan() in the prompt method reads the entire buffer, this no be possible.

Open to suggestions.

sethvargo: I think you can do it in a goroutine, but `scanner.Scan` iterates over newlines, so you might be able to cheat and write something like "abc\ndef\njkl"

verbanicm: I tried that, the first value is ok but the second one disappears.

flags/flags.go
```diff
@@ -0,0 +1,36 @@
...
+// limitations under the License.
+
+// Package flags contains the top level flags for all commands.
+package flags
+
+import "github.com/abcxyz/pkg/cli"
+
+// GlobalFlags represent the shared global flags among all sub-commands. Embed this
+// struct into all sub-commands.
+type GlobalFlags struct {
```
sethvargo, line 22: My general rule is that there's no such thing as a truly global flag; you'll have a situation where the flag doesn't actually make sense. So it's usually better to name these with a more realistic taxonomy and then embed them.

You did the embedding, but I don't think "global flags" is a good name.

verbanicm: How does `AutomationFlags` sound?

templates/model/spec.go
```diff
@@ -123,23 +123,21 @@ type Input struct {
 	// Pos is the YAML file location where this object started.
 	Pos *ConfigPos `yaml:"-"`
 
-	Name     String `yaml:"name"`
-	Desc     String `yaml:"desc"`
-	Required Bool   `yaml:"required"`
+	Name    String  `yaml:"name"`
+	Desc    String  `yaml:"desc"`
+	Default *String `yaml:"default,omitempty"`
```
verbanicm, line 128: Should we leave the required flag and use it as a way to force the user to provide a value? Thinking of a case where an empty string isnt allowed and we cant default the value, e.g. creating a new product in the gcp-org repo.

templates/commands/render.go
```diff
@@ -48,10 +51,13 @@ const (
 
 type Render struct {
 	cli.BaseCommand
+	automationFlags flags.AutomationFlags
 
 	testFS     renderFS
 	testGetter getterClient
 
+	hasTTY bool
```
drevell, line 59: Should this be a field of runParams instead, along with all the other "parameters" to realRun()?

templates/commands/render.go
```diff
@@ -323,6 +347,72 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+func (r *Render) promptForInputs(ctx context.Context, spec *model.Spec) error {
+	for _, i := range spec.Inputs {
+		_, ok := r.flagInputs[i.Name.Val]
+		if !ok {
+			defaultText := ""
+			if i.Default != nil {
+				defaultText = fmt.Sprintf(" (%s)", i.Default.Val)
+			}
+
+			v, err := r.Prompt(ctx, "Enter value for %s%s:   ", i.Name.Val, defaultText)
```
drevell, line 379: Should we print the description here too (maybe on a preceding line(s))? I could go either way.

templates/commands/render.go
```diff
@@ -323,6 +347,72 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+		}
+	}
+
+	return nil
+}
+
+// processRequiredInputs checks the flag inputs for missing inputs and sets them to
+// a default value if configured. If no default value is configured, the missing input
+// keys are returned as a slice.
+func (r *Render) processRequiredInputs(spec *model.Spec) []string {
```
drevell, line 398: I think `process` can be removed, it's one of those words that doesn't really add meaning IMO.

templates/model/spec.go
verbanicm: We can discuss this after some usage.

templates/commands/render.go
drevell, line 398: Would it make sense to combine this function with promptForInputs? I think the body of the loops can just be merged. It feels weird to have the defaulting logic spread across two functions.

templates/model/spec.go
drevell: I'm thinking *every* input is required unless it has a default.

drevell (changes requested):

templates/commands/render.go
verbanicm: I ilke it, will update

templates/commands/render.go
verbanicm: Seemed strange to bundle that logic into a function that shouldnt happen when no-prompt exists or no tty is avail.

drevell: Ah, I see.

templates/commands/render.go
```diff
@@ -323,6 +347,72 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
...
+	specInputs := make(map[string]any)
+	for _, v := range spec.Inputs {
+		specInputs[v.Name.Val] = struct{}{}
+	}
+
+	for key := range r.flagInputs {
+		if _, ok := specInputs[key]; !ok {
+			unknownInputs = append(unknownInputs, key)
+		}
+	}
```
sethvargo, line 363: Recommend sorting these keys so they are consistent:

```go
sort.Strings(unknownInputs)
```

templates/commands/render.go
```diff
@@ -323,6 +347,72 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
 	return nil
 }
 
+// processUnknownInputs checks for any unknown input flags and returns them in a slice.
+func (r *Render) processUnknownInputs(spec *model.Spec) []string {
+	unknownInputs := make([]string, 0)
```
sethvargo, line 352: I'd move this down after you've compiled `specInputs`, then you can efficiently allocate once:

```go
unknownInputs := make([]string, 0, len(r.flagInputs))
```

You might over-allocate, but you're guaranteed to only allocate once.

sethvargo (dismissed):

templates/commands/render.go
verbanicm: Created #65 to track this.

verbanicm:
Spoke with Seth offline, removing prompting for now due to complexity and other concerns.

PR #64, opened 2023-06-28
feat: allow skipping specific input files/dirs

### LINKED ISSUE #55: The spec file should not be included in the output
### TL;DR

Currently, if the user does an include action for path `'.'`, they'll get every file in the template. But they probably don't actually want spec.yaml, which is template metadata, and not really part of the template data.

Idea: remove the spec file from the template directory after loading it into memory.

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

drevell:
This adds the `skip` field in the `include` action. The lets template developers say "include this list of directories, but omit these subdirectories/files thereunder."

The template spec YAML file is now automatically skipped. If user does `include "."`, what they really want is "include every file in the template in the output, *except for this spec.yaml*".

Also added docs, see README.md for more details on how this works. This also adds some docs that were previously missing for `as`,`strip_prefix`, and `add_prefix` options on the `include` action.

Also refactored `copyRecursive()`. It started having too many parameters, so there's now a `copyParams` struct to group them together.

Closes #55.

### DIFF
--- a/templates/commands/render_action.go
+++ b/templates/commands/render_action.go
@@ -16,6 +16,7 @@ package commands
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -28,7 +29,9 @@ import (
 	"text/template"
 
 	"github.com/abcxyz/abc/templates/model"
+	"github.com/abcxyz/pkg/logging"
 	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
 )
 
 // Called with the contents of a file, and returns the new contents of the file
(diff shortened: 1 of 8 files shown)


templates/commands/render_action.go
```diff
@@ -190,65 +193,88 @@ func (n *unknownTemplateKeyError) Is(other error) bool {
 
 var templateKeyErrRegex = regexp.MustCompile(`map has no entry for key "([^"]*)"`)
 
-// "srcRoot" may be a file or directory. "pos" is only used for error messages.
-func copyRecursive(pos *model.ConfigPos, srcRoot, dstRoot string, rfs renderFS, overwrite, dryRun bool) (outErr error) {
-	return fs.WalkDir(rfs, srcRoot, func(path string, de fs.DirEntry, err error) error { //nolint:wrapcheck
+type copyParams struct {
```
sethvargo, line 196: Even though this struct isn't public, it would be great if we used the godoc-style syntax for documentation for consistency.

sethvargo (approved):

templates/commands/render_action.go
drevell: Will do in a follow-up so I don't lose my precious PR approval bit.

PR #69, opened 2023-06-28
feat: release automation

drevell:
Add a .goreleaser.yml that defines how the CLI binaries will be created, and add a GitHub workflow to run goreleaser on demand.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,0 +1,66 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+env:
+  # Global env vars for Go build.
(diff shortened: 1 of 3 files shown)


.goreleaser.yaml
```diff
@@ -0,0 +1,66 @@
...
+      - '-s'
+      - '-w'
+      - '-X={{ .ModulePath }}/internal/version.Name=abc'
+      - '-X={{ .ModulePath }}/internal/version.Version={{ .Version }}'
+      - '-X={{ .ModulePath }}/internal/version.Commit={{ .Commit }}'
+      - '-extldflags=-static'
+    goos:
+      - 'darwin'
+      - 'linux'
+      - 'windows'
```
sethvargo, line 43: Do we want to publish windows before we've validated any of the things work on Windows?

sethvargo (dismissed):
LGTM, but I'm not sure we want to publish windows binaries until we add them to the test matrix

.goreleaser.yaml
drevell: Removed the windows line.

drevell:
Related work: I filed issue #70 for tests not passing on Windows, and sent PR #71 to add mac to the test matrix.

.goreleaser.yaml
```diff
@@ -40,7 +40,7 @@ builds:
     goos:
       - 'darwin'
       - 'linux'
-      - 'windows'
+      # - 'windows'  # Coming some day
```
sethvargo, line 43: Lol ref #70? Or add a comment to #70 to also update the release automation (just so we don't forget)

sethvargo (approved):

.goreleaser.yaml
drevell: Updated #70 👍

PR #72, opened 2023-06-29
Add a template skeleton for data migration pipeline

chloechien:
- This minimal runnable version of template reads from a MySQL dump, transforms the data, and then write mutations to a Spanner table
- The template is named "data_migration_pipeline" instead of spanner specific naming, b/c in future we want the write action to be easily switched to other databases on template applied
- Tested with local spanner emulator


t/data_migration_pipeline/main.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+var (
+	input    = flag.String("input-csv-path", "", "The path of the input MySQL CSV dumped .")
+	database = flag.String("spanner-database", "", "The path of the output Spanner database.")
+	table    = flag.String("spanner-table", "", "The name of the output Spanner table.")
+)
+
+type DataModel struct {
+	/*
```
drevell, line 38: Some ideas, feel free to ignore:

- Use template inputs to fill this in: https://github.com/abcxyz/abc/#the-spec-file .  This could get awkward though. If you created an input variable named `struct_contents`, I think this might work, but I'm not sure:

    ```shell
    $ abc templates render --input=struct_contents=($cat <<EOF
    ID string
    Foo int
    Bar string
    EOF)
    ```

 - Add a `print` action (https://github.com/abcxyz/abc/#action-print) to the spec file asking the user to come here and fill this out. It might look like:

    ```shell
    - action: 'print'
      params:
        message:
          'Please go to the main.go and fill out the part where it says "your data model goes here"'
    ```

drevell (changes requested):
Partial review since I have a meeting, will resume later

t/data_migration_pipeline/spec.yaml
```diff
@@ -0,0 +1,24 @@
...
+apiVersion: 'cli.abcxyz.dev/v1alpha1'
+kind: 'Template'
+
+desc:
+  'An example template for a simple MySQL to Spanner migration pipeline.'
+steps:
+  - desc: 'Include some files and directories'
+    action: 'include'
+    params:
+      paths: ['.']
```
drevell, line 24: As mentioned above, I think it would be helpful to the user to add a `print` action here telling them what to do next.

t/data_migration_pipeline/README.md
```diff
@@ -0,0 +1,69 @@
+# Template: Data migration pipeline
+
+Simple Spanner data migration pipeline in Go, using Apache Beam to migrate from a MySQL CSV dump.
+
+To run this, please follow these steps:
+
+1. cd into an empty directory
+
+    ```shell
+    $ mkdir ~/template_tmp
```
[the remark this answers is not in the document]
chloechien: Sure, done!

t/data_migration_pipeline/main.go
chloechien: Thank you Dave, print strategy sgtm! Added.
I'm leaning towards asking users to add data model/parsing in code, to leverage the benefit of type validations and formatters in golang. In future we'd like the template to be able to replace these with user input:
- Output database
- Input format
- slicing

t/data_migration_pipeline/spec.yaml
chloechien: Done.

chloechien:
Thanks for the prompt review!

drevell (dismissed):

chloechien:
When update branch with latest changes, there is a unsigned commit. Instead of authoring that commit on behalf of me so I could sign that, will create a new Pr instead.

PR #76, opened 2023-06-29
Create dest dir if nonexistent, rather than failing

drevell:
There are no new calls to mkdir in this PR because we already have code that creates directories as needed. We just needed to remove a validation check.

Previously, we enforced that the destination dir already exists. This was intended to support the user journey where we're installing into the root of a git repo. However, a new user journey has emerged where we're installing a new directory into an existing git repo. In this case it's annoying to pre-create the directory and/or hackily precede every path in the spec file with a subdirectory name.

### DIFF
--- a/templates/commands/render.go
+++ b/templates/commands/render.go
@@ -328,7 +328,7 @@ func (r *Render) realRun(ctx context.Context, rp *runParams) (outErr error) {
 	}
 
 	// Commit the contents of the scratch directory to the output directory. We
-	// first do a dry-run to check that the copy is likely to  succeed, so we
+	// first do a dry-run to check that the copy is likely to succeed, so we
 	// don't leave a half-done mess in the user's dest directory.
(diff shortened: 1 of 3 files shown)


templates/commands/render.go
```diff
@@ -540,13 +540,13 @@ func destOK(fs fs.StatFS, dest string) error {
 	fi, err := fs.Stat(dest)
 	if err != nil {
 		if errors.Is(err, os.ErrNotExist) {
-			return fmt.Errorf("the destination directory doesn't exist: %q", dest)
+			return nil
```
verbanicm, line 543: Should we document that its ok to not exists because the render process will generate it?

verbanicm (approved):

templates/commands/render.go
drevell: Done.

verbanicm (approved):

sethvargo (approved):

PR #71, opened 2023-06-29
Run tests on linux and mac

drevell:
Windows isn't yet in the test matrix because test currently don't pass on Windows (which is issue #70).

Issue #70, opened 2023-06-29
Support windows


drevell:
### TL;DR

Currently, tests fail on windows. Let's fix them.

The test logs are attached. The failures seem mostly related to file permissions.

[logs_184.zip](https://github.com/abcxyz/abc/files/11901708/logs_184.zip)

Tasks:

 - Investigate each test failure and update either the app logic or test logic to get each one to pass
 - Update test matrix in `.github/workflows` to include windows
 - Update `.goreleaser.yaml` to include windows

How to test on windows: use the [drevell/win-test branch](https://github.com/abcxyz/abc/tree/drevell/win-test) or fork it into your own branch. This branch includes a [workflow that runs go tests on a windows VM](https://github.com/abcxyz/abc/blob/drevell/win-test/.github/workflows/win-test-temporary.yml).

pdewilde:
Can our windows support just print out "please just use WSL"

PR #78, opened 2023-06-29
Run tests on linux and mac (2nd PR attempt)

drevell:
Windows isn't yet in the test matrix because test currently don't pass on Windows (which is issue #70).(https://github.com/abcxyz/abc/pull/71).

This is a recreation of PR #71 which got into a stuck workflow state during a GitHub outage.


verbanicm (approved):
Looks like you need to update your required status checks.

sailorlqh (approved):

PR #81, opened 2023-06-29
Add support for Windows

sethvargo:
Fixes https://github.com/abcxyz/abc/issues/70

### DIFF
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -37,8 +37,13 @@ jobs:
   go_test:
     uses: 'abcxyz/pkg/.github/workflows/go-test.yml@main' # ratchet:exclude
     strategy:
+      # allow all os suites to complete, even if one fails
+      fail-fast: false
       matrix:
-        os: ['ubuntu-latest', 'macos-latest'] # TODO: add 'windows-latest' once it's known to work (issue #70)
+        os:
+          - 'macos-latest'
+          - 'ubuntu-latest'
+          - 'windows-latest'
     with:
       go_version: '1.20'
       os: '${{ matrix.os }}'
--- a/templates/commands/render_action_include_test.go
+++ b/templates/commands/render_action_include_test.go
@@ -260,6 +260,9 @@ func TestActionInclude(t *testing.T) {
 
 			ctx := logging.WithLogger(context.Background(), logging.TestLogger(t))
 
+			// Convert to OS-specific paths
+			convertKeysToPlatformPaths(tc.templateContents, tc.wantScratchContents)
+
 			tempDir := t.TempDir()
 			templateDir := filepath.Join(tempDir, templateDirNamePart)
 			scratchDir := filepath.Join(tempDir, scratchDirNamePart)
--- a/templates/commands/render_action_include_test.go
+++ b/templates/commands/render_action_include_test.go
@@ -284,12 +284,12 @@ func TestActionInclude(t *testing.T) {
 			}
 
 			gotTemplateContents := loadDirContents(t, filepath.Join(tempDir, templateDirNamePart))
-			if diff := cmp.Diff(gotTemplateContents, tc.templateContents); diff != "" {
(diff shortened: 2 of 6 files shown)


.github/workflows/lint-test.yml
```diff
@@ -37,8 +37,12 @@ jobs:
   go_test:
     uses: 'abcxyz/pkg/.github/workflows/go-test.yml@main' # ratchet:exclude
     strategy:
+      fail-fast: false
```
sethvargo, line 40: This is required - otherwise the linux tests failing will cancel any in-flight tests on windows.

templates/commands/render_test.go
```diff
@@ -34,6 +35,27 @@ import (
 	"github.com/hashicorp/go-getter/v2"
 )
 
+// cmpFileMode is a cmp option that handles the conflict between Unix and
```
sethvargo, line 38: 🐉

templates/commands/render_test.go
```diff
@@ -478,47 +504,42 @@ func TestSafeRelPath(t *testing.T) {
...
+			in:   slash + "a",
 			want: "a",
 		},
 		{
 			name: "leading_slash_with_more_dirs",
-			in:   "/a/b/c",
-			want: "a/b/c",
-		},
-		{
-			name: "leading_slash_with_more_dirs",
```
sethvargo, line 500: This was a duplicate test with the one above

templates/commands/render_test.go
```diff
@@ -536,11 +557,11 @@ func TestSafeRelPath(t *testing.T) {
 
 			got, err := safeRelPath(nil, tc.in)
 
-			if got != tc.want {
-				t.Errorf("safeRelPath(%q)=%q, want %q", tc.in, got, tc.want)
+			if got, want := got, tc.want; got != want {
+				t.Errorf("safeRelPath(%s): expected %q to be %q", tc.in, got, want)
```
sethvargo, line 561: I got really confused by the error output, so I changed this

.github/workflows/lint-test.yml
drevell: 👍 

Perhaps add this comment as a comment in the file for our descendants.

templates/commands/render_action_include_test.go
```diff
@@ -76,14 +76,14 @@ func TestActionInclude(t *testing.T) {
 				Paths: modelStrings([]string{"{{.my_dir}}/{{.my_file}}"}),
 			},
 			templateContents: map[string]modeAndContents{
-				"foo/bar.txt": {0o600, "file contents"},
+				filepath.Join("foo", "bar.txt"): {0o600, "file contents"},
```
drevell, line 79: As we just discussed offline, it would be much easier on the eyes in my opinion to leave these as forward slashes and translate them to filepath.Separator below.

templates/commands/render_test.go
drevell: 👍 👍

templates/commands/render_test.go
```diff
@@ -34,6 +35,27 @@ import (
...
+	// Furthermore, there's no reliable way to know whether the executive bit is
+	// set[2], so we ignore it.
+	//
+	// I tried doing fancy bitmasking stuff here, but apparently umasks on Windows
+	// are undefined(?), so we've resorted to substring matching - hooray.
+	//
+	// [1]: https://medium.com/@MichalPristas/go-and-file-perms-on-windows-3c944d55dd44
+	// [2]: https://github.com/golang/go/issues/41809
+	if runtime.GOOS == "windows" {
+		// "-rw-r--r--"
```
drevell, line 52: Super-nit: the very helpful comment above explains that the possible modes are 0666 and 0444, but this example is neither (0644).

templates/commands/render_test.go
drevell: Ack

templates/commands/render_test.go
drevell: SGTM

drevell (changes requested):

drevell (approved):

PR #83, opened 2023-06-29
Add default log level and mode

drevell:
The main point is that we want human-friendly log output by default, not JSON output which is currently the default. By setting ABC_LOG_MODE=dev, we accomplish that. I tested locally and confirmed this.

There was some related conversation about an abandoned alternative approach in https://github.com/abcxyz/pkg/pull/146. We chose not to do that, and did this PR instead.

Also add docs about logging, which had been overlooked.


sethvargo (dismissed):

verbanicm (dismissed):

README.md
```diff
@@ -69,6 +69,15 @@ Examples of template locations:
   by the template but are not supplied by `--inputs`. This flag will skip such
   prompting and just fail if there are any missing inputs.
 
+#### Logging
+
+Use the environment variable `ABC_LOG_MODE` to configure JSON logging. logging.
```
yolocs, line 74:
```suggestion
Use the environment variable `ABC_LOG_MODE` to configure JSON logging.
```

yolocs (dismissed):

README.md
drevell: Thanks!

yolocs (approved):

PR #84, opened 2023-06-30
Add data migration pipeline

chloechien:
- This minimal runnable version of template reads from a MySQL dump, transforms the data, and then write mutations to a Spanner table
- The template is named "data_migration_pipeline" instead of spanner specific naming, b/c in future we want the write action to be easily switched to other databases on template applied
- Tested with local spanner emulator


chloechien:
This PR is already approved in https://github.com/abcxyz/abc/pull/72, however when update branch with latest changes, there is a unsigned commit from other branches. Instead of re-authoring that commit on behalf of me so I could sign that, I feel the easiest way is to create a new Pr instead.

t/data_migration_pipeline/main_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"testing"
+
+	"github.com/apache/beam/sdks/v2/go/pkg/beam"
+	"github.com/apache/beam/sdks/v2/go/pkg/beam/testing/passert"
+	"github.com/apache/beam/sdks/v2/go/pkg/beam/testing/ptest"
+)
+
+func TestEmitResult(t *testing.T) {
```
pdewilde, line 25: Could we automate the setup of the spanner test db using `pkg/containertest`?

pdewilde (approved):

t/data_migration_pipeline/main_test.go
pdewilde: Approving as I don't think its a blocker, but would be nice if we could test the template in CI that way.

drevell (approved):

t/data_migration_pipeline/main_test.go
chloechien: Sure we could do that. By achieving an integration test, we'll do
- Docker run spanner emulator
- Docker run tests with integration env set up, a example: https://github.com/apache/beam/blob/master/sdks/go/test/integration/io/spannerio/test_helpers.go
That might be too much for a boilerplate template, let's add that later if an integration tests is needed for this pipeline, or we can trust beam io package and make this code piece light weight without integration testing.

chloechien:
Thank you both!

PR #88, opened 2023-07-06
Add deployment workflow to rest_server

kevya-google:
Adds Artifact Registry image pushing + Cloud Run deployment based on Sandwich workflows, also updates spec.yaml file to work with .env config to take in user input values based on infrastructure setup.


t/rest_server/.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,106 @@
...
+        run: |
+          # Filter out comment lines, empty lines, and lines with only whitespace.
+          grep -v '^[[:space:]]*#\|^$\|^\s*$' ${{ env.BOOTSTRAP_ENV_FILE }} >> ${GITHUB_ENV}
+      - name: 'Setup Go'
+        uses: 'actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9' # ratchet:actions/setup-go@v4
+        with:
+          go-version: '1.20.2'
+      - name: 'Build Go binary'
+        run: |
+          go mod tidy;
```
drevell, line 50: I don't think this `go mod tidy` line should be here. This workflow should build exactly what's committed, without doing anything clever to add/remove dependencies. If the go.mod/go.sum files aren't buildable, then I think it's better to fail the build rather than sweep the problem under the rug.

t/rest_server/.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,106 @@
...
+  workflow_call:
+
+concurrency:
+  group: '${{ github.workflow }}-${{ github.head_ref || github.ref }}'
+  cancel-in-progress: true
+
+env:
+  # This should be the only env variable defined in this file. Define all other
+  # variables in the bootstrap file.
+  BOOTSTRAP_ENV_FILE: '.github/config/autopush.env'
```
drevell, line 28: Maybe add some kind of TODO or other comment here that we expect to add more environments

t/rest_server/spec.yaml
```diff
@@ -14,11 +14,43 @@
 
 apiVersion: 'cli.abcxyz.dev/v1alpha1'
 kind: 'Template'
-
-desc:
-  'An example template for a simple HTTP/JSON REST server.'
+desc: 'An example template for a simple HTTP/JSON REST server.'
+inputs:
+  - name: 'automation_service_account'
+    desc: 'Automation Service Account (ex: [account_name]@[project_name].iam.gserviceaccount.com)'
```
drevell, line 20: Should this be "project_id" rather than "project_name"?

t/rest_server/spec.yaml
```diff
@@ -14,11 +14,43 @@
...
+  - name: 'wif_provider'
+    desc: 'Workload Identity Federation Provider (ex: projects/[project_id]/locations/global/workloadIdentityPools/[WIF_pool_name]/providers/[provider_name])'
+  - name: 'ar_repository'
+    desc: 'Artifact Registry Repository (repo name in AR)'
```
drevell, line 24: Could we provide an example name for the AR name? That will make it clear whether this is supposed to be some/long/path/my-images vs just my-images

t/rest_server/spec.yaml
```diff
@@ -14,11 +14,43 @@
...
+  - name: 'ar_location'
+    desc: 'Artifact Registry Location (ex: us-docker.pkg.dev)'
+  - name: 'cr_service'
+    desc: 'Cloud Run Service (service name in CR)'
```
drevell, line 28: Same here

drevell (changes requested):

drevell:
Only minor requests

t/rest_server/.github/workflows/deploy.yaml
kevya-google: Removed `go mod tidy`, this is a good point.

t/rest_server/.github/workflows/deploy.yaml
kevya-google: Added comment for future environments to be added.

t/rest_server/spec.yaml
kevya-google: Changed to project_id.

t/rest_server/spec.yaml
kevya-google: Added ci-images as an example.

t/rest_server/spec.yaml
kevya-google: Added hello-jupiter-api as an example.

t/rest_server/spec.yaml
```diff
@@ -14,11 +14,43 @@
 
 apiVersion: 'cli.abcxyz.dev/v1alpha1'
 kind: 'Template'
-
-desc:
-  'An example template for a simple HTTP/JSON REST server.'
+desc: 'An example template for a simple HTTP/JSON REST server.'
```
drevell, line 17: Can we remove "example" from this? To me, that makes it sound like it's for people learning about templates rather than being a real template that we expect people to use.

drevell (dismissed):

t/rest_server/.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,106 @@
...
+        with:
+          registry: '${{ env.AR_LOCATION }}'
+          username: 'oauth2accesstoken'
+          password: '${{ steps.auth.outputs.access_token }}'
+      - name: 'Push Docker Image'
+        run: |
+          docker push ${{ env.IMAGE }}
+
+  deploy-service:
+    needs: ['image-release']
```
verbanicm, line 74: Prefer expanded syntax for future readability:

```suggestion
needs:
  - 'image-release'
```

t/rest_server/.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,106 @@
...
+          # Filter out comment lines, empty lines, and lines with only whitespace.
+          grep -v '^[[:space:]]*#\|^$\|^\s*$' ${{ env.BOOTSTRAP_ENV_FILE }} >> ${GITHUB_ENV}
+      - name: 'Setup Go'
+        uses: 'actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9' # ratchet:actions/setup-go@v4
+        with:
+          go-version: '1.20.2'
+      - name: 'Build Go binary'
+        run: |
+          CGO_ENABLED=0 GOOS=linux go build -o restserver .
+      - name: 'Build Docker Image'
```
verbanicm, line 52: I would move the build go binary and build docker image below the auth step, it would be more efficient to fail fast with auth rather than spending time building the container only to fail in the auth step.

t/rest_server/.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,106 @@
...
+
+concurrency:
+  group: '${{ github.workflow }}-${{ github.head_ref || github.ref }}'
+  cancel-in-progress: true
+
+env:
+  # This should be the only env variable defined in this file. Define all other
+  # variables in the bootstrap file.
+  # More environments are expected to be added in the future.
+  BOOTSTRAP_ENV_FILE: '.github/config/autopush.env'
```
verbanicm, line 29: I imagine that more often than not, there will be more than one environment. It would be good to include at least 2 environments in this example. Can we add a production env just for demonstration sake?

Are you planning to have the environment name be an input value?

t/rest_server/spec.yaml
kevya-google: Removed "example".

t/rest_server/.github/workflows/deploy.yaml
kevya-google: Moved below auth.

t/rest_server/.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,106 @@
...
+      id-token: 'write'
+    name: 'Release Image'
+    runs-on: 'ubuntu-latest'
+    steps:
+      - name: 'Checkout Repo'
+        uses: 'actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab' # ratchet:actions/checkout@v3
+      - name: 'Load variables from bootstrap .env file'
+        run: |
+          # Filter out comment lines, empty lines, and lines with only whitespace.
+          grep -v '^[[:space:]]*#\|^$\|^\s*$' ${{ env.BOOTSTRAP_ENV_FILE }} >> ${GITHUB_ENV}
```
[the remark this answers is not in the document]
kevya-google: Oh, good to know! Replaced with re-usable version.

t/rest_server/.github/workflows/deploy.yaml
kevya-google: Environment as input value, conditional logic demonstrated:
Autopush
https://github.com/drevell/abc-test-1/actions/runs/5523161977/jobs/10073697846#step:3:1
Prod
https://github.com/drevell/abc-test-1/actions/runs/5523171101/jobs/10073720473#step:4:1

kevya-google: Remove production .env for now until we figure out a good variable substitution scheme

t/rest_server/.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,127 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: 'deploy'
+
+on:
+  workflow_dispatch:
+    inputs:
+      deploymentEnvironment:
```
drevell, line 20: The style for github workflow files seems to avoid snakeCase. words_with_underscores looks like the style. https://github.com/search?q=org%3Aabcxyz+workflow_call&type=code

t/rest_server/.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,127 @@
...
+concurrency:
+  group: '${{ github.workflow }}-${{ github.head_ref || github.ref }}'
+  cancel-in-progress: true
+
+env:
+  # This should be the only env variable defined in this file. Define all other
+  # variables in the bootstrap file.
+  # More environments are expected to be added in the future.
+  AUTOPUSH_ENV_FILE: '.github/config/autopush.env'
+  PRODUCTION_ENV_FILE: '.github/config/production.env'
```
drevell, line 42: There seems to be an inconsistency here. If production is an environment, then I think it should be part of the "choice" input. If it's not an environment, then maybe this line should be removed.

t/rest_server/.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,127 @@
...
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    name: 'Release Image'
+    runs-on: 'ubuntu-latest'
+    steps:
+      - name: 'Checkout Repo'
+        uses: 'actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab' # ratchet:actions/checkout@v3
+      - name: 'Load variables from Autopush .env file'
+        if: "${{ inputs.deploymentEnvironment == 'autopush' }}"
```
drevell, line 55: This "if" seems to be unnecessary. Or is it future-proofing?

t/rest_server/.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,127 @@
...
+          registry: '${{ env.AR_LOCATION }}'
+          username: 'oauth2accesstoken'
+          password: '${{ steps.auth.outputs.access_token }}'
+      - name: 'Setup Go'
+        uses: 'actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9' # ratchet:actions/setup-go@v4
+        with:
+          go-version: '1.20.2'
+      - name: 'Build Go binary'
+        run: |
+          CGO_ENABLED=0 GOOS=linux go build -o restserver .
```
drevell, line 78: Is the GOOS necessary? SInce this is running on ubuntu-latest, it will automatically build for Linux. We don't seem to do this elsewhere: https://github.com/search?q=org%3Aabcxyz%20%22go%20build%22&type=code

t/rest_server/.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,127 @@
...
+    name: 'Release Image'
+    runs-on: 'ubuntu-latest'
+    steps:
+      - name: 'Checkout Repo'
+        uses: 'actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab' # ratchet:actions/checkout@v3
+      - name: 'Load variables from Autopush .env file'
+        if: "${{ inputs.deploymentEnvironment == 'autopush' }}"
+        uses: 'abcxyz/pkg/.github/actions/override-workflow-variables@main'
+        with:
+          env_file: '${{ env.AUTOPUSH_ENV_FILE }}'
```
drevell, line 58: Can we simplify away the AUTOPUSH_ENV_FILE variable just by loading `.github/config/${{inputs.deployment_environment}}.env`?

That eliminates the need for:
 - This "if"
 - The `*_ENV_FILE` variables
 - Separate override-workflow-variables steps for each environment

drevell (changes requested):

t/rest_server/.github/workflows/deploy.yaml
kevya-google: Good idea, changed to remove env variable.

t/rest_server/.github/workflows/deploy.yaml
kevya-google: Removed GOOS, unsure if it's needed.

t/rest_server/.github/workflows/deploy.yaml
kevya-google: Removed if per your suggestion below.

t/rest_server/.github/workflows/deploy.yaml
kevya-google: Removed both env files per your suggestion below.

t/rest_server/.github/workflows/deploy.yaml
kevya-google: Changed to snake_case

drevell (approved):

PR #90, opened 2023-07-07
Refactor copyRecursive to add callback

drevell:
This PR lays the groundwork for the "destination includes" feature coming in a following PR.

We add a callback to the copyRecursive function that lets the handling of each file be specified as it's crawled (overwrite or skip). In the following PR we'll add a new option to back up each file if it's being overwritten. It also informs the caller of the names of each crawled file.

Also, fix a bug found while testing this change: add the `O_TRUNC` option when overwriting a file. When overwriting a longer file with a shorter file, we were leaving some of the old file contents at the end of the file.

Also some misc cleanups:

 - rename `from`->`src` and `to`->`dst` in render_action_test.go. We agreed on this renaming but missed some spots.

 - factor out `copyFile()` from the anonymous function inside copyRecursive(). It was getting too big, and we'll need it in the next PR when we want to copy files one at a time.

 - Add a `string_replacement` step to the spec file in TestRealRun to make the test more realistic.

### DIFF
--- a/templates/commands/render_action.go
+++ b/templates/commands/render_action.go
@@ -210,9 +210,6 @@ type copyParams struct {
 	// dryRun skips actually copy anything, just checks whether the copy would
 	// be likely to succeed.
 	dryRun bool
-	// skip is a set of paths not to be copied. These paths are relative to
-	// srcRoot. This lets certain special files be excluded from the output.
-	skip []string
 	// visitor is an optional function that will be called for each file in the
 	// source, to allow customization of the copy operation on a per-file basis.
 	visitor copyVisitor
--- a/templates/commands/render_action.go
+++ b/templates/commands/render_action.go
@@ -31,7 +31,6 @@ import (
 	"github.com/abcxyz/abc/templates/model"
 	"github.com/abcxyz/pkg/logging"
 	"golang.org/x/exp/maps"
-	"golang.org/x/exp/slices"
 )
 
 // Called with the contents of a file, and returns the new contents of the file
@@ -137,6 +136,8 @@ func parseGoTmpl(tpl string) (*template.Template, error) {
 	return template.New("").Funcs(templateFuncs()).Option("missingkey=error").Parse(tpl) //nolint:wrapcheck
 }
 
+var templateKeyErrRegex = regexp.MustCompile(`map has no entry for key "([^"]*)"`)
+
 // pos may be nil if the template is not coming from the spec file and therefore
 // there's no reason to print out spec file location in an error message. If
 // template execution fails because of a missing input variable, the error will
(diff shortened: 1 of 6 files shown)


templates/commands/render_action.go
```diff
@@ -211,87 +210,120 @@ type copyParams struct {
...
 
-		if _, err := io.Copy(wf, rf); err != nil {
-			return fmt.Errorf("Copy(): %w", err)
-		}
+func copyFile(pos *model.ConfigPos, rfs renderFS, src, dst string, mode fs.FileMode, dryRun bool) (outErr error) {
+	readFile, err := rfs.Open(src)
+	if err != nil {
+		return model.ErrWithPos(pos, "Open(): %w", err) //nolint:wrapcheck
+	}
+	defer func() { outErr = errors.Join(outErr, readFile.Close()) }()
```
pdewilde, line 310: That's neat, never though of altering a return type using a defer

pdewilde (approved):

templates/commands/render_action.go
drevell: Yeah. It doesn't feel super clean to me but afaik it's the only way to return an error from a deferred function.

templates/commands/render_action_include.go
```diff
@@ -35,12 +37,13 @@ func actionInclude(ctx context.Context, inc *model.Include, sp *stepParams) erro
 		return err
 	}
 
-	skipStrs := make([]string, len(inc.Skip))
-	for i, skip := range inc.Skip {
-		skipStrs[i], err = parseAndExecuteGoTmpl(skip.Pos, skip.Val, sp.inputs)
+	skip := make(map[string]bool, len(inc.Skip))
```
sethvargo, line 40: Nit: `struct{}` instead of `bool` since this is just set membership.

templates/commands/render_action_include.go
```diff
@@ -71,15 +74,19 @@ func actionInclude(ctx context.Context, inc *model.Include, sp *stepParams) erro
...
-		absSrc := filepath.Join(sp.templateDir, walkRelPath)
+		// By default, we copy from the template directory. We also support
+		// grabbing files from the destination directory, so we can modify files
+		// that already exist in the destination.
+		fromDir := sp.templateDir
+		absSrc := filepath.Join(fromDir, walkRelPath)
 		absDst := filepath.Join(sp.scratchDir, relDst)
 
-		skip := skipStrs
+		skipNow := maps.Clone(skip)
```
sethvargo, line 84: Why do we need to clone?

drevell: We just want to modify the map for the current iteration of the loop, without affecting future iterations.

This supports the case where the spec files is `include`ing multiple paths. We only want to add this file to the blocklist in the case where the current source dir is the template root. If we're also `include`ing from other dirs that are not the template root, they should get the extra blocklist entry.

sethvargo (approved):

PR #92, opened 2023-07-07
Use version from build metadata vs hardcoded

drevell:
This guarantees that `abc -version` won't get out of sync with the actual release.

Also fix a trivial unrelated problem where the default log level appeared redundantly in two places and used `warning` instead of the correct `warn`.

I tried logging the version number as a debug-level log message on abc startup, but the abcxyz cli package does not allow access to flags from main(), only from subcommand Run() functions.

### DIFF
--- a/cmd/abc/abc.go
+++ b/cmd/abc/abc.go
@@ -21,14 +21,15 @@ import (
 	"os/signal"
 	"syscall"
 
+	"github.com/abcxyz/abc/internal/version"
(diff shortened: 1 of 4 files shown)


pdewilde (approved):

yolocs:
Where is the internal package?

drevell:
> Where is the internal package?

🙃 fixed

yolocs (approved):

PR #93, opened 2023-07-07
Add tests for abc.go

drevell:
Also fix a place where os.Stdout was hardcoded into render.go. It should use the output stream provided by the cli package for testability.

### DIFF
--- a/cmd/abc/abc.go
+++ b/cmd/abc/abc.go
@@ -17,7 +17,6 @@ package main
 import (
 	"context"
 	"fmt"
-	"io"
 	"os"
 	"os/signal"
 	"syscall"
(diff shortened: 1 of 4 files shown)


cmd/abc/abc.go
```diff
@@ -51,13 +52,17 @@ func main() {
...
 
-func realMain(ctx context.Context) error {
-	return rootCmd().Run(ctx, os.Args[1:]) //nolint:wrapcheck
+// argsTrimmed should be os.Args[1:], or the equivalent for testing.
+func realMain(ctx context.Context, argsTrimmed []string, stdout, stderr io.Writer) error {
+	cmd := rootCmd()
+	cmd.SetStdout(stdout)
+	cmd.SetStderr(stderr)
+	return cmd.Run(ctx, argsTrimmed) //nolint:wrapcheck
 }
```
sethvargo, line 68: I don't think you need to do this. In tests, you can ask the Command to create a pipe, which will override stdin, stdout, and stderr to buffers: https://github.com/abcxyz/pkg/blob/f75fec6a9ad6e1b5d6b5a8fbb494f28644ad18af/cli/command_test.go#L308-L309

cmd/abc/abc_test.go
```diff
@@ -14,9 +14,81 @@
...
+			t.Parallel()
+
+			if len(tc.onlyGOOSes) > 0 && !slices.Contains(tc.onlyGOOSes, runtime.GOOS) {
+				t.Skipf("this subtest only runs if GOOS is one of %s", tc.onlyGOOSes)
+			}
+
+			ctx := context.Background()
+			stdout := &bytes.Buffer{}
+			stderr := &bytes.Buffer{}
+			err := realMain(ctx, tc.args, stdout, stderr)
```
sethvargo, line 82: This could be replaced with `rootCmd().Pipe()`

cmd/abc/abc.go
drevell: Fixed, this is cleaner, thanks.

sethvargo (approved):

PR #96, opened 2023-07-11
feat: `include` files from destination dir

drevell:
With this change, spec files can now `include` a file from the destination dir rather than the template dir. This lets an existing file be modified; before, a template could only overwrite an existing file without regard for its contents.

When a file is included from the destination, it goes on an allowlist of files that are allowed to be overwritten. This stands in contrast to default policy of refusing to overwrite existing files (unless `--force-overwrite`). It would be pointless to take a file from the destination and then complain about overwriting it later; replacing it was the whole point of including.

Before a file is overwritten in the destination directory, the old version of the file is backed up in `~/.abc/backups/$timestamp/`.

### DIFF
--- a/templates/commands/render.go
+++ b/templates/commands/render.go
@@ -79,6 +79,7 @@ type renderFS interface {
 
 	// These methods correspond to methods in the "os" package of the same name.
 	MkdirAll(string, os.FileMode) error
+	MkdirTemp(string, string) (string, error)
 	OpenFile(string, int, os.FileMode) (*os.File, error)
 	ReadFile(string) ([]byte, error)
 	RemoveAll(string) error
@@ -199,6 +200,10 @@ func (r *realFS) MkdirAll(name string, perm os.FileMode) error {
 	return os.MkdirAll(name, perm) //nolint:wrapcheck
 }
 
+func (r *realFS) MkdirTemp(dir, pattern string) (string, error) {
+	return os.MkdirTemp(dir, pattern) //nolint:wrapcheck
+}
+
 func (r *realFS) Open(name string) (fs.File, error) {
 	return os.Open(name) //nolint:wrapcheck
 }
@@ -257,15 +262,11 @@ func (r *Render) Run(ctx context.Context, args []string) error {
 	if err != nil {
 		return fmt.Errorf("os.UserHomeDir: %w", err)
 	}
-	randToken, err := randU64()
-	if err != nil {
-		return err
-	}
 	backupDir := filepath.Join(
 		homeDir,
 		".abc",
 		"backups",
-		fmt.Sprintf("%d-%d", time.Now().Unix(), randToken))
+		fmt.Sprint(time.Now().Unix()))
 
 	return r.realRun(ctx, &runParams{
 		backupDir:    backupDir,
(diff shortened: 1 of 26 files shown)


templates/commands/render.go
```diff
@@ -250,7 +254,14 @@ func (r *Render) Run(ctx context.Context, args []string) error {
 		return fmt.Errorf("os.Getwd(): %w", err)
 	}
 
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return fmt.Errorf("os.UserHomeDir: %w", err)
+	}
+	backupDir := filepath.Join(homeDir, ".abc", "backups", strconv.FormatInt(time.Now().Unix(), 10))
```
sethvargo, line 261: We might want to use:

```go
backupsDir := filepath.Join(homeDir, ".abc", "backups")
os.MkdirTemp(backupsDir, "*")
```

or: 

```go
now := time.Now().UTC().Unix()
backupsDir := filepath.Join(homeDir, ".abc", "backups")
os.MkdirTemp(backupsDir, fmt.Sprintf("%d-*", now))
```

yolocs (approved):

templates/commands/render.go
sethvargo: So, the reason to use `os.Mktempdir` is that it will _guarantee_ there's no conflict: https://cs.opensource.google/go/go/+/refs/tags/go1.20.6:src/os/tempfile.go;l=80-111

In addition to the randomness and locking, it actually retries if there's a conflict.

I agree we shouldn't make the backup dir until it's needed; can we move this closer inline to when we know it'll be required then?

drevell: Done. It's a little ugly because it's awkward to fake in tests.

If you want to go back to the original implementation which was simple but had a miniscule probabilty of collision in the case where two 64-bit random numbers collide within the same second, that's also good with me.

yolocs (approved):

drevell:
Seth, I auto-merged this one without thinking it through, I can address any further comments in a followup.

templates/commands/render_test.go
```diff
@@ -505,10 +592,29 @@ steps:
...
+			if diff := cmp.Diff(gotBackupContents, tc.wantBackupContents, cmpopts.EquateEmpty()); diff != "" {
+				t.Errorf("backups directory contents were not as expected (-got,+want): %s", diff)
+			}
 		})
 	}
 }
 
+// Since os.MkdirTemp adds an extra random token, we strip it back out to get
+// determistic results.
+func stripFirstPathElem(m map[string]string) map[string]string {
```
sethvargo, line 607: Hmm yea I agree this is a bit weird, but wouldn't we have the same issue if we used the randint?

drevell: It's a lot simpler with the randint. You just generate the path you *would* use at startup time. You do this at the top-level, in realRun(). Then when it's time to create a backup file, you just MkdirAll the folder to create-if-not-exists. For the lower-level functions like copyRecursive and TestCopyRecursive, the backup directory is just a string to them. The tests just write to t.TempDir(), and the real prod code takes a directory name from the higher level that happens to have a randint in it.

drevell: Let me try to explain that better: MkdirAll is idempotent, using a randint that was generated at startup. MkdirTemp isn't idempotent, so we have to ensure it only runs once.

sethvargo: Hmm okay - I'm fine if we:

1. Document this super duper well
2. Return a very specific error if the directory already exists

PR #98, opened 2023-07-11
Add flags to go-template context for print action

### LINKED ISSUE #85: Add the ability to reference CLI flags in template expressions
### TL;DR

Add the ability to reference CLI flags in template expressions.

This could be adding a `var flags map[string]string` to the template engine so end users could reference them in template expressions like `{{.flags["dest"]}}`.

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

drevell:
Fixes #85.

To be able to pass the entire set of flags around, I factored them out of the Render struct into a new dedicated struct, `renderFlags`.

Q: Why only expose some flags? Why expose `--dest` but not `--force-overwrite`?

A: API parsimony. We should limit the things we expose to just the things which we know are useful, since we'll have to support this in the future. We can always expose more flags later.

Q: Why only expose these flags to the print action?

A: We want to avoid the temptation for template authors to be inappropriately clever. We think we've provided a single correct way to do things, and we don't want people macking weird hacks unnecessarily. We want template rendering to be deterministic, and not change in surprising ways depending on flag values.

Since we're adding a new `{{.flags}}` namespace, we now forbid a template from declaring an input named `flags`, to avoid collisions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,6 +371,17 @@ Params:
 - `message`: the message to show. May use template expressions (e.g.
   `{{.my_input}}`).
 
+  The print action has special access to an extra template variable named
+  `flags` containing the values of some of the command line flags. This can be
+  useful to print a message like
+  `Template rendering is done, now please to go the {{.flags.dest}} directory and run a certain command.`
+  The available values are:
+
+  - `{{.flags.dest}}`: the value of the the `--dest` flag, e.g. `.`
+  - `{{.flags.source}}`: the template location that's being rendered, e.g.
+    `github.com/abcxyz/abc.git//t/my_template`
+  - `{{.flags.spec}}`: the value of the `--spec` flag, e.g. `spec.yaml`
+
 Example:
 
 ```yaml
--- a/templates/commands/render.go
+++ b/templates/commands/render.go
@@ -53,18 +53,34 @@ const (
 
 type Render struct {
 	cli.BaseCommand
+	flags renderFlags
 
 	testFS     renderFS
 	testGetter getterClient
+}
 
-	source             string
-	flagSpec           string
-	flagDest           string
-	flagGitProtocol    string
-	flagLogLevel       string
-	flagForceOverwrite bool
-	flagKeepTempDirs   bool
-	flagInputs         map[string]string
+type renderFlags struct {
+	// Positional arguments:
+	source string
+
+	// Flag arguments (--foo):
+	dest           string
+	gitProtocol    string
+	logLevel       string
+	forceOverwrite bool
+	inputs         map[string]string // these are just the --input values from flags; doesn't inclue values from config file or env vars
+	keepTempDirs   bool
+	spec           string
+}
+
+func (r *renderFlags) forTemplate() map[string]any {
+	// We only expose certain fields to go-templating; thse are the ones that we
+	// can imagine beneficial use cases for.
+	return map[string]any{
+		"dest":   r.dest,
+		"source": r.source,
+		"spec":   r.spec,
+	}
 }
 
 // Abstracts filesystem operations.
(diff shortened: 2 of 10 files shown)


templates/commands/render.go
```diff
@@ -53,18 +53,34 @@ const (
 
 type Render struct {
 	cli.BaseCommand
+	flags renderFlags
```
sethvargo, line 56: Do you want this to be a pointer? As there are more-and-more flags, it will get expensive to copy this.

drevell: I tried that, but liked it better as a value. Then the zero values are automatically present for unit tests. I'll pay the microsecond cost of copying.

drevell: I see what you mean: why copy all the flags here, if only a subset will be used? Wouldn't it be simpler just to copy a subset in the first place?

It's a little more complicated than that. There's other code that's not part of this PR that is also relevant.

Currently, we have two places in the "action" functions that use flags. The "include" action uses the "dest" flag, and this "print" action uses the "dest", "source", and "spec" flags. Now that we have two cases of accessing flags from an action function, I'm thinking about a clearn future-proof solution rather than having a union-of-subsets-of-flags-that-are-currently-needed-by-action-functions in the stepParams. By placing all the flags here and just pulling the subset that are needed in each action, it's clean and future proof. I'm anticipating needing more flags in our future action functions.

sethvargo (approved):

templates/commands/render_action.go
```diff
@@ -142,7 +142,7 @@ var templateKeyErrRegex = regexp.MustCompile(`map has no entry for key "([^"]*)"
 // there's no reason to print out spec file location in an error message. If
 // template execution fails because of a missing input variable, the error will
 // be wrapped in a unknownTemplateKeyError.
-func parseAndExecuteGoTmpl(pos *model.ConfigPos, tmpl string, inputs map[string]string) (string, error) {
+func parseAndExecuteGoTmpl[T any](pos *model.ConfigPos, tmpl string, inputs map[string]T) (string, error) {
```
yolocs, line 145: What if we just make inputs `map[string]any`?

README.md
```diff
@@ -371,6 +371,17 @@ Params:
 - `message`: the message to show. May use template expressions (e.g.
   `{{.my_input}}`).
 
+  The print action has special access to an extra template variable named
+  `flags` containing the values of some of the command line flags. This can be
+  useful to print a message like
+  `Template rendering is done, now please to go the {{.flags.dest}} directory and run a certain command.`
```
yolocs, line 377: This feels a little over engineered to me, given the current stage of this project. We can probably save this for later when there are things beyond `dest` we'd like to print.

yolocs: Not against wrapping flags to pass them around though.

templates/commands/render.go
```diff
@@ -53,18 +53,34 @@ const (
...
 
-	source             string
-	flagSpec           string
-	flagDest           string
-	flagGitProtocol    string
-	flagLogLevel       string
-	flagForceOverwrite bool
-	flagKeepTempDirs   bool
-	flagInputs         map[string]string
+func (r *renderFlags) forTemplate() map[string]any {
```
yolocs, line 76: Since this is for print action only, would it make more sense to let `render_action_print.go` take this responsibility? Let it retrieve values from the flags and build this map.

README.md
verbanicm: I have a use case for this for bootstrapping a new GCP org from the setup template.

yolocs: I can imagine it but I was thinking a more basic version where we just the relative path to point to a readme :)

templates/commands/render_action.go
drevell: I don't think the type system will allow that (passing a map[string]string as a map[string]any). Feel free to send me a patch if you can get it to work.

templates/commands/render.go
drevell: That's a good simplification, thanks. I'll send you a PR shortly.

README.md
drevell: I think it's good as-is. It's important to show helpful messages to the user.

PR #102, opened 2023-07-17
Add more log messages for template rendering

chloechien:
1. Move r.setLogEnvVars() before flag parse, so cli could correctly intake flag value
2. Add logs everywhere
example
```
INFO	commands/render.go:592	created temporary template directory at: /tmp/abc-template-copy-7745125156508404672
INFO	commands/render.go:593	copied source template github.com/abcxyz/abc.git//examples/templates/render/hello_jupiter into temporary directory /tmp/abc-template-copy-7745125156508404672
INFO	commands/render.go:346	created temporary scratch directory at: /tmp/abc-scratch-13218785992439717935
INFO	commands/render.go:393	created backup directory at my_home/.abc/backups/1689614607/2066743529
INFO	commands/render_action.go:380	backed up previous contents of main.go at my_home/.abc/backups/1689614607/2066743529/main.go
INFO	commands/render.go:604	removing all temporary directories (skip this with --keep-temp-dirs)
```

```
INFO	commands/render.go:592	created temporary template directory at: /tmp/abc-template-copy-5304447187202649298
INFO	commands/render.go:593	copied source template github.com/abcxyz/abc.git//examples/templates/render/hello_jupiter into temporary directory /tmp/abc-template-copy-5304447187202649298
INFO	commands/render.go:346	created temporary scratch directory at: /tmp/abc-scratch-8610268306021048948
DEBUG	commands/render_action.go:347	copyFile: from /tmp/abc-template-copy-5304447187202649298/main.go to /tmp/abc-scratch-8610268306021048948/main.go
DEBUG	commands/render.go:474	completed template action include
DEBUG	commands/render_action.go:93	walkAndModify: wrote modification to /tmp/abc-scratch-8610268306021048948/main.go
DEBUG	commands/render.go:474	completed template action string_replace
DEBUG	commands/render.go:409	template render (dry run) succeed
INFO	commands/render.go:393	created backup directory at my_home/.abc/backups/1689614632/3213846716
DEBUG	commands/render_action.go:347	copyFile: from main.go to my_home/.abc/backups/1689614632/3213846716/main.go
INFO	commands/render_action.go:380	backed up previous contents of main.go at my_home/.abc/backups/1689614632/3213846716/main.go
DEBUG	commands/render_action.go:347	copyFile: from /tmp/abc-scratch-8610268306021048948/main.go to main.go
DEBUG	commands/render.go:411	template render succeed
INFO	commands/render.go:604	removing all temporary directories (skip this with --keep-temp-dirs)
```

Closes #26

PR #103, opened 2023-07-17
Use structured logging instead

sethvargo:
Quick follow-up of https://github.com/abcxyz/abc/pull/102, this uses structured KV pairs instead of string interpolation.


chloechien (approved):
Thank you!

drevell (approved):

PR #104, opened 2023-07-18
Add an API for running Render programmatically

drevell:
This exposes the RenderFlags struct for external use, and adds the Render.RunWithFlags method.

This is PR 1 of 3-ish that will add a test framework for templates. The goal is to (1) allow template developers to test their templates and (2) allow abc CLI developers to verify that code changes haven't broken templates.

Reviewers, all of the interesting changes are in render.go, and the rest is just changed capitalization and renaming.

### DIFF
--- a/templates/commands/render.go
+++ b/templates/commands/render.go
@@ -56,24 +56,50 @@ const (
 
 type Render struct {
 	cli.BaseCommand
-	flags renderFlags
+	flags *RenderFlags
 
 	testFS     renderFS
 	testGetter getterClient
 }
 
-type renderFlags struct {
+// RenderFlags describes what template to render and how.
+type RenderFlags struct {
 	// Positional arguments:
-	source string
+
+	// Source is the location of the input template to be rendered.
+	//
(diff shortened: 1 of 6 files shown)


templates/commands/render.go
```diff
@@ -235,10 +261,21 @@ func (r *realFS) WriteFile(name string, data []byte, perm os.FileMode) error {
...
+	rf := &RenderFlags{}
+	if err := rf.Parse(args); err != nil {
 		return err
 	}
 
+	return r.RunWithFlags(ctx, rf)
+}
+
+// RunWithFlags renders a template according to the given flags, taking the
+// template form rf.Source and producing output in rf.Dest. This should be
```
yolocs, line 273:
```suggestion
// template from rf.Source and producing output in rf.Dest. This should be
```

yolocs (approved):

templates/commands/render.go
drevell: I'll fix this in a followup PR so I can keep my approval bit 😄  thanks!

Issue #35, opened 2023-06-20
Design and implement post-rendering validation


drevell:
### TL;DR

As a template developer, I want to be able to add checks after template rendering to make sure the resulting code is correct. If these validations fail, I want the CLI to show an error message to the user telling them that the template spec file that they're is buggy, and ask them to file a bug to the template owner.

Ideas for use cases:
 - Assert that the resulting code can build and compile
 - Assert that a given string or regex does or doesn't appear. For example, "template_replace_me" shouldn't appear because it should be replaced in rendering.
 - Assert that unit tests pass
 - EDIT: maybe also `go mod tidy`

One possible design: the template spec.yaml file could declare a list of validations, like:
```
- validations:
  - validator: 'go_lint'  
  - validator: 'go_compile_and_test'
  - validator: 'strings_must_not_appear'
    strings:
      - 'template_replace_me'
```

drevell:
Meeting note: I'm going to work on this, and @chloechien will contribute npm support. We will want a design doc.

yolocs:
How to validate a template could vary a lot. How about we leave it to the template repo GH workflow to do? My rough thoughts are:

I'm in my template repo. As a presubmit check, it would:

1. Use the abc CLI (either `latest` or `head`) to render the result into a folder
2. From this folder I could do various checks
   - The basic one: check if the content matches my "want" version - the CLI could provide this functionality since it's so common
   - Run lint and unit tests - the CLI doesn't need to worry about this; I get to choose what tools to use to run my test or lint
   - All kinds of other checks if needed

PR #105, opened 2023-07-18
Move test helpers to testutil/

drevell:
This is change 2 of 4-ish for that will add a test framework for templates. The goal is to (1) allow template developers to test their templates and (2) allow abc CLI developers to verify that code changes haven't broken templates.

Reviewers, the only mildly interesting parts of this PR are in render_test.go and testutil/*. Everything else is just pointing to the new location.

This change:

 - Makes the test helpers usable outside the "commands" package
 - Changes the names slightly for conciseness

The convention I chose is to import the testutil package as "abctestutil" to avoid colliding with the github.com/abcxyz/pkg/testutil package.

### DIFF
--- a/testutil/dirmap.go
+++ b/testutil/dirmap.go
@@ -1,0 +1,105 @@
+package testutil
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+type ModeAndContents struct {
+	Mode     os.FileMode
+	Contents string
+}
+
+// LoadDirContents reads all the files recursively under "dir", returning their
+// contents as a map[filename]->contents. Returns nil if dir doesn't exist.
+func LoadDirContents(t *testing.T, dir string) map[string]ModeAndContents {
+	t.Helper()
+
+	if _, err := os.Stat(dir); err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return nil
+		}
+		t.Fatal(err)
+	}
+	out := map[string]ModeAndContents{}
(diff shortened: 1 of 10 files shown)


testutil/dirmap.go
```diff
@@ -0,0 +1,121 @@
+// Copyright 2023 The Authors (see AUTHORS file)
```
sethvargo, line 1: I'm partial to naming this package "abctest" instead of "testutil". You're already aliasing the package since it conflicts with other "testutil" packages, but I think "abctest" is more descriptive, especially if we plan on having others use it.

yolocs:
Curious about where this goes - Do we expect template authors to write Go test to test their template?

drevell:
> Curious about where this goes - Do we expect template authors to write Go test to test their template?

Hmm that was my goal but now that you mention it it seems flawed. I'll write up a short doc for this and send it to you. My updated design thoughts are:

- There should be an `abc templates test` command
- Test inputs and outputs should be specified in `template-test.yaml` files that look like:

       ```
       kind: 'TemplateTest'
       golden_output_dir: './want_outputs'
       inputs:
         person_name: 'Alice'
         service_account: 'asdf@jkl.com'
       ```
 - There should also be a way to test all templates recursively using something `abc templates test ./...` or `abc templates test-recursive`, to be used in presubmits. We can detect all tests by crawling the directory and looking for yaml files containing the right `apiVersion` and `kind: TemplateTest`. Too bad Go's stdlib doesn't support extglobs like `**`.

This would avoid forcing people to write Go for tests, and avoid forcing them to install `go` to run the tests.

drevell:
> 

Actually, after talking with @chloechien, we don't need to build this right now. #35 will provide the template validation that we need for the immediate future. I'm going to abandon this PR.

PR #106, opened 2023-07-18
Instructions for releasing and downloading

### LINKED ISSUE #42: Set up a release process for binaries
### TL;DR

We should set up goreleaser to release binary artifacts according to all the best practices. We can use JVS and lumberjack as examples. We should support architectures amd64+arm64, and OSes linux+macOS.

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

drevell:
Closes #42.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ tool. "Rendering" a template is when you use the `abc` CLI to download some
 template code, do some substitution to replace parts of it with your own values,
 and write the result to a local directory.
 
-## One-time CLI installation
+## Installation
 
 There are two ways to install:
 
@@ -105,11 +105,7 @@ There are two ways to install:
       Example `curl` command, please substitute the version number you're
       downloading:
 
-          $ curl -O -L https://github.com/abcxyz/abc/releases/download/v1.2.3/abc_1.2.3_linux_amd64.tar.gz
-
-    - Unpack the tar file:
-
-          $ tar xvzf abc_1.2.3_linux_amd64.tar.gz
+          $ curl -sSL https://github.com/abcxyz/abc/releases/download/v1.2.3/abc_1.2.3_linux_amd64.tar.gz | tar -xzv abc
 
     - Now you will have an `abc` file that you can run. Perhaps place it in your
       `$PATH`.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -87,14 +87,35 @@ and write the result to a local directory.
 
 ## One-time CLI installation
 
-The quick answer is to just run
-`go install github.com/abcxyz/abc/cmd/abc@latest` .
+There are two ways to install:
```
verbanicm, line 90: Should this section just be called `Installation`, people will want to update this and one-time makes it sound like it auto updates or something?

README.md
```diff
@@ -87,14 +87,35 @@ and write the result to a local directory.
...
+        `darwin_amd64`
+
+      Example `curl` command, please substitute the version number you're
+      downloading:
+
+          $ curl -O -L https://github.com/abcxyz/abc/releases/download/v1.2.3/abc_1.2.3_linux_amd64.tar.gz
+
+    - Unpack the tar file:
+
+          $ tar xvzf abc_1.2.3_linux_amd64.tar.gz
```
verbanicm, line 112: This can be streamlined so it doesn't leave the tar file laying around and only extracts the binary file:

```shell
  curl -sSL https://github.com/abcxyz/abc/releases/download/v1.2.3/abc_1.2.3_linux_amd64.tar.gz | tar -xzv abc
```

README.md
```diff
@@ -87,14 +87,35 @@ and write the result to a local directory.
...
+
+    - Now you will have an `abc` file that you can run. Perhaps place it in your
+      `$PATH`.
```
kevya-google, line 115: If we're suggesting this, maybe we can also include:
`mv abc /usr/local/bin` or something.

README.md
```diff
@@ -565,3 +586,33 @@ with the corresponding inputs:
...
+  major version.
+
+- Push a signed tag in git, with the tag named with your version number, with a
+  message saying why you're creating this release. For example:
+
+      $ git tag -s -a v0.2.1 -m 'Release the bugfix for crashing during leap second'
+      $ git push origin v0.2.1
+
+- A GitHub workflow will be triggered by the tag push and will handle
+  everything. You will see the new release created within a few minutes.
```
kevya-google, line 618: Add link to release workflow?

README.md
drevell: I dunno about this, if the user already knows how $PATH works they can do it themselves, and if they don't, I don't want to try to explain it here. I'm reluctant to advise the user to touch their system directories.

kevya-google (dismissed):

verbanicm (approved):

PR #107, opened 2023-07-19
Add pipeline dry run template

rachelcao-g:


t/data_migration_pipeline/main.go
```diff
@@ -79,6 +80,14 @@ func main() {
 	// Convert each line to a data model
 	dataModels := emitResult(s, lines)
 
+	// Turn on dry run monitor
+	log.Println(*dryRun)
```
chloechien, line 84: please also add description text

t/data_migration_pipeline/main.go
```diff
@@ -79,6 +80,14 @@ func main() {
...
+	if *dryRun {
+		log.Println("dry run start")
```
chloechien, line 86: Usually in a dry run, we also need to know metrics like, e.g. total count of records, how many extra processing we did (e.g. backfill an empty ID). You can use beam.NewCounter to output and print.

rachelcao-g: Done

t/data_migration_pipeline/main.go
```diff
@@ -23,17 +23,22 @@ import (
...
 	"github.com/apache/beam/sdks/v2/go/pkg/beam/io/textio"
 	"github.com/apache/beam/sdks/v2/go/pkg/beam/runners/direct"
+	"github.com/apache/beam/sdks/v2/go/pkg/beam/x/beamx"
 )
 
 var (
 	input    = flag.String("input-csv-path", "", "The path of the input MySQL CSV dumped .")
 	database = flag.String("spanner-database", "", "The path of the output Spanner database.")
 	table    = flag.String("spanner-table", "", "The name of the output Spanner table.")
+	dryRun   = flag.Bool("dry-run", true, "whether the specified run is a dry run")
```
verbanicm, line 37: This feels odd to me to default a dry run to true.

In general prefer default false for boolean flags: https://github.com/abcxyz/readability/wiki/Go-Best-Practices#boolean-flags

verbanicm, line 37: Prefix these with `flag*` for clarity (i.e. flagInput)

t/data_migration_pipeline/main.go
```diff
@@ -52,18 +57,26 @@ func parseDataModel(record []string) *DataModel {
...
 	}, lines)
+
 	return dataModels
 }
 
+func ctxWithPtransformID(id string) context.Context {
+	ctx := context.Background()
+	ctx = metrics.SetPTransformID(ctx, id)
+	return ctx
+}
```
sethvargo, line 78: This seems incorrect. We should be threading through a parent context (e.g. from `main` or `realMain`) to properly handle cancellation.

sethvargo: For example, I'd expect:

```go
func main() {
  ctx, cancel := signal.NotifyContext(context.Background(),
    syscall.SIGINT, syscall.SIGTERM)
  defer cancel()

  ctx = metrics.SetPTransformID(ctx, "pipeline")

  // everything that accepts a `context.Context` uses `ctx`...

  direct.Execute(ctx, p)
}
```

sethvargo (changes requested):

t/data_migration_pipeline/main.go
rachelcao-g: Done

rachelcao-g: Done

t/data_migration_pipeline/main.go
rachelcao-g: Updated

sethvargo (approved):

verbanicm (dismissed):

t/data_migration_pipeline/main.go
```diff
@@ -52,37 +59,58 @@ func parseDataModel(record []string) *DataModel {
...
 }
 
 func main() {
+	// Handle context cancellation.
+	ctx, cancel := signal.NotifyContext(context.Background(),
+		syscall.SIGINT, syscall.SIGTERM)
+	defer cancel()
+
+	// Set up metric
+	ctx = metrics.SetPTransformID(ctx, "pipeline")
```
chloechien, line 83: Any reason you set SetPTransformID? This is a simple one-step pipeline and we dont necessarily need to distinguish different PTransforms in the metrics. Also if you want to have this, a PTransform id should be clear which tranform stage it is.

t/data_migration_pipeline/main.go
```diff
@@ -52,37 +59,58 @@ func parseDataModel(record []string) *DataModel {
...
 	}
 
 	// Convert each line to a data model
-	dataModels := emitResult(s, lines)
+	dataModels := emitResult(ctx, s, lines)
+
+	// Verify data on dry run mode
+	if *flagDryRun {
+		log.Println("dry run start....")
+		if _, err := beamx.RunWithMetrics(ctx, p); err != nil {
```
chloechien, line 102: You should print metrics in both dry run and real run. The only difference is for dry run, we don't do spanner write.

rachelcao-g: got it

t/data_migration_pipeline/main.go
```diff
@@ -20,20 +20,27 @@ import (
...
-	input    = flag.String("input-csv-path", "", "The path of the input MySQL CSV dumped .")
-	database = flag.String("spanner-database", "", "The path of the output Spanner database.")
-	table    = flag.String("spanner-table", "", "The name of the output Spanner table.")
+	flagInput    = flag.String("input-csv-path", "", "The path of the input MySQL CSV dumped .")
+	flagDatabase = flag.String("spanner-database", "", "The path of the output Spanner database.")
+	flagTable    = flag.String("spanner-table", "", "The name of the output Spanner table.")
+	flagDryRun   = flag.Bool("dry-run", false, "whether the specified run is a dry run")
 )
 
+var count = beam.NewCounter("dry-run", "count")
```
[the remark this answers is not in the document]
rachelcao-g: Done

t/data_migration_pipeline/README.md
```diff
@@ -49,6 +49,8 @@ When pipeline starts, the client library automatically checks for SPANNER_EMULAT
     ```shell
     $ go run main.go -input-csv-path "test-data.csv" -spanner-database "projects/[your-project-id]/instances/test-instance/databases/testdb" -spanner-table "mytable"
     ```
+    - Add `-dry-run=true` to active the dry run mode.
```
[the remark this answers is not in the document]
rachelcao-g: Done

t/data_migration_pipeline/main.go
rachelcao-g: if I don't add it, it gives `Unable to dump metrics: provided context doesn't contain metrics Store.` Following the example https://pkg.go.dev/github.com/apache/beam/sdks/v2/go/pkg/beam#example-package-MetricsDeclaredAnywhere Not sure whether it's a good example

chloechien: Got that. I might missed sth, fond this example not setting PTransformID: https://github.com/spotify/beam/blob/bddf2be8057fd2f7966451b8a4a4a90d8ff02820/sdks/go/test/load/pardo/pardo.go#L103

You can also set that, I'll defer to your choice, if so, do 
ctx = metrics.SetPTransformID(ctx, "emitResult")

rachelcao-g: my 2 cents after reading the code: in summary, the mechanism are different. Actually `SetPTransformID` is nothing related to `RunWithMetrics`. `DumpToLog` must have `SetPTransformID` to setup store and based on it to find counter stored in context. `RunWithMetrics` requires ParDo with counter before hand. 
I guess print out metrics is different in local run once and streaming? Another example using a total different solution https://beam.apache.org/documentation/programming-guide/#using-metrics

t/data_migration_pipeline/main.go
```diff
@@ -52,37 +59,58 @@ func parseDataModel(record []string) *DataModel {
...
-	spannerio.Write(s, *database, *table, dataModels)
+	// Verify data on dry run mode
+	if _, err := beamx.RunWithMetrics(ctx, p); err != nil {
+		log.Fatalf("Pipeline failed: %v", err)
+	}
+	metrics.DumpToLog(ctx)
+
+	// Terminate the pipeline if the dry run mode is active
+	if *flagDryRun {
+		return
```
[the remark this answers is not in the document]
rachelcao-g: Done

t/data_migration_pipeline/main.go
rachelcao-g: The other thing is to keep the case simple, I don't want to overkill the case

chloechien: I see, if it's required by DumpToLog, lets go with this solution.

chloechien (approved):

verbanicm (approved):

t/data_migration_pipeline/main.go
```diff
@@ -52,37 +58,57 @@ func parseDataModel(record []string) *DataModel {
...
 }
 
 func main() {
+	// Handle context cancellation.
+	ctx, cancel := signal.NotifyContext(context.Background(),
+		syscall.SIGINT, syscall.SIGTERM)
+	defer cancel()
+
+	// Set up metric
+	ctx = metrics.SetPTransformID(ctx, "emitResult")
```
yolocs, line 82: Would recommend using the `realMain` pattern. E.g. https://github.com/abcxyz/github-metrics-aggregator/blob/7e8df8db3fa0dbe69b9a7833f8205002a8d9b66a/pipeline/leech/main.go#L53

yolocs (approved):

PR #110, opened 2023-07-23
Update CLI package and refactor render flags

sethvargo:
This moves render flags to be instantiated when Flags is called and adds a private function that makes testing easier.


templates/commands/render.go
```diff
@@ -86,31 +86,25 @@ are accepted:
...
 }
 
-// RunWithFlags renders a template according to the given flags, taking the
-// template form rf.Source and producing output in rf.Dest. This should be
-// considered an unstable API and is primarily intended for testing. Use at your
-// own risk and accept the possibility of breaking changes.
-func (c *RenderCommand) RunWithFlags(ctx context.Context, rf *RenderFlags) error {
-	c.flags = rf
+func (c *RenderCommand) Run(ctx context.Context, args []string) error {
+	if err := c.Flags().Parse(args); err != nil {
```
drevell, line 96: Maybe I'm missing something here, or maybe there's a problem, you can tell me which.

The point of having `RunWithFlags()` was to offer an API for programmatically invoking a RenderCommand, providing the flags as a struct. In this PR, I see a `Flags()` function that seems to be intended to support that use case, but it's not actually usable because the flag struct returned from Flags() will be discarded and re-parsed at line 96.

drevell (changes requested):

templates/commands/render.go
```diff
@@ -54,7 +54,7 @@ const (
 
 type RenderCommand struct {
 	cli.BaseCommand
-	flags *RenderFlags
+	renderFlags RenderFlags
```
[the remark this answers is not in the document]
sethvargo: Done.

templates/commands/render.go
sethvargo: > but it's not actually usable because the flag struct returned from Flags() will be discarded and re-parsed at line 96.

I'm not sure what you mean. The result, after running `c.Flags().Parse(args)`, is that the `RenderCommand` has fully parsed flags and arguments into values on the structure.

The general point of the CLI is that, after Parse, you never directly interact with args or flags again. All the state is stored on that instance of the Command (or embedded structures as is in this case).

> The point of having RunWithFlags() was to offer an API for programmatically invoking a RenderCommand, providing the flags as a struct.

For internal testing, create a `RenderCommand` and set `cmd.flags` directly on the struct.

I think there's a leaky abstraction here, where you're trying to use the RenderCommand as both a CLI entry **and** a library. If there are use cases for rendering outside of the command, we should probably extract the rendering logic into a dedicated package.

drevell: I see, SGTM.

drevell (approved):

sethvargo:
@drevell I had to rebase. Can you re-approve please?

Issue #101, opened 2023-07-14
Add `append` action to add to the end of a file


drevell:
### TL;DR

For the use case of "adding a new `output` to the end of outputs.tf", it would be very useful to have an append action. Like this:

```
- action: 'append'
  params:
    path: 'foo/bar/outputs.tf'
    with: |
      output {
         blah blah
      }
```

cc @verbanicm @kevya-google

PR #111, opened 2023-07-26
WIP: get some feedback on append action

pdewilde:
Adding append #101

### DIFF
--- a/templates/commands/render_action_append.go
+++ b/templates/commands/render_action_append.go
@@ -26,10 +26,14 @@ func actionAppend(ctx context.Context, ap *model.Append, sp *stepParams) error {
 		return err
 	}
 
+	with, err := parseAndExecuteGoTmpl(ap.With.Pos, ap.With.Val, sp.inputs)
+	if err != nil {
+		return err
+	}
+
 	if err := walkAndModify(ctx, ap.Path.Pos, sp.fs, sp.scratchDir, path, func(buf []byte) ([]byte, error) {
-		// todo: this only works well if encoding of file is same as the string
 		// todo: should we add a newline before/after appending or leave that to user?
-		return append(buf, []byte(ap.With.Val)...), nil
+		return append(buf, []byte(with)...), nil
 	}); err != nil {
 		return err
 	}
--- a/templates/commands/render_action_append.go
+++ b/templates/commands/render_action_append.go
@@ -1,0 +1,38 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


templates/commands/render_action_append.go
```diff
@@ -0,0 +1,38 @@
...
+
+package commands
+
+import (
+	"context"
+
+	"github.com/abcxyz/abc/templates/model"
+)
+
+func actionAppend(ctx context.Context, ap *model.Append, sp *stepParams) error {
```
pdewilde, line 23: Would like feedback on use of parse&execute and walk&modify. I based off of what I saw in the string replace action, but don't fully understand why I'm parsing a go template when its not really a template.

templates/commands/render_action_append.go
```diff
@@ -0,0 +1,38 @@
...
+	path, err := parseAndExecuteGoTmpl(ap.Path.Pos, ap.Path.Val, sp.inputs)
+	if err != nil {
+		return err
+	}
+
+	if err := walkAndModify(ctx, ap.Path.Pos, sp.fs, sp.scratchDir, path, func(buf []byte) ([]byte, error) {
+		// todo: this only works well if encoding of file is same as the string
```
pdewilde, line 30: What do we think about different encodings here? Seems like it could be problematic

templates/model/spec.go
```diff
@@ -571,6 +577,41 @@ func (s *StringReplacement) UnmarshalYAML(n *yaml.Node) error {
...
+	*s = Append(*shadow)
+	s.Pos = yamlPos(n)
+
+	return nil
+}
+
+// Validate implements Validator.
+func (s *Append) Validate() error {
+	return errors.Join(
+		notZero(s.Pos, s.Path, "path"),
```
pdewilde, line 610: Do these validations make sense? I'm assuming we want it to be an error if missing rather than noop.

templates/commands/render_action_append.go
drevell: We use go-templates here so the spec.yaml can use different filenames depending on the input. Like:

```
action: 'append'
params:
  path: 'dbconfigs/{{.dbname}}.cfg
  with: ...
```

We try to support go-template expansion in as many places as possible in the spec file.

templates/commands/render_action_append.go
drevell: So far we've been assuming everything is utf8. Filed https://github.com/abcxyz/abc/issues/112 to think more about encodings.

templates/model/spec.go
drevell: Yeah, this looks good to me.

drevell (commented):
This all looks good to me so far.

PR #114, opened 2023-08-01
Pdewilde/append action

### LINKED ISSUE #101: Add `append` action to add to the end of a file

pdewilde:
Resolves #101

Add an append function.

I also sorted some actions in render.go alphabetically, as there didn't seem to be any pattern.


templates/commands/render_action_append.go
```diff
@@ -0,0 +1,42 @@
...
+		return err
+	}
+
+	with, err := parseAndExecuteGoTmpl(ap.With.Pos, ap.With.Val, sp.inputs)
+	if err != nil {
+		return err
+	}
+
+	if err := walkAndModify(ctx, ap.Path.Pos, sp.fs, sp.scratchDir, path, func(buf []byte) ([]byte, error) {
+		// todo: should we add a newline before/after appending or leave that to user?
```
pdewilde, line 35: Discussion question #1

templates/commands/render_action_append_test.go
```diff
@@ -0,0 +1,145 @@
...
+			want:            map[string]string{"my_file.txt": "abc foo deffoobar"},
+		},
+		{
+			name:            "empty_file_works",
+			path:            "my_file.txt",
+			with:            "foo",
+			initialContents: map[string]string{"my_file.txt": ""},
+			want:            map[string]string{"my_file.txt": "foo"},
+		},
+		// TODO: discuss in code review if this is desired behavior.
```
pdewilde, line 58: Discussion question #2

templates/commands/render_action_append.go
sethvargo: My gut feeling is "no". It's fairly easy to communicate "the string is appended exactly as-is; no newlines are added" versus adding the newline and then needing to support a use case where the customer doesn't want to add the newline (e.g. they want to add a trailing line comment).

templates/commands/render_action_append.go
```diff
@@ -0,0 +1,42 @@
...
+		return append(buf, []byte(with)...), nil
```
sethvargo, line 36: It's worth noting that we'll never iterate over the appended data. That's probably the _right_ decision, but consider:

1. An appended string that has the string "BANANA"
1. A regex replace rule for "BANANA"

If the actions are #1 and then #2, then the string will be replaced. However, if the actions are written as #2 and then #1, only instances of "BANANA" will be replaced in the original template, but the appended string will still contain "BANANA".

Again, I think this is probably the _right_ decision, but we should document this pretty heavily.

sethvargo: As an aside, I wonder if we need debug functionality to print out the template after each action /cc @drevell

templates/commands/render_action_append_test.go
sethvargo: This seems correct to me, although we should consider unifying all "file not found" errors into a single error type (not in this PR).

templates/model/spec.go
```diff
@@ -224,6 +225,10 @@ func (s *Step) UnmarshalYAML(n *yaml.Node) error {
 		s.StringReplace = new(StringReplace)
 		unmarshalInto = s.StringReplace
 		s.StringReplace.Pos = s.Pos
+	case "append":
```
sethvargo, line 228: I wonder if "append" is too generic of a name. Even within this PR, I think we could have two functions:

- `append_file`
- `append_string`

pdewilde: Is there a use for append_string that couldn't be done using the template language instead?

templates/commands/render_action_append.go
pdewilde: I feel like this isn't unique to append. Pretty much any file modification will not be commutative. The same could be for two regex replaces in a row.

1. REGEX("FOO", "FOOBAR")
2. REGEX("BAR", "BARFOO")

If its 1 then 2, FOO->FOOBARFOO
If its 2 then 1, FOO->FOOBAR

I'm not certain extra documentation is needed besides general documentation that all actions are run in order.

templates/commands/render_action_append.go
pdewilde: (inlining for easier reading)
@sethvargo  My gut feeling is "no". It's fairly easy to communicate "the string is appended exactly as-is; no newlines are added" versus adding the newline and then needing to support a use case where the customer doesn't want to add the newline (e.g. they want to add a trailing line comment).

pdewilde: I guess the main thing I'm not sure how to fix would be if someone wants to remove the trailing newlines before appending. Worth making an extra flag or something?

```
This is some text file.
It has text.
I want to finish this
```

If you wanted to append `sentence.` to the end, currently you would be forced to do it after the newline `append("sentence\n)`

```
This is some text file.
It has text.
I want to finish this
 sentence.
```

What you really want is
```
This is some text file.
It has text.
I want to finish this sentence.
```

I guess you might be able to do it with a regex replace though, as it seems that the default is not to match "end of text" line-per-line but with the entire file https://github.com/google/re2/wiki/Syntax

^^ Might be worth considering if we should document that in regex_replace, as its very different behavior than most people expect with regex from GREP.

pdewilde: Added regex replace documentation: #118

templates/commands/render_action_append.go
drevell: Filed #121 for improved template debugging support as suggested by Seth above.

templates/commands/render_action_append.go
pdewilde: Hmm, if we do that, then this we would need a special flag for newline, as there wouldn't be another way to append to the end of a file but remove the \n

drevell: What seems best to me is to ensure the file ends with newline, by default. The user could disable this with a new option `ensure_ending_newline: false`.

templates/commands/render_action_append_test.go
drevell: Looks good to me

templates/commands/render_action_append.go
pdewilde: I didn't read closely enough, you are suggesting a special mode for regex spanning lines.

templates/model/spec.go
pdewilde: Thinking more about it, is it worth having an append? The regex_replace action as it currently stands could do the exact same thing. A simple test I made:

```
		{
			name: "append",
			initContents: map[string]string{
				"a.txt": "apple banana\nbanana apple\napple apple\n", //nolint:dupword
			},
			rr: &model.RegexReplace{
				Paths: modelStrings([]string{"."}),
				Replacements: []*model.RegexReplaceEntry{
					{
						Regex: model.String{Val: "$"},
						With:  model.String{Val: "foobar\n"},
					},
				},
			},
			want: map[string]string{
				"a.txt": "apple banana\nbanana apple\napple apple\nfoobar\n", //nolint:dupword
			},
		},
```

drevell: My proposal is that we can have a single `append` action that supports with `with` and `with_file`. We can wait to build the `with_file` option until somebody needs it.

pdewilde: What would the `with_file` do, append one file onto another?

drevell (dismissed):

templates/commands/render_action_append.go
pdewilde: Added new param "skip-ensure-newline" and updated documentaiton. I am wondering if I need to support \r\n newlines or not

templates/commands/render_action_append.go
pdewilde: @sethvargo Do you still think we need documentation somewhere? I think its pretty intuitive that any mutations are done in order.

templates/commands/render_action_append.go
```diff
@@ -0,0 +1,50 @@
...
+		return err
+	}
+
+	with, err := parseAndExecuteGoTmpl(ap.With.Pos, ap.With.Val, sp.inputs)
+	if err != nil {
+		return err
+	}
+
+	if !ap.SkipEnsureNewline.Val {
+		// TODO: would we rather just cast to bytes than take dependency?
```
pdewilde, line 36: @drevell Thoughts? Seems like a style question. Technically casting to byte array would be faster, but harder to read.

templates/commands/render_action_append.go
drevell: I think we shouldn't support \r\n newlines at, least for now.

README.md
```diff
@@ -408,6 +408,29 @@ Example:
       thing'
 ```
 
+### Action: `append`
+
+Appends a string on the end of a given file. File must already exist. If no
+newline at tend of `with` parameter, one will be added unless
```
drevell, line 414: "tend" seems like a typo here

README.md
```diff
@@ -408,6 +408,29 @@ Example:
...
+`skip-ensure-newline` is set to `true`.
+
+Params:
+- `path`: File in which to do the replacement.
```
drevell, line 418: All the other actions take `paths` (plural) rather than `path`, should we keep that for consistency? Maybe there's not much legitimate need to append to multiple files? I'm on the fence.

templates/commands/render_action_append.go
sethvargo: I think we should document it, even if it's just in the README/examples. _Someone_ will open an issue, and it's always good to have something to point to.

templates/model/spec.go
sethvargo: I agree it's possible with regex, but I don't think it's a good devex. I like Dave's proposal:

```yaml
- action: 'append'
  with: 'dfjkdladf'

- action: 'append'
  with_file: './my-file.txt'
```

that would add `dfjkdladf` and `$(cat ./my-file.txt)` to the template.

drevell: > What would the with_file do, append one file onto another?

Yeah, that was my thought. It won't be a common use case, but suppose you want to append a huge amount of data to a file; rather than inlining the huge amount of data in the spec.yaml you could store it as append_me.txt in git then use `append` `with_file` to tack it onto the end of an existing file.

I think we shouldn't build this now though; we may never need it.

templates/model/spec.go
```diff
@@ -571,6 +577,42 @@ func (s *StringReplacement) UnmarshalYAML(n *yaml.Node) error {
...
 }
 
+// Append is an action that appends some output to the end of the file.
+type Append struct {
+	// Pos is the YAML file location where this object started.
+	Pos *ConfigPos `yaml:"-"`
+
+	Path              String `yaml:"path"`
+	With              String `yaml:"with"`
+	SkipEnsureNewline Bool   `yaml:"skip-ensure-newline"`
```
drevell, line 587: The convention in spec.yaml is to use underscores rather than dashes. As in `regex_replace`, `strip_prefix`, etc

drevell (changes requested):

templates/model/spec.go
pdewilde: OK, I'll leave `with_file` out for now.

README.md
pdewilde: `-t`

templates/model/spec.go
pdewilde: `s/skip-ensure-newline/skip_ensure_newline/g`

README.md
pdewilde: That's a good question, I don't have a strong opinion.

`path` is simpler and someone could use `regex_replace` if they really needed paths, but I do like having uniformity in APIs

drevell: Lets go with `paths` for uniformity. It's easy to wrap a single path with `[` `]`.

sethvargo (approved):
LGTM pending s/path/paths

README.md
pdewilde: Changed and tests added

README.md
```diff
@@ -408,6 +410,32 @@ Example:
...
+Appends a string on the end of a given file. File must already exist. If no
+newline at end of `with` parameter, one will be added unless
+`skip_ensure_newline` is set to `true`.
+
+If you need to remove an existing trailing newline before appending, use
+`regex_replace` instead.
+
+Params:
+- `paths`: List of files and/or directory trees to append to end of.
+  May use template expressions (e.g. `{{.my_input}}`).
```
drevell, line 424: Let's add `Directories will be crawled recursively and every file underneath will be processed.` to the end of the docs for `paths`; we're adding that in #123 to every action doc.

drevell (approved):

README.md
```diff
@@ -408,6 +410,32 @@ Example:
...
+
+Appends a string on the end of a given file. File must already exist. If no
+newline at end of `with` parameter, one will be added unless
+`skip_ensure_newline` is set to `true`.
+
+If you need to remove an existing trailing newline before appending, use
+`regex_replace` instead.
+
+Params:
+- `paths`: List of files and/or directory trees to append to end of.
```
sethvargo, line 423: I'm a little worried about accepting multiple values as an array instead of an array of a concrete type:

```yaml
paths: ['a', 'b']
```

versus

```yaml
paths:
  - path: 'a'
  - path: 'b'
```

The second is more verbose, but gives us flexibility in the future to add control values in the future:

```yaml
paths:
  - path: 'a'
    banana: 'apple'
  - path: 'b'
```

pdewilde: This was done to be consistent with all the existing actions which use paths as a list. If we want to change that, it should probably be a separate issue that changes all known templates and actions at once. (Assuming we are not guaranteeing a stable API yet)

sethvargo: https://github.com/abcxyz/abc/issues/130

drevell (approved):

PR #116, opened 2023-08-02
fix: regex replacement with subgroups

drevell:
Replacing an entire regex match worked, but replacing just a subgroup within a regex was buggy. This fixes the bug and adds some tests that would have caught the problem.

### DIFF
--- a/templates/commands/render_action_regexreplace.go
+++ b/templates/commands/render_action_regexreplace.go
@@ -116,7 +116,7 @@ func replaceWithTemplate(allMatches [][]int, b []byte, rr *model.RegexReplaceEnt
 			}
 		}
 		replaceAtStartIdx := oneMatch[subgroupNum*2] // bounds have already been checked in the caller
-		replaceAtEndIdx := oneMatch[subgroupNum+1]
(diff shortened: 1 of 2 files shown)


templates/commands/render_action_regexreplace.go
```diff
@@ -116,7 +116,7 @@ func replaceWithTemplate(allMatches [][]int, b []byte, rr *model.RegexReplaceEnt
 			}
 		}
 		replaceAtStartIdx := oneMatch[subgroupNum*2] // bounds have already been checked in the caller
-		replaceAtEndIdx := oneMatch[subgroupNum+1]
+		replaceAtEndIdx := oneMatch[subgroupNum*2+1]
```
sethvargo, line 119: Does https://pkg.go.dev/regexp#Regexp.SubexpIndex make any of this easier?

sethvargo (approved):

templates/commands/render_action_regexreplace.go
drevell: We actually do call that above, at line 113 😀

PR #117, opened 2023-08-02
feat: add toHyphenCase template funcs

verbanicm:
- Move the template funcs into their own file for readability and clarity as the list grows
- Add new template functions for `toHyphenCase`, `toUpperHyphenCase`, `toLowerHyphenCase`
- Chose to use the name HyphenCase instead of KebabCase as its not immediately clear what that means if you have never heard of it


templates/commands/template_funcs.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"regexp"
+	"strings"
+)
+
+var (
+	// hyphenOrSnakeCaseKeep are all the characters to keep for generating snake_case or hyphen-case strings.
+	// The regex is used to match all characters not specified here so we can remove them.
+	hyphenOrSnakeCaseKeep = regexp.MustCompile(`[^a-zA-Z0-9-_ ]+`)
```
drevell, line 25: Would these be cleaner as https://pkg.go.dev/strings#Replacer rather than regex?

drevell (changes requested):

templates/commands/template_funcs.go
drevell: Never mind, this doesn't work for this use case.

drevell (approved):

PR #118, opened 2023-08-02
Pdewilde/regex multiline test documentation

pdewilde:
Generally when using a tool such as GREP, if your regex has a `^` it will match the start of every line, but the default for RE2 is to match the start of the entire file/string unless the `m` multiline flag is set within the regex.

Added tests for this behavior and documentation so users are not surprised by this behavior.


pdewilde:
Oops, based off wrong branch, will cut a new PR.

PR #119, opened 2023-08-02
Clarify behavior of ^ and $ in regex_replace action.

pdewilde:
Generally when using a tool such as GREP, if your regex has a ^ it will match the start of every line, but the default for RE2 is to match the start of the entire file/string unless the m multiline flag is set within the regex.

Added tests for this behavior and documentation so users are not surprised by this behavior.


templates/commands/render_action_regexreplace_test.go
```diff
@@ -52,6 +52,42 @@ func TestActionRegexReplace(t *testing.T) {
...
+		{
+			name: "override_multiline_true",
+			initContents: map[string]string{
+				"a.txt": "apple banana\nbanana apple\napple apple\n",
+			},
+			rr: &model.RegexReplace{
+				Paths: modelStrings([]string{"."}),
+				Replacements: []*model.RegexReplaceEntry{
+					{
+						Regex: model.String{Val: "(?flags:re m)\n$"},
```
pdewilde, line 82: Hmm, this was working before when I had `(?flags:re m)^apple` I wonder why its saying regex not valid now. Was trying to get a more compelling example, but it seems no good deed goes unpunished.

drevell: You'd do `(?m:your_regex)`.

pdewilde: Thanks!

README.md
```diff
@@ -456,6 +456,11 @@ Params:
     readability. Use a non-capturing group (like `(?:abc)|(?:abc)`) if you need
     grouping without capturing.
 
+    Note that by default, RE2 doesn't use multiline mode, so ^ and $ will match
+    the start and end of the entire file, rather than each line. To enter
+    multiline mode you need to set the flag by including this: `(?m:YOUR_REGEX_HERE)`.
+    More information available in RE2 docs.
```
yolocs, line 462: nit: add a link to the RE2 doc?

yolocs (approved):

README.md
pdewilde: There was one in the previous paragraph, do you think another is useful?

yolocs: Don't bother then :)

drevell (approved):

PR #123, opened 2023-08-03
Clarify docs after a user question

drevell:
This makes it more clear that directories are processed recursively and all files within are handled.

Also autoformat the file.


README.md
```diff
@@ -416,7 +417,8 @@ given string with a given replacement string.
 Params:
 
 - `paths`: a list of files and/or directories in which to do the replacement.
-  May use template expressions (e.g. `{{.my_input}}`).
+  May use template expressions (e.g. `{{.my_input}}`). Directories will be
+  crawled recursively and every file underneath will be processed.
```
pdewilde, line 421: What if you specify directories/files that overlap, will the same modification be applied multiple times?

pdewilde:
I would also like to see the https://github.com/abcxyz/abc/blob/main/templates/commands/render_action.go#L43 walk and modify function made a bit more clear. I thought it was recursively trying to find the file in the path, not that it walked starting from the path you give it.

drevell:
> I would also like to see the https://github.com/abcxyz/abc/blob/main/templates/commands/render_action.go#L43 walk and modify function made a bit more clear. I thought it was recursively trying to find the file in the path, not that it walked starting from the path you give it.

Can you suggest alternative wording? The comment about as clear as I know how to be: `//Recursively traverses the directory or file scratchDir/relPath, calling the given visitor for each file.`. But maybe that comment is ambiguous to others in a way that I can't see.

README.md
drevell: Let's leave that undocumented for now until #124 is fixed, since the behavior will shortly change, and it's an edge case.

pdewilde:
> 

I initially assumed it was recursively crawling to find the specified file/path, not using that as a start point for crawling. 

Maybe something like
```
 //Starts from scratchDir/relPath, calling the given visitor for the file, or recursively all files in the tree if a directory is provided.
 ```

pdewilde (approved):

yolocs (approved):

PR #128, opened 2023-08-04
Make files only be processed once

### LINKED ISSUE #124: Touch each file only once per action
### TL;DR

As a template developer, I want to avoid accidentally double-processing any files in a single action.

Currently, if a file appears multiple times in in the `paths` list for an action, it will be re-processed. This is probably not what the template author intended. In the following example, if `first_directory` and `second_directory` refer to the same directory, then the files in that directory will be un-intentionally subjected to repeated `string_replace`ing.

```
action: 'string_replace'
params:
  paths: ['{{.first_directory}}', '{{.second_directory}}']
  replacements:
    // snip
```

The desired behavior would be to touch each file only once per `action`. We can keep a tracker (a map maybe) of which files (or directories?) have already been visited for the current action.

Discussed offline with @pdewilde . We don't think a template developer would ever want the opposite behavior where a file could be double-processed.

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

pdewilde:
Currently walkAndModify can cause a file to be processed multiple times if there are duplicate references in an action's `paths` input.

This adds a "seen" map to ensure that each file is only visited a single time.

Resolves #124


templates/commands/render_action.go
```diff
@@ -40,8 +40,16 @@ type walkAndModifyVisitor func([]byte) ([]byte, error)
...
+// The seen map ensures that the same file isn't visited multiple times in subsequent
+// calls. It must be non-nil.
+func walkAndModify(ctx context.Context, pos *model.ConfigPos, rfs renderFS, scratchDir, relPath string, v walkAndModifyVisitor, seen map[string]struct{}) error {
 	logger := logging.FromContext(ctx).Named("walkAndModify")
+	if seen == nil {
+		// Intentionally forcing caller to initialize map to discourage them from just calling with nil and not using
+		// the map for future calls.
+		return fmt.Errorf("to avoid handling the same file multiple times, the seen map provided to " +
+			"walkAndModify must be non-nil")
+	}
```
sethvargo, line 52: Why do they need to pass in a map at all? Shouldn't the `seen` map be local to the walkAndModify function? Or is it because this is called for each action?

pdewilde: Yeah, its called for each function. Within a particular call to walkAndModify there is no danger of multiple passes (unless something evil is done with symlinks, which this won't solve either).

The danger comes with subsequent calls, so some information needs to be preserved between them.

pdewilde: An alternate solution would be to change walkAndModify to accept multiple starting paths.

drevell: I think passing a list of paths to walkAndModify would be clean, and I don't see any reason why it wouldn't work.

templates/commands/render_action.go
```diff
@@ -64,6 +72,10 @@ func walkAndModify(ctx context.Context, pos *model.ConfigPos, rfs renderFS, scra
 			return nil
 		}
 
+		if _, ok := seen[path]; ok {
```
drevell, line 75: I'm not sure, but I wonder if we should call filepath.Clean() (https://pkg.go.dev/path/filepath#Clean) on this to make sure we're seeing a normalized version of the filename.

Suppose somebody does this weird thing:

```
action: 'include'
paths: ['a.txt', './a.txt']
```

We'd want to detect that `mydir/a.txt` and `./mydir/a.txt` are the same file.

In fact, we can probably put the call to filepath.Clean() in `safeRelPath()` and fix this entire class of bugs forever.

templates/commands/render_action_test.go
```diff
@@ -224,6 +224,113 @@ func TestWalkAndModify(t *testing.T) {
...
+		seen            map[string]struct{}
+		initialContents map[string]string
+		want            map[string]string
+		wantErr         string
+
+		// fakeable errors
+		readFileErr  error
+		statErr      error
+		writeFileErr error
+	}{
```
drevell, line 247: Let's add a test case where the same file is named in multiple different ways, like `a.txt` and `./a.txt`

drevell (changes requested):

templates/commands/render_action_test.go
pdewilde: See `repeated_file_directory_only_visited_once` in `render_action_test.go`

templates/commands/render_action.go
pdewilde: I put in a test and it seems to already be handled correctly, do you think further modification is needed?

templates/commands/render_action.go
pdewilde: Changed

templates/commands/render_action.go
drevell: No need. I think what we're seeing is that filepath.Join() already calls clean on its output, so we're already getting the behavior we wanted. Thanks for checking.

templates/commands/render_action.go
```diff
@@ -37,65 +37,85 @@ import (
...
-	if _, err := rfs.Stat(walkFrom); err != nil {
-		if os.IsNotExist(err) {
-			return model.ErrWithPos(pos, `path %q doesn't exist in the scratch directory, did you forget to "include" it first?"`, relPath) //nolint:wrapcheck
-		}
-		return model.ErrWithPos(pos, "Stat(): %w", err) //nolint:wrapcheck
-	}
-
-	return filepath.WalkDir(walkFrom, func(path string, d fs.DirEntry, err error) error { //nolint:wrapcheck
+	for _, relPathPos := range relPaths {
+		relPathPos := relPathPos
```
drevell, line 52: Is this necessary? I don't see that relPathPos appears in any closures or escapes a single loop iteration in any way. But I could be missing something.

templates/commands/render_action.go
```diff
@@ -37,65 +37,85 @@ import (
...
+			if d.IsDir() {
+				return nil
+			}
 
-		relToScratchDir, err := filepath.Rel(scratchDir, path)
-		if err != nil {
-			return model.ErrWithPos(pos, "Rel(): %w", err) //nolint:wrapcheck
-		}
+			if _, ok := seen[path]; ok {
+				// File already processed.
```
drevell, line 77: Maybe add a debug log here?

templates/commands/render_action.go
```diff
@@ -37,65 +37,85 @@ import (
...
+			// because the file already exists.
+			if err := rfs.WriteFile(path, newBuf, ownerRWXPerms); err != nil {
+				return model.ErrWithPos(pos, "Writefile(): %w", err) //nolint:wrapcheck
+			}
+			logger.Debugw("wrote modification", "path", path)
+
+			return nil
+		})
+		if err != nil {
+			return fmt.Errorf("walkAndModify error for action: %w", err)
```
drevell, line 115: Maybe just `return err` here? I don't see the extra text adding much context. Typically if we annotate an error with the name of the function that failed, we do it in the caller, like:

```
if err := doThing(); err != nil {
  return fmt.Errorf("doThing: %w", err)
}
```

templates/commands/render_action_regexreplace.go
```diff
@@ -66,24 +66,21 @@ func actionRegexReplace(ctx context.Context, rr *model.RegexReplace, sp *stepPar
 		}
 	}
 
-	for _, p := range rr.Paths {
-		if err := walkAndModify(ctx, p.Pos, sp.fs, sp.scratchDir, p.Val, func(b []byte) ([]byte, error) {
-			for i, rr := range rr.Replacements {
-				cr := compiledRegexes[i]
-				allMatches := cr.FindAllSubmatchIndex(b, -1)
+	if err := walkAndModify(ctx, sp.fs, sp.scratchDir, rr.Paths, func(b []byte) ([]byte, error) {
```
drevell, line 69: Hm this might be a preexisting bug. We should do call `parseAndExecuteGoTmpl` on rr.Paths before passing them to walkAndModify, so things like `paths: ['{{.myinput}}/a.txt}}']` are handled.

Or, maybe we could have walkAndModify do the template expansion on paths, so all the callers don't have to have the same boilerplate?

drevell (dismissed):

templates/commands/render_action.go
pdewilde: The linter threw a fit so I did that, should I just disable the linter for that line?

drevell: Yeah, do the usual //nolint: thing.

templates/commands/render_action_regexreplace.go
pdewilde: I had noticed that too but originally thoguht it was intentional. I filed a bug #142 for the broken actions as well as a feature request #143  for your suggestion.

templates/commands/render_action.go
pdewilde: Added

templates/commands/render_action.go
pdewilde: No, I'll remove.

drevell (dismissed):

drevell (approved):

Issue #126, opened 2023-08-04
Add go template function for `sort`ing strings


kevya-google:
### TL;DR

Given a comma-separated string input, for lexicographical sorting a new function is needed.

Suggested in https://github.com/abcxyz/abc/pull/122#discussion_r1284654283

PR #129, opened 2023-08-05
Add sortStrings template functions

sethvargo:
Closes https://github.com/abcxyz/abc/issues/126

### DIFF
--- a/templates/commands/render_action.go
+++ b/templates/commands/render_action.go
@@ -121,17 +121,10 @@ func templateAndCompileRegexes(regexes []model.String, inputs map[string]string)
 // templateFuncs returns a function map for adding functions to go templates.
 func templateFuncs() template.FuncMap {
 	return map[string]any{
-		"contains":   strings.Contains,
-		"replace":    strings.Replace,
-		"replaceAll": strings.ReplaceAll,
-		"sortStrings": func(in []string) []string {
-			// The default sort.Strings behavior modifies the string in place. I
-			// thought it would be very weird if rendering a template changed the
-			// order of an input further down the stack.
-			cp := append([]string{}, in...)
-			sort.Strings(cp)
-			return cp
-		},
+		"contains":          strings.Contains,
+		"replace":           strings.Replace,
+		"replaceAll":        strings.ReplaceAll,
+		"sortStrings":       sortStrings,
 		"split":             strings.Split,
 		"toLower":           strings.ToLower,
 		"toUpper":           strings.ToUpper,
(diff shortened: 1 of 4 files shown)


templates/commands/render_action.go
```diff
@@ -120,10 +120,15 @@ func templateAndCompileRegexes(regexes []model.String, inputs map[string]string)
...
-		"replaceAll":        strings.ReplaceAll,
+	return map[string]any{
+		"contains":   strings.Contains,
+		"replace":    strings.Replace,
+		"replaceAll": strings.ReplaceAll,
+		"sortStrings": func(in []string) []string {
+			cp := append([]string{}, in...)
+			sort.Strings(cp)
+			return cp
+		},
```
sethvargo, line 131: The default `sort.Strings` behavior modifies the string in place. I thought it would be very weird if rendering a template changed the order of an input further down the stack.

kevya-google: I would add a comment explaining that it's sorted in-place and that's why this function looks different from the others.

kevya-google (approved):

templates/commands/render_action.go
sethvargo: Done.

kevya-google (dismissed):

kevya-google:
Also, are tests needed for this?

templates/commands/render_action.go
```diff
@@ -120,10 +120,18 @@ func templateAndCompileRegexes(regexes []model.String, inputs map[string]string)
...
-		"replaceAll":        strings.ReplaceAll,
+	return map[string]any{
+		"contains":   strings.Contains,
+		"replace":    strings.Replace,
+		"replaceAll": strings.ReplaceAll,
+		"sortStrings": func(in []string) []string {
+			// The default sort.Strings behavior modifies the string in place. I
+			// thought it would be very weird if rendering a template changed the
+			// order of an input further down the stack.
+			cp := append([]string{}, in...)
```
drevell, line 131: Maybe slices.Copy() here to make this easier to read?

drevell (changes requested):

templates/commands/render_action.go
sethvargo: It's `slices.Clone`, but sure. Done.

sethvargo:
> Also, are tests needed for this?

There were no tests for the other template functions, so I wrote a bunch. I also pulled out the function into its own function in template_funcs.go and added dedicated tests for it.

templates/commands/render_action_test.go
```diff
@@ -689,3 +689,122 @@ func TestUnknownTemplateKeyError_ErrorsIsAs(t *testing.T) {
 		t.Errorf("errors.As() returned false, should return true when called with an error of type %T", as)
 	}
 }
+
+// These are basic tests to ensure the template functions are mounted. More
+// exhaustive tests are at template_funcs_test.go.
+func TestTemplateFuncs(t *testing.T) {
```
drevell, line 695: Thanks for this.

drevell (approved):

PR #135, opened 2023-08-08
check in csp trusted type for react

rachelcao-g:
Goal: improve React security

Security concern:
React applies automatic escaping , but it has limitation - no strict contextual auto-escaping. For example, it allows innerHTML pass by dangerouslySetInnerHTML, which is easily to get XSS attack.  

Solution:
1.  Sanitization the content using DOMPurify package. 
cons: only works for dangerouslySetInnerHTML. 

2. Add Trusted Types - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/require-trusted-types-for 
pros: works for all types of inner context, such as ref.inner  
cons: not support in all browsers. 

addition:
I decided to move the backend api call, to make the template easier to start. The backend call would be shown in the demo.

### DIFF
--- a/t/react_template/public/index.html
+++ b/t/react_template/public/index.html
@@ -9,6 +9,8 @@
       name="description"
       content="React template"
     />
+    <!-- sctrict to add trusted type when using dangerouslySetInnerHTML -->
+    <meta http-equiv="Content-Security-Policy" content="require-trusted-types-for 'script'" />
     <!--
       manifest.json provides metadata used when your web app is installed on a
       user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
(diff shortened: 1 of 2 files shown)


drevell (approved):

drevell:
LGTM but please get review from somebody who knows Typescript too 🙂

sethvargo (approved):

t/react_template/public/index.html
```diff
@@ -9,6 +9,8 @@
       name="description"
       content="React template"
     />
+    <!-- sctrict to add trusted type when using dangerouslySetInnerHTML -->
```
chloechien, line 12: spelling? 
Suggest  , we'll be adding more in future.

chloechien (approved):
Thanks Rachel!

t/react_template/.gitignore
```diff
@@ -3,3 +3,4 @@ node_modules
 /cypress/videos
 /cypress/screenshot
 
+/build
```
chloechien, line 6: Empty line at the end?

chloechien (approved):

PR #137, opened 2023-08-09
Optionally prompt for template inputs

drevell:
What is this: if the user supplies the `--prompt` flag, then the user will be prompted for template inputs (except those already provided by `--input`).

Why? Because this makes it easy to run a template without knowing its inputs in advance, and can help avoid lengthy command lines. It also makes for a good demo.

Why is it opt-in with `--prompt` rather than defaulting to on? Consider this motivating use case: there's a playbook abc command that the user is copy-pasting, and this command has always worked with the  --input values provided in the playbook. There is no need for prompting. Then the upstream template developer adds a new input that is not anticipated by the playbook. In this case, we'd rather just fail outright than have the CLI prompt for the missing input. If we prompted the user, they'd be tempted to enter something creative rather than realize that their playbook needs to be updated.

Example of a prompt for an input that doesn't have a default:

```
Input name:    service_account
Description:   the name of the GCP service account that ....

Enter value: 
```

Example of a prompt for an input that *does* have a default:

```
Input name:    db_config_filename
Description:   an optional output location to use rather than the usual default location
Default:       my_config_dir/my_database_config.json

Enter value, or leave empty to accept default: 
```


templates/commands/render.go
```diff
@@ -355,6 +351,89 @@ func (c *RenderCommand) checkUnknownInputs(spec *model.Spec) []string {
...
+				defaultStr = `""`
+			}
+			fmt.Fprintf(tw, "\nDefault:\t%s", defaultStr)
+			tw.Flush()
+
+			// Design note: we can't treat "empty string" as meaning "use
+			// the default value" because empty string is a legitimate input
+			// value.
+
+			fmt.Fprintf(sb, "\n\nAccept the (d)efault or (p)rovide your own value? (d/p): ")
```
drevell, line 410: I'm not so sure about this, it doesn't seem very clean. But I can't think of a better way. LMK.

sethvargo: Can you provide sample output of a couple prompts in the description?

drevell: Done, the part I'm concerned about is the `Accept the (d)efault or (p)rovide your own value? (d/p)`.

yolocs: I wonder if we should just use defaults if they are available.

sethvargo: What are the situations where the empty string is a valid value **and** the default value is not the empty string?

drevell: > I wonder if we should just use defaults if they are available.

That feels risky to me. The user might be surprised that they silently got a value that doesn't work for them.

drevell: > What are the situations where the empty string is a valid value **and** the default value is not the empty string?

I think that's what I needed to hear. In the cases where empty string is a valid input, it can also be the default. I'll update the code.

drevell: Updated, LMK what you think. The example prompts in the PR description have also been updated.

sethvargo (approved):

templates/commands/render.go
```diff
@@ -353,6 +349,80 @@ func (c *RenderCommand) checkUnknownInputs(spec *model.Spec) []string {
...
+	}
+
+	return nil
+}
+
+// promptForInputs looks for template inputs that were not provided on the
+// command line and prompts the user for them. This mutates c.flags.Inputs.
+//
+// This must only be called when the user specified --prompt and the input is a
+// terminal (or in a test).
```
yolocs, line 382: nit: consider copy the example in the PR description over :)

yolocs (approved):

templates/commands/render.go
drevell: Will do in a followup so I don't lose my approval bit.

PR #138, opened 2023-08-09
fix: spec.yaml wasn't always being skipped

drevell:
Background: there's a special case in `abc` where the spec.yaml file isn't included in the template output, even if there's an `include` for path `.`. It turns out this was buggy, and is being fixed here.

Since we were comparing unnormalized paths to normalized paths, we'd sometimes overlook that `./spec.yaml` and `spec.yaml` were the same file. The visitor function provided to copyRecursive was being called with the output of `filepath.Rel()`, which calls `filepath.Clean()` implicitly. But the blocklist was being initialized with un-Clean()ed paths. So the fix is just to `Clean()` the path before adding it to the blocklist.

cc @kevya-google who noticed and reported this.


sethvargo (approved):

drevell:
This is obsolete and unnecessary since #141, closing.

PR #139, opened 2023-08-09
fix: log level and mode were being ignored

drevell:
We were setting the right environment variables to configure the `logging` package, but since we never called `logging.NewFromEnv`, it had no effect.

This fixes JSON output being printed to the console unexpectedly.

This is hard to unit test, but I manually verified that the output levels and modes look good. There's no more surprise JSON output.

cc @verbanicm who noticed this problem.

### DIFF
--- a/templates/commands/render.go
+++ b/templates/commands/render.go
@@ -98,6 +98,7 @@ func (c *RenderCommand) Run(ctx context.Context, args []string) error {
 	}
 
 	c.setLogEnvVars()
+	ctx = logging.WithLogger(ctx, logging.NewFromEnv("ABC_"))
 
 	fSys := c.testFS // allow filesystem interaction to be faked for testing
 	if fSys == nil {


templates/commands/render.go
```diff
@@ -98,6 +98,7 @@ func (c *RenderCommand) Run(ctx context.Context, args []string) error {
 	}
 
 	c.setLogEnvVars()
+	ctx = logging.WithLogger(ctx, logging.NewFromEnv("ABC_"))
```
verbanicm, line 101: @drevell We are already loading from env here: https://github.com/abcxyz/abc/blob/main/cmd/abc/abc.go#L55

I think you need to call your `setLogEnvVars` in that file. It would be nice to centralize this somehow instead of in each Run function. Maybe log level should only set via envvars and not via flags?

sethvargo: This is required here unfortunately, because setLogEnvVars depends on things that are defined by parsing the flags (-log-level), so it can't be done higher up the stakc.

I've fixed this in https://github.com/abcxyz/pkg/pull/138 with the move to slog, but it might take a couple weeks for golangci and the community to support Go 1.21 everywhere yet. That will have a flag for setting `-log-level` and support this kind of mutation.

So we can:

1. Merge this
2. Remove the log-level flag and custom logic for now (you can still use LOG_LEVEL=debug)
3. Leave it and we'll fix it in the migration

yolocs (approved):

PR #141, opened 2023-08-09
Remove the `--spec` flag

drevell:
This was a group decision at this week's team meeting. We think it's cleaner to have a template be unambiguously identified by its `<location>` string, rather than a (location,--spec) tuple. This simplifies the mental model for users.

So, now we always look for `spec.yaml` in the root directory of the template location that was provided by the user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ Usage: `abc templates render [flags] <template_location>`
 
 The `<template_location>` parameter is any directory-like location that can be
 downloaded by the library that we use for template downloads,
-github.com/hashicorp/go-getter. It should contain a `spec.yaml` file in its root
-(or use the `--spec` flag to specify an alternate spec file name).
+github.com/hashicorp/go-getter. It should contain a `spec.yaml` file in its
+root.
 
 Examples of template locations:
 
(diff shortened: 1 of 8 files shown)


verbanicm:
One last thought, what if someone has a file named spec.yaml in their template needed for some other purpose, should we at least allow them to specify a different name even if it has to exist in the CWD?

sethvargo (approved):

chloechien:
By reading the test case: spec_yaml_in_subdir_should_not_be_skipped
If they put spec.yaml in any subdir the file can still be usable and be copied to the destination dir. Only the root dir spec.yaml is locked down for abc cli rendering usage, which is still developer friendly

chloechien (approved):

drevell:
> One last thought, what if someone has a file named spec.yaml in their template needed for some other purpose, should we at least allow them to specify a different name even if it has to exist in the CWD?

I don't think so. That would remove the main benefit that we're trying to achieve here, which is for templates to be unambiguously identified by their location string. There's another way to solve the above problem, though: use `include as` to include `spec.whatever.yaml` as `spec.yaml`.

PR #145, opened 2023-08-11
fix: files were incorrectly skipped

drevell:
We were returning fs.SkipDir when skipping files *and* directories, but we should have only returned it when skipping a directory. When we return SkipDir to skip a file, we accidentally were skipping the remainder of the files in the directory.


drevell:
cc @kevya-google this is why we were seeing the weird missing App.tsx error yesterday.

sethvargo (approved):

Issue #121, opened 2023-08-02
Support template debugging user journeys


drevell:
### TL;DR

As a template developer, I want tools to understand why my `actions` aren't having the effect I intended.

Ideas:
 - ~~Offer the ability to fail with a helpful error message if an action doesn't affect any files~~ some actions, like `print`, are working as intended when they don't affect any files.
 - ~~Offer the ability to track changes to a particular file~~ we can skip this because it's solved by the next bullet point below
 - Offer the ability to show a diff of changes to each file due to each action

Maybe some of these use cases can be met with improved debug logging? Not sure.

sqin2019:
@drevell I am thinking two approaches for this problem:
1. use git, we can see changes for each step execution with `git log`, we are also able to check if there are any diffs or not for each step. we will end up having a `.git` folder in the destination though.
2. do not use git, in each action, we need to add a function to output the changes.

I prefer the first one, since it scales (no changes required for new actions) and clean, it does require the user to have git installed which should not be a concern as git is somehow required. 

Just quickly check with you to see if I am missing anything and what your thoughts are.

PR #146, opened 2023-08-11
feat: add --debug-scratch-contents flag

drevell:
This is a feature for template authors that makes it easier to understand the state of the scratch directory.
If the flag is provided, we print a recursive directory listing of the scratch directory after each step of the spec file.

This addresses part of #121. It was useful in debugging #145.

### DIFF
--- a/templates/commands/render.go
+++ b/templates/commands/render.go
@@ -460,34 +460,42 @@ func executeSpec(ctx context.Context, spec *model.Spec, sp *stepParams) error {
 		}
 		logger.Debugw("completed template action", "action", step.Action.Val)
 		if sp.flags.DebugScratchContents {
-			sb := &strings.Builder{}
-			fmt.Fprintf(sb, "Scratch dir contents after step %d (starting from 0), which is action type %q, defined at spec file line %d:\n",
-				i, step.Action.Val, step.Action.Pos.Line)
-			err := filepath.WalkDir(sp.scratchDir, func(path string, d fs.DirEntry, err error) error {
-				if err != nil {
-					return err // some filesystem error happened
-				}
-				if d.IsDir() {
-					// it's not possible to have an empty directory in the
-					// scratch directory, and directory names will be shown as
-					// part of filenames, so we don't show plain directory
-					// names. Like in Git.
-					return nil
-				}
-				rel, err := filepath.Rel(sp.scratchDir, path)
-				if err != nil {
-					return fmt.Errorf("filepath.Rel(): %w", err)
-				}
-				fmt.Fprintf(sb, "  %s\n", rel)
-				return nil
-			})
+			contents, err := scratchContents(ctx, i, step, sp)
 			if err != nil {
-				return fmt.Errorf("error crawling scratch directory: %w", err)
+				return err
 			}
-			logger.Info(sb.String())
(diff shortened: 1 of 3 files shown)


templates/commands/render_flags.go
```diff
@@ -122,6 +126,13 @@ func (r *RenderFlags) Register(set *cli.FlagSet) {
 		Usage: "Prompt the user for template inputs that weren't provided as flags.",
 	})
 
+	f.BoolVar(&cli.BoolVar{
```
sethvargo, line 129: I wonder:

- Should this be under a new "TEMPLATE AUTHORS" or "DEVELOPMENT" flag header?
- Should this be hidden?

sethvargo (dismissed):

templates/commands/render.go
```diff
@@ -454,11 +454,38 @@ func (c *RenderCommand) checkInputsMissing(spec *model.Spec) []string {
...
 	logger := logging.FromContext(ctx).Named("executeSpec")
 
-	for _, step := range spec.Steps {
+	for i, step := range spec.Steps {
 		if err := executeOneStep(ctx, step, sp); err != nil {
 			return err
 		}
 		logger.Debugw("completed template action", "action", step.Action.Val)
+		if sp.flags.DebugScratchContents {
+			sb := &strings.Builder{}
```
sethvargo, line 463: Nit: it might be nice to move this to a function that returns a `string` so this becomes:

```go
if sp.flags.DebugScratchContents {
  logger.Info(debugScratch(sp.scratchDir))
}
```

templates/commands/render.go
```diff
@@ -454,11 +454,38 @@ func (c *RenderCommand) checkInputsMissing(spec *model.Spec) []string {
...
+			fmt.Fprintf(sb, "Scratch dir contents after step %d (starting from 0), which is action type %q, defined at spec file line %d:\n",
```
sethvargo, line 464: Should this be structured (K=V) instead of string interpolated?

templates/commands/render.go
drevell: Done.

templates/commands/render.go
drevell: I just coded that up to see how readable it is, and it's much harder to read.

Here's what it loiks like as a plain string:

```
INFO	logScratchContents	commands/render.go:501	Scratch dir contents after step 1 (starting from 0), which is action type "string_replace", defined at spec file line 41:
  .editorconfig
  .eslintignore
  ... snip ...
```

And here's what it looks like with structured logging:
```
INFO	logScratchContents	commands/render.go:498	Scratch dir contents:   .editorconfig
  .eslintignore
...snip...
  public/index.html
  public/manifest.json
	{"step_index_starting_from_zero": 1, "action": "string_replace", "spec_yaml_line_number": 41}
```

I can change it if you really think it's important, but I think in this case it makes sense to choose readability.

templates/commands/render_flags.go
drevell: Added a new section TEMPLATE AUTHORS as suggested.

I don't think we want it to be hidden; I think we want it to be reasonably discoverable by template authors.

sethvargo (approved):

PR #149, opened 2023-08-15
Output module version from buildinfo if not built by goreleaser

### LINKED ISSUE #148: Supply hash for --version flag when from source, rather than HEAD
### TL;DR

Currently when you install with `go install` you get:
```
$ abc --version
abc source (HEAD, linux/amd64)
```

It would be helpful to get the git hash of the commit, as HEAD tells you nothing about the version of the CLI you are running. An additional improvement would be to also show the date of the commit.

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

drevell:
This fixes #148. Before, we weren't outputting the version number unless it had been placed in ldflags by goreleaser.

The outputted module version includes a partial git sha, so it unambiguously identifies the code that's running.

Example output when installed with `go install`:

```
$ abc -version
abc v0.0.1-alpha6.0.20230815192907-8d9c1c1a0444 (HEAD, linux/amd64)
```

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -26,7 +26,13 @@ var (
 
 	// Version is the main package version. This can be overridden by the build
 	// process.
-	Version = "source"
+	Version = func() string {
+		info, ok := debug.ReadBuildInfo()
+		if !ok {
+			return "source"
+		}
+		return info.Main.Version // e.g. "v0.0.1-alpha6.0.20230815191505-8628f8201363"
+	}()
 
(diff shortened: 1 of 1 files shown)


drevell:
cc @yolocs this might be applicable to JVS and Lumberjack also

yolocs:
> cc @yolocs this might be applicable to JVS and Lumberjack also

Great! Time to copy/paste.

Issue #127, opened 2023-08-04
Add `for_each` to repeat actions for a list of values


drevell:
### TL;DR

As a template author, I want to be able to run the same set of actions repeatedly for each element of a list. For example, I might have a list of "environments" and I want to create a subdirectory for each environment containing some files.

One idea of how this might work:

```
input:
  - name: 'environments'
    desc: 'a comma-separated list of environments, e.g. staging,prod'

steps:
  - action: 'for'
    params:
      each:  'split(input.environments, ",")'  # a CEL expression that outputs a list
      as: 'environment'  # Adds a new variable "{{.environment}}" into the go-template context
      sub_steps:
        - action: 'include'
          params:
            paths: ['main.tf']
            as: ['{{.environment}}/main.tf']

        # more per-environment actions go here

# After the "for" action, execution resumes without the {{.environment}}
```

Or, if we don't want to use CEL:

```
input:
  - name: 'environments'
    desc: 'a comma-separated list of environments, e.g. staging,prod'

steps:
  - action: 'for'
    params:
      each: '{{.environments}}'
      var_name: 'environment'
      split_char: ','
      sub_steps:
        - action: 'include'
          params:
            paths: ['main.tf']
            as: ['{{.environment}}/main.tf']

```

cc @verbanicm @pdewilde @kevya-google

PR #150, opened 2023-08-16
Add AST models for `for_each`

drevell:
This is the first of several PRs implementing `for_each` (see #127). This just adds the data structures that the YAML is parsed into, and some parsing tests.

### DIFF
--- a/templates/model/spec.go
+++ b/templates/model/spec.go
@@ -112,7 +112,7 @@ func (s *Spec) Validate() error {
 	return errors.Join(
 		oneOf(s.Pos, s.APIVersion, []string{"cli.abcxyz.dev/v1alpha1"}, "apiVersion"),
 		oneOf(s.Pos, s.Kind, []string{"Template"}, "kind"),
-		notZero(s.Pos, s.Desc, "desc"),
+		notZeroModel(s.Pos, s.Desc, "desc"),
 		nonEmptySlice(s.Pos, s.Steps, "steps"),
 		validateEach(s.Inputs),
 		validateEach(s.Steps),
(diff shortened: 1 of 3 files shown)


templates/model/spec.go
```diff
@@ -205,34 +206,38 @@ func (s *Step) UnmarshalYAML(n *yaml.Node) error {
 	// on the value of "action".
 	var unmarshalInto any
 	switch s.Action.Val {
-	case "print":
-		s.Print = new(Print)
-		unmarshalInto = s.Print
-		s.Print.Pos = s.Pos // Set an approximate position in case yaml unmarshaling fails later
+	case "append":
```
drevell, line 209: Note for reviewers: I alphabetized this list because it was feeling disorganized.

templates/model/spec.go
```diff
@@ -112,7 +112,7 @@ func (s *Spec) Validate() error {
 	return errors.Join(
 		oneOf(s.Pos, s.APIVersion, []string{"cli.abcxyz.dev/v1alpha1"}, "apiVersion"),
 		oneOf(s.Pos, s.Kind, []string{"Template"}, "kind"),
-		notZero(s.Pos, s.Desc, "desc"),
+		notZeroModel(s.Pos, s.Desc, "desc"),
```
drevell, line 115: Note for reviewers: I renamed this because now I want two versions of this function: one that takes a value wrapped in a valWIthPos[T], and another one that just takes a T.

templates/model/spec_test.go
```diff
@@ -721,6 +721,172 @@ params:
...
+			name: "for_each_no_steps",
+			in: `desc: 'mydesc'
+action: 'for_each'
+params:
+  iterator:
+    key: 'environment'
+    values: ['dev', 'prod']
+`,
+			wantValidateErr: `field "steps" is required`,
+		},
```
sethvargo, line 889: Can you add a test where `values` is given a single string and where `values_from` is given a slice?

sethvargo (approved):

templates/model/spec_test.go
drevell: Will do in a follow-up to keep my PR approval.

PR #157, opened 2023-08-18
Update to use slog

sethvargo:


drevell (approved):
Thanks for contributing this

drevell (approved):

PR #160, opened 2023-08-21
feat: for_each over CEL expression

drevell:
Previously, we could only iterate over a hardcoded YAML list of values. This PR adds the `values_from: 'some_cel_expression'` field to programmatically compute a list of strings to iterate over.

We may want to add caching of compiled CEL programs in the future to avoid paying a few milliseconds per compilation. We don't currently have any use cases where a for_each action will run often enough for that to matter, but it's conceivable that the latency would be noticeable for deeply-nested for_each actions.

Still to do: write docs and examples


templates/commands/cel.go
```diff
@@ -0,0 +1,132 @@
...
+		// splitting. This is based on some quick googling of other peoples'
+		// custom split functions.
+		cel.Function(
+			"split",
+			cel.MemberOverload(
+				"string_split",
+				[]*cel.Type{cel.StringType, cel.StringType},
+				types.NewListType(types.StringType),
+				cel.BinaryBinding(func(lhs, rhs ref.Val) ref.Val {
+					toSplit := lhs.Value().(string)
```
sethvargo, line 48: The linter is sad on these type assertions, but it's unclear if there's a way to return an error from this function.

templates/commands/cel.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	prog, err := env.Program(ast)
+	if err != nil {
+		return nil, expr.Pos.AnnotateErr(fmt.Errorf("failed constructing CEL program: %w", err))
+	}
+
+	logger := logging.FromContext(ctx).Named("celCompile")
+	latency := time.Since(startedAt)
+	logger.Debugw("CEL compilation and loading time", "duration_usec", latency.Microseconds())
```
sethvargo, line 85:
```suggestion
	logger.Debugw("cel compilation and loading time", "duration_usec", latency.Microseconds())
```

per https://github.com/abcxyz/readability/wiki/Go-Best-Practices#logging

sethvargo, line 85: One of the nice things about structured logging is that we can provide both machine and human friendly values. It might be nice to have:

```suggestion
	logger.Debugw("CEL compilation and loading time", 
		"duration_usec", latency.Microseconds(),
		"duration_human", latency.String())
```

templates/commands/cel.go
```diff
@@ -0,0 +1,132 @@
...
+// converted to the given type, then an error will be returned. For example, if
+// the CEL expression is "hello" and outPtr points to an int, an error will
+// returned because CEL cannot treat "hello" as an integer.
+func celEval(ctx context.Context, scope *scope, pos *model.ConfigPos, prog cel.Program, outPtr any) error {
+	startedAt := time.Now()
+
+	// The CEL engine needs variable values as a map[string]any, but we have a
+	// map[string]string, so convert.
+	scopeMapAny := make(map[string]any, len(scope.All()))
+	for varName, varVal := range scope.All() {
```
sethvargo, line 104: Nit: can `scope.All()` change between invocations or is it an expensive calculation? If so, we might want to grab it as a local variable once.

templates/commands/cel.go
```diff
@@ -0,0 +1,132 @@
...
+	outRefVal := outPtrRefVal.Elem()
+
+	celAny, err := celOut.ConvertToNative(outRefVal.Type())
+	if err != nil {
+		return pos.AnnotateErr(fmt.Errorf("CEL expression result couldn't be converted to %s. The CEL engine error was: %w", outRefVal.Type(), err))
+	}
+
+	outRefVal.Set(reflect.ValueOf(celAny))
+
+	logger := logging.FromContext(ctx).Named("celEval")
```
sethvargo, line 127: I don't think this is the correct syntax anymore for the new logger.

templates/model/pos.go
```diff
@@ -48,8 +48,8 @@ func (c *ConfigPos) AnnotateErr(err error) error {
 		return nil
 	}
 
-	pos := "(position unknown)" // This can happen when field values are defaults, rather that coming from the config file
-	if *c != (ConfigPos{}) {
+	pos := "(position unavailable)" // This can happen when field values are defaults, rather that coming from the config file
+	if c != nil && *c != (ConfigPos{}) {
```
sethvargo, line 52: What are the situations in which `c` can be nil? Are there other places where we're assuming it's not nil?

templates/commands/cel.go
drevell: I think the code as written here only invokes it once. Maybe I'm missing your point though.

templates/commands/cel.go
drevell: It's actually possible, the return value is a kind of implicit "Maybe" type. Fixed.

templates/commands/cel.go
drevell: That is nice, thanks. Done.

templates/model/pos.go
drevell: Currently this is just a convenience for tests. If you think it's important, I can update the tests to make it always non-nil, at the cost of a little verbosity.

templates/commands/cel.go
pdewilde: [Aren't proper nous and abbreviations usually exempted from this?](https://google.github.io/styleguide/go/decisions.html#:~:text=Error%20strings%20should%20not%20be,%22Something%20bad%20happened.%22))

templates/commands/cel.go
sethvargo: It's called on L103 and L104. I'm curious if the result is (1) expensive or (2) can change between invocations. Because then we might want:

```go
	scopes := scope.All()
	scopeMapAny := make(map[string]any, scopes)
	for varName, varVal := range scopes {
```

templates/model/pos.go
sethvargo: Nah - it's fine

templates/commands/cel.go
drevell: Oh of course, fixed. Thanks for catching that.

sethvargo (approved):

PR #161, opened 2023-08-22
Refactor spec.go to remove UnmarshalYAML boilerplate

### LINKED ISSUE #30: Clean up tech debt in YAML unmarshaling
### TL;DR

Our current YAML parsing works, but there might be a less-boilerplatey way to do YAML unmarshaling while:

- enforcing required fields
- enforcing that no extra unknown fields were present
- doing custom validation on field values

The upstream YAML parser has an interesting bug https://github.com/go-yaml/yaml/issues/460 where it will enforce that no extra fields are present, but that feature doesn't work if you call `node.Decode()`. That may limit our options for doing this cleanly.

Example of problematic code: in `spec.go`, there is some repeated almost-copy-pasted boilerplate code that looks like this:

```go
// UnmarshalYAML implements yaml.Unmarshaler.
func (s *Spec) UnmarshalYAML(n *yaml.Node) error {
	knownYAMLFields := []string{"apiVersion", "kind", "desc", "inputs", "steps"}
	if err := extraFields(n, knownYAMLFields); err != nil {
		return err
	}

	type shadowType Spec
	shadow := &shadowType{} // see "Q2" in file comment above
	if err := n.Decode(shadow); err != nil {
		return err
	}

	*s = Spec(*shadow)
	s.Pos = yamlPos(n)
	return nil
}
```

Copy-pasted code like this is almost always harmful.

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

drevell:
Before this, each UnmarshalYAML function was a copy-pasted block of 20-ish lines of boilerplate. With this change, each UnmarshalYAML function turns into a one-liner which calls unmarshalPlain.

This requires a somewhat regrettable amount of reflection code, but it's arguably no more confusing than what was there before, while having the benefit of being more concise and DRY.

Fixes #30.

I'm attempting to use stacked PRs here; I intend to switch the base branch of this PR to `main` once the `drevell/for-each-values-from` branch is merged.


sethvargo (dismissed):

templates/model/spec.go
```diff
@@ -80,7 +72,7 @@ func newDecoder(r io.Reader) *yaml.Decoder {
 // Spec represents a parsed spec.yaml file describing a template.
 type Spec struct {
 	// Pos is the YAML file location where this object started.
-	Pos *ConfigPos `yaml:"-"`
+	Pos ConfigPos `yaml:"-"`
```
yolocs, line 75: I wonder why de-pointer it?

yolocs (approved):

Issue #130, opened 2023-08-07
Make paths accept objects


sethvargo:
I'm worried about accepting multiple values as an array instead of an array of a concrete type:

```yaml
paths: ['a', 'b']
```

versus

```yaml
paths:
  - path: 'a'
  - path: 'b'
```

The second is more verbose, but gives us flexibility in the future to add control values in the future:

```yaml
paths:
  - path: 'a'
    banana: 'apple'
  - path: 'b'
```

This has come up a few times as we try to add additional options to control how a particular file behaves in the list.

drevell:
cc @kevya-google, this may be an alternative API for skip-if-exists behavior

drevell:
There's also the possibility of accepting both options in the parser. The user could choose between:

```
- paths: ['a.txt']
```

... and ...
```
- paths:
  - path: a.txt 
```

They'd choose the first one for conciseness, and the second one for fine-grained configuration of each path.

Downsides: complexity of code and documentation; users might be confused by having two ways to do it.

@sethvargo WDYT?

sethvargo:
It's definitely possible; we'd have to decode into `any` and then use something like `mapstructure` to get it into the proper Go types, but it's definitely possible. I think my biggest concern is that, the moment you need to customize _one_ file, you now have to break the entire list out (unless we have the paths be a crazy union type of `[]string | struct`, so you could do:

```yaml
paths:
  - 'a.txt'
  - 'b.txt'
  - path: 'c.txt'
    only_if: 'today.is_tuesday'
```

But honestly that feels even worse to me...

kevya-google:
Do multiple includes work?
```
paths: ['a', 'b'] # no options
paths: ['c', 'd'] # apply option 1 to c and d only
    apply_option_1: true
paths: ['d', 'e', 'f'] # apply option 2 to d, e, f
    apply_option_2: true
paths: ['g']
    apply_option_3: true
```
Or is this too messy?

sethvargo:
No, but paths could be a slice:

```yaml
paths:
  - paths: ['a', 'b']
    only_if: 'banana'
  - paths: ['c', 'd']
    notify: 'kitchen'
```

It's pretty similar to what I proposed above, but allows for multiple files to be specified instead of a single one.

kevya-google:
> No, but paths could be a slice:
> 
> ```yaml
> paths:
>   - paths: ['a', 'b']
>     only_if: 'banana'
>   - paths: ['c', 'd']
[...]

PR #162, opened 2023-08-23
Allow YAML decoding to change depending on apiVersion

drevell:
This is a hack, but I think it's the best option we have. It's a hack because it uses mutable global state where it really doesn't theoretically need to, but we're limited by our libraries.

In this PR we create a global variable that stores the apiVersion of the YAML document currently being parsed. We add a mutex to ensure that only one document can be parsed at a time.

Alternatives considered:

 - threadlocals: don't exist in Go (https://github.com/golang/go/issues/18802)
 - use a different YAML parser library that plumbs a context.Context to UnmarshalYAML(): github.com/goccy/go-yaml almost worked, but it doesn't provide line/column position information, which is a dealbreaker.
 - Just skip it and don't provide schema versioning: the schema would turn into a mess over time.
 - Write our own YAML decoder, potentially reusing https://pkg.go.dev/github.com/goccy/go-yaml/parser: would be lots of work.

This is a prelude to #130.


sethvargo:
Thanks Dave - my initial reaction is that we should just update the parsing logic to handle both arrays and single string values, passing everything as a slice thereafter (with string values being single-item slices).

If we really want versioned parsers, then I think we version _the parser itself_, which would have different decoding logic. Then, during the top-level parse, we'd have something like:

```go
if parsed.apiVersion == "1" {
  return v1parser.Parse(node)
else if parsed.apiVersion == "2" {
  return v2parser.Parse(node)
}
}
```

But I'm not super convinced we need to introduce versioning at this stage (but I'm open to being told I'm wrong 😄).

kevya-google:
> Thanks Dave - my initial reaction is that we should just update the parsing logic to handle both arrays and single string values, passing everything as a slice thereafter (with string values being single-item slices).

So how long would this change be kept moving forward, would there be a point where we remove support for the old version? Or would we keep both indefinitely?

drevell:
> Thanks Dave - my initial reaction is that we should just update the parsing logic to handle both arrays and single string values, passing everything as a slice thereafter (with string values being single-item slices).
> 
> If we really want versioned parsers, then I think we version _the parser itself_, which would have different decoding logic. Then, during the top-level parse, we'd have something like:
> 
> ```go
> if parsed.apiVersion == "1" {
>   return v1parser.Parse(node)
> else if parsed.apiVersion == "2" {
>   return v2parser.Parse(node)
> }
> }
> ```
> 
> But I'm not super convinced we need to introduce versioning at this stage (but I'm open to being told I'm wrong 😄).

Yeah, I see what you mean, and that would work. I see some tradeoffs here:

Option 1: Versioned parser:
 - Pro: each parser is easy to read
 - Con: absolutely massive copypasta. Discourages versioning because each version incurs new copypasta.

Option 2: Accept either list-of-string or list-of-object:
 - Pro: easy to code, no copypasta, no versioning concerns
 - Con: devex for YAML authors. The user just wants to know "what's the type of this field?" and the answer is "weep while trying to understand our polymorphic glory which is so easy to use that you're scrolling through examples trying to find something that you think might be what you're looking for but you're not sure." Like GitHub Actions (maybe k8s too, I dunno). I think polymorphic cleverness is often a misfeature in YAML schemas, but maybe I'm in the minority.

Option 3: The locked-global approach:
 - Pro: simple, avoids copypasta. Good devex for YAML authors (the right way to do it is the only way to do it).
 - Con: locking hackery. Only works for small changes to the schema. Combines logic for multiple schema versions in the same parser code.

So I think we'll have two kinds of schema change:
 - Minor tweaks like this one that don't warrant a new revision of the models, where Option 3 is appropriate.
 - Major tweaks that DO warrant a new revision of the models, where Option 1 is appropriate.

Can you live with that?

@sethvargo

yolocs:
> Option 1: Versioned parser:

> Pro: each parser is easy to read
> Con: absolutely massive copypasta. Discourages versioning because each version incurs new copypasta.

At first glance, I'm definitely a fan of this. Could you help me understand why copypasta will be a must? Is it possible to refactor common yaml parse functionality out to be a common lib, and have each versioned parser use it?

sethvargo:
> > Option 1: Versioned parser:
> 
> > Pro: each parser is easy to read
> > Con: absolutely massive copypasta. Discourages versioning because each version incurs new copypasta.
> 
> At first glance, I'm definitely a fan of this. Could you help me understand why copypasta will be a must? Is it possible to refactor common yaml parse functionality out to be a common lib, and have each versioned parser use it?

I had a similar question. Could we do:

```go
package parser


type V1 struct {
  //...
}

func (v *V1) Thing()
func (v *V1) OtherThing()

type V2 struct {
  V1
}

// Only override OtherThing parsing.
func (v *V2) OtherThing()
```


> - Minor tweaks like this one that don't warrant a new revision of the models, where Option 3 is appropriate.
> - Major tweaks that DO warrant a new revision of the models, where Option 1 is appropriate.
> Can you live with that?

To be honest, the locked global feels _way_ more complex and error-prone than a little polymorphism. We can also aggressively unit test the polymorphism to handle edge cases. I hear your point about DevEx, but this is pretty common (most k8s things are []string in the data model, but accept single-string values and array-string values in the YAML).

drevell:
OK, I can accept the polymorphic approach. That doesn't require any special versioning code so I'll close this.

Issue #166, opened 2023-08-24
Template Subfolder convention in README/existing tempaltes


kevya-google:
### TL;DR

Establish a recommended convention of having template files that are intended to be included placed in a subfolder one directory below the spec.yaml. This way, other files that are repo-specific (such as .vscode and .gitignore) or metadata for the template won't need to be excluded as part of the spec.yaml (for example, imagine a README.md for describing how to use the template on the same level as the spec.yaml and another README.md in the template folder meant to be included in the render describing details about the template code). However, this shouldn't be strictly enforced, providing some flexibility to template authors as there could be some templates that have a need for follow up actions, different files for different environments, other layout requirements, etc.

Context: https://docs.google.com/document/d/12iN2g1Idz6fW5ZqaokTmSFZf5kHXaWeewULWy2q5GfE/edit#heading=h.z1xyeclrumgj

PR #171, opened 2023-08-24
Template Subfolder convention

kevya-google:
Addresses https://github.com/abcxyz/abc/issues/166 for existing templates in t folder.

Template files to be included are in a folder named "contents" (other options considered were "template", "render", "target", "src", "include")

Updates spec.yaml includes, usage example README, separates READMEs for each template. 
Also adds some default blank values for template inputs to ensure render works according to the READMEs.


README.md
```diff
@@ -185,7 +185,7 @@ https://github.com/hashicorp/go-getter library, including:
 In essence, a template is a directory or directory-like object containing a
 "spec file", named `spec.yaml`
 ([example](https://github.com/abcxyz/abc/blob/main/examples/templates/render/hello_jupiter/spec.yaml)),
-and other files such as source code and config files.
+and other files such as source code and config files. The recommended convention is to have all files intended to be included placed in a subfolder named "contents". This way, non-template related files (GitHub workflows, metadata, template READMEs, etc.) can exist next to the spec.yaml.
```
drevell, line 188: Let's make this a new paragraph

t/react_template/spec.yaml
```diff
@@ -18,25 +18,34 @@ desc: 'An example template using React library'
 inputs:
   - name: 'automation_service_account'
     desc: 'Automation Service Account (ex: [account_name]@[project_id].iam.gserviceaccount.com)'
+    default: ''
   - name: 'wif_provider'
     desc: 'Workload Identity Federation Provider (ex: projects/[project_id]/locations/global/workloadIdentityPools/[WIF_pool_name]/providers/[provider_name])'
+    default: ''
```
drevell, line 24: I dunno about this, template rendering will "succeed" but produce broken GH workflows.

drevell (changes requested):

t/react_template/spec.yaml
kevya-google: When we added the workflows, now inputs for these are required for this template to render.

I think eventually extracting out the workflows/Dockerfiles as discussed in [go/vendor-abcxyz-apps](http://goto.google.com/vendor-abcxyz-apps) is the best move.

But for now, this is blocking the template from being able to be rendered locally, without the workflows. The env file just looks like:

```
AUTOMATION_SERVICE_ACCOUNT=
WIF_PROVIDER=
AR_REPOSITORY=
AR_LOCATION=
CR_SERVICE=
REGION=
PROJECT_ID=
```

with the default blanks. Maybe I can change the defaults to comments of the value hints in the desc for each input?

README.md
kevya-google: moved to new paragraph.

kevya-google:
Also, if this is too large to review, I can split this up into smaller PRs, one for updating each template.

t/react_template/spec.yaml
drevell: I think this is a design problem that we can't just hack around this way. If we're going to include the workflows in the template, we should try hard to make them work. So I think we should either not have defaults or factor them out.

kevya-google: Ok, then I think maybe I'll prioritize factoring them out first before continuing with this PR. What do you think?

drevell: That sounds fine to me

kevya-google:
Will re-open and split up once .github workflows and Dockerfiles are factored out of existing tempaltes.

PR #172, opened 2023-08-24
Fix rest_server main.go errors, go mod tidy

kevya-google:
Fix rest_server slog errors from https://github.com/abcxyz/abc/pull/157.

Can merge to https://github.com/abcxyz/abc/tree/kevya/template-subfolder first or main if https://github.com/abcxyz/abc/pull/171 merges first.


t/rest_server/contents/main.go
```diff
@@ -94,8 +95,8 @@ func main() {
 
 	flag.Parse()
 	if err := realMain(logging.WithLogger(ctx, logger)); err != nil {
-		done() // deferred function calls won't execute due to logger.Fatal() never returning
-		logger.Fatal(err)
```
drevell, line 98: Shouldn't we log this error?

drevell (changes requested):

t/rest_server/contents/main.go
kevya-google: github.com/abcxyz/pkg/logging's logger doesn't have a Fatal. What should I do?

drevell: Well, the behavior of Fatal was to log then exit, so I think you can log at the most severe level available. And there's already a call to Exit() below.

kevya-google: Ok, I was going to do logging.LevelEmergency but I think LevelError sounded more appropriate.

drevell: I think logger.ErrorContext() is more conventional than logger.Log(..., logging.LevelError)

kevya-google: Got it, thank you!

drevell (approved):

PR #174, opened 2023-08-25
feat: CEL validation rules for template inputs

drevell:
Each input in spec.yaml can now optionally have a list of "rules" (CEL predicates) to check it for validity. An example use case is to check that a given string is suitable to be used as a GCP project ID.

Partially implements #28.

Example error message:

```
input validation failed:

Input name:   my_input
Input value:  foo
Rule:         my_input.startsWith("ham")
Rule msg:     Must start with "ham"
```

We support validating multiple inputs together. For example, you could require that the combined length of two strings is less than some number, or you could require that exactly one of two inputs is non-empty. Every input is in scope in every validation function.

We also include information about the validation rules when prompting for user inputs. This will help the user input something valid. Example prompt:

```
Input name:   animal
Description:  your favorite animal
Rule:         size(animal) > 1
Rule msg:     length must be greater than 1

Enter value: 
```

Upcoming work will add custom CEL functions like `gcp.matches_service_account` for commonly needed validations.

### DIFF
--- a/templates/commands/render.go
+++ b/templates/commands/render.go
@@ -373,6 +373,41 @@ func (c *RenderCommand) resolveInputs(ctx context.Context, spec *model.Spec) err
 		}
 	}
 
+	if err := c.validateInputs(ctx, spec.Inputs); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *RenderCommand) validateInputs(ctx context.Context, inputs []*model.Input) error {
+	scope := newScope(c.flags.Inputs)
+
+	sb := &strings.Builder{}
+	tw := tabwriter.NewWriter(sb, 8, 0, 2, ' ', 0)
+
+	for _, input := range inputs {
+		for _, rule := range input.Rules {
+			var ok bool
+			err := celCompileAndEval(ctx, scope, rule.Rule, &ok)
+			if ok && err == nil {
+				continue
+			}
+
+			fmt.Fprintf(tw, "\nInput name:\t%s", input.Name.Val)
+			fmt.Fprintf(tw, "\nInput value:\t%s", c.flags.Inputs[input.Name.Val])
+			writeRule(tw, rule, false, 0)
+			if err != nil {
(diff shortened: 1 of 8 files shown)


templates/commands/render.go
```diff
@@ -373,6 +373,41 @@ func (c *RenderCommand) resolveInputs(ctx context.Context, spec *model.Spec) err
...
+			fmt.Fprintf(tw, "\nInput value:\t%s", c.flags.Inputs[input.Name.Val])
+			writeRule(tw, rule, false, 0)
+			if err != nil {
+				fmt.Fprintf(tw, "\nCEL error:\t%s", err.Error())
+			}
+			fmt.Fprintf(tw, "\n") // Add vertical relief between validation messages
+		}
+	}
+
+	tw.Flush()
```
sethvargo, line 407: Check the returned error

templates/commands/render_flags.go
```diff
@@ -51,6 +51,9 @@ type RenderFlags struct {
 	// Inputs provide values that are substituted into the template. The keys in
 	// this map must match the input names in the Source template's spec.yaml
 	// file.
+	//
+	// This is mutable, even after flag parsing. It may be updated when default
+	// values are added and as the user is prompted for inputs.
```
sethvargo, line 56: For the future: I wonder if we need a way to indicate that inputs are "finalized". I can't think of a good reason right now.

sethvargo (approved):

templates/commands/render.go
drevell: OK, will follow up.

templates/commands/render_flags.go
drevell: I think the way I'd approach this more cleanly is to have this variable be immutable after parsing, and then keep track of the inputs-with-defaults-and-prompting as just a local variables. I just want to avoid having long-lived global-esque mutable state. But probably not needed right now, like you say.

Issue #39, opened 2023-06-20
Type-checking / validation for inputs


drevell:
### TL;DR

Currently, all template inputs (`--input=foo-bar`) are treated as strings. We should offer the ability to restrict inputs in certain ways, like "must be parseable as a positive integer" or "must be one of the choices foo, bar, baz." The inputs will still be strings, since we're just doing textual replacement within files, but we still want to know that the string represents a number/bool/whatever.

One possible idea for how this could look:
```
inputs:
  - name: 'concurrent_workers'
    required: false
    default: 10
    desc: 'The number of background workers that will serve requests'
    validation:
      - rule: positive_integer
```

Other validation rules could be:
- `go_bool`: matches `true` or `false`, and nothing else
- `tf_bool`: a string that's a valid terraform boolean (which happens to be the same as a go boolean)
- `integer`
- `float`
- `alphanumeric`
- `go_identifier`: only allows characters that are valid in a Go identifier

Custom:
```
 - validation:
      - rule: custom_regex
        - '[A-Za-z][0-9A-Za-z]+'
```
This could be used to check the length for things like project IDs.

We could also offer CEL validation as an alternative to regex:

```
 - validation:
      - rule: custom_cel
        - 'some_cel_expression' # must return true
``` 

Or we could just make the validation a Go template DSL. The curly braces could be implied, so the user doesn't actually write them. We could add our own functions into the FuncMap for the template.

LIke:
```
- validation:
  - rule: 'is_tf_bool .input'
```

Maybe we'd also want to support custom validation using regex or CEL (https://github.com/google/cel-go).

sethvargo:
My suggestion here is to explore CEL or a similar configuration language instead of doing a DIY here.

drevell:
Updated, thanks.

drevell:
This is now fully finished and documented. A new release v0.0.1-alpha8 has been released including this feature.

PR #175, opened 2023-08-28
feat: custom CEL functions for matching GCP service accounts

drevell:
These new CEL functions are intended to be used for input validation. They validate that their input matches the expected format. This PR adds functions for validating GCP service accounts and project IDs.

Part of #39.


pdewilde (approved):
neat!

sethvargo (approved):

PR #177, opened 2023-08-30
Add parser and validator for inputs.yaml, which will be used for golden test rendering.

chloechien:


templates/model/inputs.go
```diff
@@ -0,0 +1,77 @@
...
+
+type Inputs struct {
+	// Pos is the YAML file location where this object started.
+	Pos ConfigPos `yaml:"-"`
+
+	Inputs []*InputValue `yaml:"inputs"`
+}
+
+// DecodeInputs unmarshals the YAML Spec from r.
+func DecodeInputs(r io.Reader) (*Inputs, error) {
```
drevell, line 44: As far as I can see, this function is only used in tests. Should it go in a _test.go file?

templates/model/inputs.go
```diff
@@ -0,0 +1,77 @@
...
+// InputValue represents one of the parsed "input" fields from the inputs.yaml file.
+type InputValue struct {
+	// Pos is the YAML file location where this object started.
+	Pos ConfigPos `yaml:"-"`
+
+	Name  String `yaml:"name"`
+	Value String `yaml:"value"`
+}
+
+type Inputs struct {
```
drevell, line 36: Should we add `apiVersion` and `kind` here, like in spec.yaml? That will allow us to evolve the schema over time.

templates/model/inputs.go
```diff
@@ -0,0 +1,77 @@
...
+	"errors"
+	"fmt"
+	"io"
+
+	"gopkg.in/yaml.v3"
+)
+
+// This file parses a YAML file that describes template input.
+
+// InputValue represents one of the parsed "input" fields from the inputs.yaml file.
```
drevell, line 27: I know we agreed to naming this inputs.yaml in our meeting, but now I'm wondering if it should be something more general, like "test.yaml". If we ever decide to add fields to the YAML (like `test_description` or `test_flags`) then the name `inputs.yaml` would be misleading.

templates/model/inputs.go
drevell, line 36: By convention, all capitalized (exported) structs should have comments.

templates/model/inputs.go
```diff
@@ -0,0 +1,77 @@
...
+	var inputs Inputs
+	if err := dec.Decode(&inputs); err != nil {
+		return nil, fmt.Errorf("error parsing YAML inputs file: %w", err)
+	}
+
+	return &inputs, validateEach(inputs.Inputs)
+}
+
+// UnmarshalYAML implements yaml.Unmarshaler.
+func (i *Inputs) UnmarshalYAML(n *yaml.Node) error {
```
drevell, line 57: Typically we group the methods of a struct together next to the struct definition.

templates/model/inputs_test.go
```diff
@@ -0,0 +1,90 @@
...
+		want    *Inputs
+		wantErr string
+	}{
+		{
+			name: "simple_inputs_should_succeed",
+			in: `inputs:
+- name: 'person_name'
+  value: 'iron_man'
+- name: 'dog_name'
+  value: 'iron_dog'`,
```
drevell, line 41: 🤣

templates/model/pos_test.go
```diff
@@ -36,7 +36,7 @@ func TestPosErrorf(t *testing.T) {
 			pos:     &ConfigPos{10, 11},
 			fmtStr:  "Oh no! some number %d: %w",
 			args:    []any{345, fmt.Errorf("wrapped error")},
-			wantErr: "at spec.yaml line 10 column 11: Oh no! some number 345: wrapped error",
```
drevell, line 39: Maybe we need to add file information to ConfigPos. Before, there was only one YAML file so we could assume it was always `spec.yaml`. But now there's a second YAML file so perhaps we need to print its name. What do you think?

This might be very tricky to implement. Our YAML parsing library doesn't make it easy to get this information. So we might want to do it in a followup PR.

drevell (changes requested):

templates/model/pos_test.go
chloechien: Agree. As discussed off-line, we'll impl this in following PRs.

templates/model/inputs.go
chloechien: Got that, done!

templates/model/inputs.go
chloechien: Done

chloechien: Good idea, added!

templates/model/inputs.go
chloechien: We'll use that in test_record, test_render action in following up PRs. :)

templates/model/inputs.go
chloechien: Good call, done!

templates/model/inputs.go
drevell: Oh, of course.

drevell (approved):

templates/model/inputs.go
drevell: SGTM.

PR #178, opened 2023-08-30
Implement gcp_matches_project_number CEL func

drevell:
This will be used in spec.yaml input fields to verify that the user provided a GCP project number:

For #39 .

### DIFF
--- a/templates/commands/cel.go
+++ b/templates/commands/cel.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


templates/commands/cel.go
```diff
@@ -151,6 +155,55 @@ var celFuncs = []cel.EnvOption{
...
+				str, ok := strAny.(string)
+				if !ok {
+					return types.NewErr("internal error: argument was %T but should have been a string", strAny)
+				}
+
+				u64, err := strconv.ParseUint(str, 10, 64)
+				if err != nil {
+					return types.Bool(false) // A string that's not parseable as a uint is not a valid project number
+				}
+				return types.Bool(u64 >= minProjectNum)
```
sethvargo, line 186: Would it be better/easier to regex match on 0-9 or \d?

sethvargo (approved):

templates/commands/cel.go
drevell: Yeah, that would work, but it would mean that the implementation for `gcp_matches_project_number(string)` would be substantially different than `gcp_matches_project_number(int)`. It seems cleaner to have the two overloaded functions be as similar as possible. And I don't think we'd want to use regex for the int implementation; using a regex to check the magnitude of an integer seems gross.

If we were only providing one version of this function that accepted string, then I'd probably do it that way.

PR #181, opened 2023-08-31
Detect and handle old-style paths-as-list-of-strings

drevell:
We do YAML magic to detect when old-style input was provided (the `paths` field is a list of strings) and automatically parse it into new-style structs.

This is part of @kevya-google's work on supporting path objects. This provides backward compatibility with preexisting spec.yaml files.

### DIFF
--- a/templates/model/spec_test.go
+++ b/templates/model/spec_test.go
@@ -378,12 +378,40 @@ params: `,
 			wantValidateErr: `at spec.yaml line 1 column 1: field "message" is required`,
 		},
 		{
-			name: "include_old_style", // not path objects, paths are just strings
+			name: "include_success_paths_are_string", // not path objects, paths are just strings
 			in: `desc: 'mydesc'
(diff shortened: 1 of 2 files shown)


templates/model/spec_test.go
```diff
@@ -378,25 +378,23 @@ params: `,
 			wantValidateErr: `at spec.yaml line 1 column 1: field "message" is required`,
 		},
 		{
-			name: "include_success",
```
kevya-google, line 381: Instead of replacing the original include_success test, can we have include_old_style as a separate, additional test?

kevya-google (approved):
Thanks for handling this!

templates/model/spec_test.go
drevell: Sure, done. Added another test too that just occurred to me.

sethvargo (approved):

PR #186, opened 2023-09-01
Update Include Action to handle both strings and Path objects

kevya-google:
Mentioned in #130, @drevell, @sethvargo and I decided to support both the old method of string paths in addition to path objects for greater flexibility in configuring Include options.


drevell (dismissed):

templates/commands/render/render_action_include.go
```diff
@@ -28,6 +28,16 @@ import (
 )
 
 func actionInclude(ctx context.Context, inc *model.Include, sp *stepParams) error {
+	for _, path := range inc.Paths {
+		err := includePath(ctx, path, sp)
+		if err != nil {
+			return err
+		}
```
sethvargo, line 35: Nit:

```suggestion
		if err := includePath(ctx, path, sp); err != nil {
			return err
		}
```

sethvargo (approved):

drevell (approved):

PR #188, opened 2023-09-05
Refactor the `model` package into subpackages

drevell:
We used to just have one set of YAML models for the spec file. Recently we introduced another set of models for the golden test YAML files, and another one is coming for manifest files. This refactor will help avoid the model package becoming a mess.

The new structure is one package per YAML file: one package for spec files, one for test.yaml files, and (coming soon) one for manifest files.

All of the changes here are trivial: we're only moving files, changing capitalization, and changing comments.


chloechien (approved):
Thank you Dave!

sethvargo (approved):

yolocs (approved):

Issue #191, opened 2023-09-06
Template upgrades


drevell:
### TL;DR

Allow old template output directories to be upgraded when a new template version is released.

### Detailed design

```markdown
(copying portions of an internal design doc)

When the user runs abc templates render to render a template, we'll save a "manifest" file into the destination directory. This file will contain the information needed to cleanly upgrade to a new version of the template in the future. It contains the template location, inputs, and hashes of all the output files. Example:


# Generated by the `abc templates` command, do not modify
template_location: 'github.com/abcxyz/abc.git//t/template_name'
template_hash: '24bda78ab23...'  # Recursive directory hash of template
inputs:
  gcp_project_id: '12345'
  service_account: 'foo@bar'
hashes:
  - file: 'main.tf'
    hash: 'sha256:1a2b3c4d5e6f...'
  - file: 'terraform.tf'
    hash: 'sha256:968ad987e87fb9...'
```

The manifest file will be named .abc/manifest_<location>.lock.yaml.  It will be placed in the template destination directory (which is usually the CWD where the user runs abc templates ...).

We'll add a new command:

 templates upgrade [--force-overwrite] [--input=foo=bar] [--abort-on-conflict] [--location=github.com...] <manifest_filename>```

This will:

1. download the latest version of the template
1. if the template hasn't changed (its [dirhash](https://pkg.go.dev/golang.org/x/mod/sumdb/dirhash) is the same), then exit with a message
1. render the new template with the inputs that were saved in the manifest
1. merge the new template output with the existing files in the filesystem. For each file in the new template output:
   a. If there's no existing file at that path, just save the new output file.
   a. If there is an existing file at that path, but it hasn't been customized (the hash matches the hash in the manifest), then overwrite it with the new file.
   a. If the file HAS been customized (its hash doesn't match the manifest), then call this a "conflict." Tell the user to manually merge the existing/old file with the new file. We'll put the old/existing file at $FILE.old and the new file at $FILE.new...
   a. If there was a file that was generated by the old version of the template that is not generated by the new version of the template: delete it if it wasn't customized, or tell the user to handle the conflict if it has been customized.

Flags:
* --abort-on-conflict: if the upgrade can't apply cleanly due to customizations, abort rather than ask the user to resolve the conflict.
* --force-overwrite: overwrite all output files, even if they have been customized by the user.
* --input=foo=bar: two use cases:
  * provide any inputs that are required by the new template that weren't part of the original template
[...]

PR #194, opened 2023-09-07
Add YAML model structs for manifest files

drevell:
Part of #191.

These structs will represent the contents of the new "manifest" file, which contains all the information we'll need in the future to cleanly upgrade a template output.

This PR contains custom unmarshaling but not custom marshaling. That will be a separate PR.

Example manifest file:

```
api_version: 'cli.abcxyz.dev/v1alpha1'
template_location: 'github.com/abcxyz/abc.git//t/rest_server'
template_dirhash: 'h1:5891b5b522d5df086d0ff0b110fbd9d21bb4fc7163af34d08286a2e846f6be03'
inputs:
  - name: 'my_input_1'
    value: 'my_value_1'
  - name: 'my_input_2'
    value: 'my_value_2'
output_hashes:
  - file: 'a/b/c.txt'
    hash: 'h1:b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c'
  - file: 'd/e/f.txt'
    hash: 'h1:7d865e959b2466918c9863afca942d0fb89d7c9ac0c99bafc3749504ded97730'
```

Some of the field names have been changed since the design doc because the old names seemed unclear.

### DIFF
--- a/templates/model/manifest/manifest.go
+++ b/templates/model/manifest/manifest.go
@@ -1,0 +1,127 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package manifest
+
+import (
+	"errors"
+	"io"
+
+	"github.com/abcxyz/abc/templates/model"
+	"gopkg.in/yaml.v3"
+)
+
+// Decode unmarshals the YAML Spec from r. This function exists so we can
+// validate the Spec model before providing it to the caller; we don't want the
+// caller to forget, and thereby introduce bugs.
+//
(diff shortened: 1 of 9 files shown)


templates/model/manifest/manifest.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"errors"
+	"io"
+
+	"github.com/abcxyz/abc/templates/model"
+	"gopkg.in/yaml.v3"
+)
+
+// Decode unmarshals the YAML Spec from r. This function exists so we can
+// validate the Spec model before providing it to the caller; we don't want the
```
yolocs, line 26: I actually don't know if it's intentional
```suggestion
// Decode unmarshals the YAML Manifest from r. This function exists so we can
// validate the Manifest model before providing it to the caller; we don't want the
```

templates/model/manifest/manifest_test.go
```diff
@@ -0,0 +1,220 @@
...
+
+	cases := []struct {
+		name             string
+		in               string
+		want             *Manifest
+		wantUnmarshalErr string
+		wantValidateErr  string
+	}{
+		{
+			name: "simple-success",
```
yolocs, line 39: I think we always use underscore in test names

yolocs (commented):
nits

templates/model/manifest/manifest_test.go
drevell: Fixed

templates/model/manifest/manifest.go
drevell: Thanks for catching that, fixed.

sethvargo (approved):

PR #195, opened 2023-09-07
Advise the user to update if api_version is unknown

drevell:
We anticipate adding new features and YAML fields. When a template uses these new fields but the user is using an old binary, then there will be error about the api_version being unrecognized. This error might be confusing, so now we'll also tell the user to go update their binary.

### DIFF
--- a/templates/model/validate_test.go
+++ b/templates/model/validate_test.go
@@ -1,0 +1,64 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


sethvargo (dismissed):

yolocs (approved):

yolocs (approved):

templates/model/validate_test.go
```diff
@@ -0,0 +1,64 @@
...
+	for _, tc := range cases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			got := IsKnownAPIVersion(nil, String{Val: tc.in}, "api_version")
+			for _, wantErr := range tc.wantErr {
+				if diff := testutil.DiffErrString(got, wantErr); diff != "" {
+					t.Fatal(diff)
```
yolocs, line 59:
```suggestion
					t.Error(diff)
```

yolocs (approved):

sethvargo (approved):

templates/model/validate_test.go
drevell: Skipping this suggestion just because I don't want to lose a couple hours to getting this PR re-reviewed, and I think it's a pretty minor point.

PR #196, opened 2023-09-11
Add simple file globbing to walkAndModify (all actions besides Include)

### LINKED ISSUE #143: Add template expansion to walkAndModify to reduce boilerplate in actions.
### TL;DR

Currently, if you want template expansion to user-provided paths for your action, it must be implemented within the action. This proposal would be to move it into the walkAndModify function itself.

### Detailed design

```markdown
`walkAndModify` would have paths passed into the `relPaths` argument would have the following mutation done:


	paths := make([]model.String, 0, len(relPaths))
	for _, p := range relPaths {
		path, err := parseAndExecuteGoTmpl(p.Pos, p.Val, inputs)
		if err != nil {
			return err
		}
		paths = append(paths, model.String{Pos: p.Pos, Val: path})
	}


A new `inputs` argument would need to be added, which would `inputs map[string]string` and would come from the `stepParams` struct which gets passed to each action.

This would reduce boilerplate in authoring actions while also reducing bugs such as #142.
```

### Alternatives considered

```markdown
## Status Quo:

Keeps the `walkAndModify` method signature and implementation from becoming too complex. Easier for implementer to understand what their action is doing and how the paths argument gets used.

## Template All Inputs

Another idea would be to execute templates on all arguments before they get to the action code. This could have problems if there are strings that shouldn't have templates applied, and makes escaping more characters necessary.
```

### Additional information

_No response_

kevya-google:
Adds functionality to parse globs in paths after processPaths, used in walkAndModify (every action apart from Include) + relevant unit tests. Also small refactor to move processPaths() out from all actions to inside walkAndModify.

Removes safeRelPath and stat file existence checking from walkAndModify tests since stat is no longer needed because glob paths will only match files that exist and safeRelPath already happens as part of processPaths().

processGlobs() to be added to Include Action in a follow-up PR (splitting as the logic is somewhat complex to fit all in one).

Also closes #143 as go template processing has been moved out to processPaths().

co-authored by: @drevell


templates/commands/render/render_action_include.go
```diff
@@ -68,6 +66,23 @@ func includePath(ctx context.Context, inc *spec.IncludePath, sp *stepParams) err
...
 
+		// By default, we copy from the template directory. We also support
+		// grabbing files from the destination directory, so we can modify files
+		// that already exist in the destination.
+		fromDir := sp.templateDir
+		if inc.From.Val == "destination" {
+			fromDir = sp.flags.Dest
+		}
+
+		absSrcPaths, err := filepath.Glob(filepath.Join(fromDir, walkRelPath))
```
drevell, line 77: For a followup PR, it would be cool to add a debug log saying eg `glob *.txt expanded to [a.txt, b.txt]`

templates/commands/render/render_action_include_test.go
```diff
@@ -58,6 +58,67 @@ func TestActionInclude(t *testing.T) {
...
+				"dont-include-hello.txt": {0o600, "dont include hello file contents"},
+			},
+			wantScratchContents: map[string]modeAndContents{
+				"hello_its_me.txt": {0o600, "hello its me file contents"},
+				"hellooo.txt":      {0o600, "hellooo file contents"},
+				"hello-test.txt":   {0o600, "hello-test file contents"},
+			},
+		},
+		{
+			name: "file_extension_globbing",
```
drevell, line 103: Maybe another test case where the glob matches a directory name? Like `a/*/b.txt`

drevell (approved):

templates/commands/render/render_action_include.go
sethvargo, line 77: Do we need to clean the result of the join with `filepath.Clean`? I think this would theoretically allow someone to execute a template outside of their PWD - is that desired?

templates/commands/render/render_action_include_test.go
sethvargo: It's also probably worth noting that `a/*/b.txt` will include `a/foo/b.txt` and `a/bar/b.txt`, but **not** `a/foo/bar/b.txt` since `*` matches any sequence of **non-Separator** characters

templates/commands/render/render_action_include.go
drevell: We're already protected from this by the call to `safeRelPath` above.

kevya-google:
Converting to draft until other PR changes are in (filepath processing)

templates/commands/render/render_action_include.go
kevya-google: Added debug logging to processGlobs.

templates/commands/render/render_action_include_test.go
kevya-google: Test case star_glob_paths covers directory matching.

templates/commands/render/action.go
```diff
@@ -47,26 +47,26 @@ type walkAndModifyVisitor func([]byte) ([]byte, error)
 //
 // If the visitor returns modified file contents for a given file, that file
 // will be overwritten with the new contents.
-func walkAndModify(ctx context.Context, rfs renderFS, scratchDir string, relPaths []model.String, v walkAndModifyVisitor) error {
+//
+// rawPaths is expected to have been cleaned already from processPaths.
+func walkAndModify(ctx context.Context, sp *stepParams, rawPaths []model.String, v walkAndModifyVisitor) error {
 	logger := logging.FromContext(ctx).With("logger", "walkAndModify")
 	seen := map[string]struct{}{}
+	rfs := sp.fs
```
drevell, line 55: Can we eliminate the `rfs` variable and inline `sp.fs` everywhere that rfs was used? Same for `scratchDir` and `sp.scratchDir`

templates/commands/render/action.go
```diff
@@ -47,26 +47,26 @@ type walkAndModifyVisitor func([]byte) ([]byte, error)
...
+		return err
+	}
+	globbedPaths, err := processGlobs(ctx, paths, sp.scratchDir)
+	if err != nil {
+		return err
+	}
 
-		err = filepath.WalkDir(walkFrom, func(path string, d fs.DirEntry, err error) error {
+	for _, absPath := range globbedPaths {
+		pos := absPath.Pos
```
drevell, line 68: Same here

templates/commands/render/action.go
```diff
@@ -47,26 +47,26 @@ type walkAndModifyVisitor func([]byte) ([]byte, error)
 //
 // If the visitor returns modified file contents for a given file, that file
 // will be overwritten with the new contents.
-func walkAndModify(ctx context.Context, rfs renderFS, scratchDir string, relPaths []model.String, v walkAndModifyVisitor) error {
+//
+// rawPaths is expected to have been cleaned already from processPaths.
```
drevell, line 51: Maybe update the docs for walkAndModify to mention globs

templates/commands/render/action.go
```diff
@@ -151,6 +181,15 @@ func processPaths(paths []model.String, scope *common.Scope) ([]model.String, er
 		if err != nil {
 			return nil, err
 		}
+
+		// on Windows, escaping is disabled in glob parsing. Therefore, to keep
+		// glob parsing behavior consistent across different OSes, escaping is
+		// not permitted. This is checked in processPaths because FromSlash converts
+		// separators to backslashes for Windows.
+		if strings.Contains(p.Val, `\`) {
+			return nil, p.Pos.Errorf(`escaping glob paths is not permitted: %q`, p.Val)
```
drevell, line 190: Maybe we can make this error message a little more helpful by adding the word "backslashes" somewhere, like "backslashes aren't allowed in path globs"

drevell (dismissed):
Some small notes

templates/commands/render/action.go
kevya-google: Removed and inlined.

templates/commands/render/action.go
kevya-google: Removed and inlined.

templates/commands/render/action.go
kevya-google: Updated to mention processGlobs.

templates/commands/render/action.go
kevya-google: Added backslashes to error message.

templates/commands/render/action.go
```diff
@@ -47,29 +47,27 @@ type walkAndModifyVisitor func([]byte) ([]byte, error)
 //
 // If the visitor returns modified file contents for a given file, that file
 // will be overwritten with the new contents.
-func walkAndModify(ctx context.Context, rfs renderFS, scratchDir string, relPaths []model.String, v walkAndModifyVisitor) error {
+//
+// paths are expected to be cleaned already from processPaths before being
+// passed into processGlobs.
```
drevell, line 52: I was thinking of a comment like "rawPaths is a list of globs"

drevell (approved):

templates/commands/render/action.go
kevya-google: Oh, I see. It could also be go template strings or non-glob strings. What about:
"rawPaths is a list of path strings that need to be processed (processPaths, processGlobs) before walking through."

drevell: SGTM. Maybe "will be" rather than "need to be".

kevya-google: Updated comment, rewrapped.

drevell (approved):

PR #202, opened 2023-09-11
Rename test cmd to golden-test

chloechien:


templates/commands/golden_test/test_record.go
```diff
@@ -12,11 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package test implements the template test related subcommands.
-package test
+// Package golden_test implements golden test related subcommands.
+package golden_test
```
drevell, line 16: By convention, package names don't contain underscores.

drevell (changes requested):

templates/commands/golden_test/test_record.go
chloechien: Got that! updated to goldentest, same for folder name.

drevell (approved):

drevell (approved):

PR #208, opened 2023-09-13
Add Path Processing for all actions using Paths

kevya-google:
Processes Paths from actions applying Go templating, adjusting for safe relative paths, and changing to OS-specific slashes. Also eventually simple file globbing in https://github.com/abcxyz/abc/pull/196.


templates/commands/render/action.go
```diff
@@ -141,6 +141,29 @@ func templateAndCompileRegexes(regexes []model.String, scope *common.Scope) ([]*
...
+	for i, p := range pathsCopy {
+		p.Val, err = parseAndExecuteGoTmpl(p.Pos, p.Val, sp.scope)
+		if err != nil {
+			return nil, err
+		}
+		p.Val, err = safeRelPath(p.Pos, p.Val)
+		if err != nil {
+			return nil, err
+		}
+		p.Val = filepath.FromSlash(p.Val)
```
drevell, line 160: Hm I think FromSlash should happen before safeRelPath. I think we want safeRelPath to see the true local-OS-style path.

templates/commands/render/action.go
```diff
@@ -141,6 +141,29 @@ func templateAndCompileRegexes(regexes []model.String, scope *common.Scope) ([]*
 	return compiled, merr
 }
 
+// processPaths processes a list of input String paths for go templating, relative paths,
+// OS-specific slashes, and (soon) file globbing.
+func processPaths(paths []model.String, sp *stepParams) ([]model.String, error) {
```
drevell, line 146: Maybe just pass the sp.scope here instead of the entire sp?

templates/commands/render/action_include.go
```diff
@@ -36,36 +36,32 @@ func actionInclude(ctx context.Context, inc *spec.Include, sp *stepParams) error
...
+		return err
+	}
+
+	for i, p := range incPaths {
 		// During validation in spec.go, we've already enforced that either:
 		//  - len(inc.As) == 0
 		//  - len(inc.As) == len(inc.Paths)
-		as := walkRelPath
+		as := p.Val
 		if len(inc.As) > 0 {
```
drevell, line 63: Maybe len(asPaths) would be a bit readable than len(inc.As), even though they're the same length

drevell (changes requested):
I think we need some tests here. The globbing of the "skip" paths seems especially ripe for bugs and therefore seems like it should be thoroughly tested.

kevya-google:
> I think we need some tests here. The globbing of the "skip" paths seems especially ripe for bugs and therefore seems like it should be thoroughly tested.

Tests pending.

templates/commands/render/action.go
kevya-google: Reordered.

templates/commands/render/action.go
kevya-google: Changed to only pass scope.

templates/commands/render/action_include.go
kevya-google: Good point, changed to asPaths.

templates/commands/render/action.go
kevya-google: Reordering this seems to break tests for windows: https://github.com/abcxyz/abc/actions/runs/6190617112/job/16807205539?pr=208

kevya-google: Reordered, fixed test.

templates/commands/render/action_include_test.go
```diff
@@ -83,7 +83,7 @@ func TestActionInclude(t *testing.T) {
 					},
 				},
 			},
-			wantErr: `path "../file.txt" must not contain ".."`,
+			wantErr: fmt.Sprintf("path %q must not contain \"..\"", filepath.FromSlash("../file.txt")),
```
drevell, line 86: When a string has quotes in in, the convention is to use a raw string with backticks so you don't have to escape the quotes.

So rather than this: 

```
"path %q must not contain \"..\""
```

one could do this:

```
`path %q must not contain ".."`
```

templates/commands/render/action_include_test.go
```diff
@@ -260,6 +260,47 @@ func TestActionInclude(t *testing.T) {
...
+				"subfolder/file3.txt": {0o600, "file 3 contents"},
+				"spec.yaml":           {0o600, "spec contents"},
+			},
+			wantScratchContents: map[string]modeAndContents{
+				"file1.txt":           {0o600, "file 1 contents"},
+				"subfolder/file3.txt": {0o600, "file 3 contents"},
+			},
+		},
+		{
+			name: "skip_file_in_subfolder",
```
drevell, line 285: Do we need tests for skipping directories and skipping multiple files

drevell (changes requested):

templates/commands/render/action_include_test.go
kevya-google: Changed to backticks.

templates/commands/render/action_include_test.go
kevya-google: Added tests for these.

templates/commands/render/action.go
```diff
@@ -141,6 +141,29 @@ func templateAndCompileRegexes(regexes []model.String, scope *common.Scope) ([]*
...
+	copy(pathsCopy, paths)
+
+	var err error
+	for i, p := range pathsCopy {
+		p.Val, err = parseAndExecuteGoTmpl(p.Pos, p.Val, scope)
+		if err != nil {
+			return nil, err
+		}
+		p.Val = filepath.FromSlash(p.Val)
+		p.Val, err = safeRelPath(p.Pos, p.Val)
```
sethvargo, line 157: I'd be helpful to comment this. Why do we need to call FromSlash before safeRelPath?

templates/commands/render/action.go
```diff
@@ -141,6 +141,29 @@ func templateAndCompileRegexes(regexes []model.String, scope *common.Scope) ([]*
...
 
+// processPaths processes a list of input String paths for go templating, relative paths,
+// OS-specific slashes, and (soon) file globbing.
+func processPaths(paths []model.String, scope *common.Scope) ([]model.String, error) {
+	pathsCopy := make([]model.String, len(paths))
+	copy(pathsCopy, paths)
+
+	var err error
+	for i, p := range pathsCopy {
+		p.Val, err = parseAndExecuteGoTmpl(p.Pos, p.Val, scope)
```
sethvargo, line 152: The calls to `p.Val = ...` are only safe because the input is not a pointer (`[]model.String` vs `[]*model.String`). If that ever changes to a pointer, this would be modifying the result in place.

It might be safer and more idiomatic to either:

1. Accept the slice of pointers, deep-copy each element, return a new slice
2. Accept the slice of pointers, modify in place, return only an error (because the provided slice will have been modified).

I'm just a little worried about the probability of a future bug here given the implicit dependency on copy-by-value.

Non-blocking.

templates/commands/render/action.go
```diff
@@ -141,6 +141,29 @@ func templateAndCompileRegexes(regexes []model.String, scope *common.Scope) ([]*
...
+		p.Val = filepath.FromSlash(p.Val)
+		p.Val, err = safeRelPath(p.Pos, p.Val)
+		if err != nil {
+			return nil, err
+		}
+		pathsCopy[i].Val = p.Val
+	}
+
+	return pathsCopy, nil
+}
```
sethvargo, line 165: Style nit: I think it's generally more common to create an empty slice and then append values while iterating over the given slice (instead of the current implementation which makes a copy of the slice and modifies it in place):

```go
	pathsCopy := make([]model.String, 0, len(paths))
	for _, p := range paths {
    val, err := parseAndExecuteGoTmpl(p.Pos, p.Val, scope)
		if err != nil {
			return nil, err
		}

    val, err = safeReplPath(p.Pos, filepath.FromSlash(val))
		if err != nil {
			return nil, err
		}

		pathsCopy = append(pathsCopy, model.String{
			Val: val,
			Pos: p.Pos,
			// ...
		})
	}

	return pathsCopy, nil
```

templates/commands/render/action_include.go
```diff
@@ -36,36 +36,32 @@ func actionInclude(ctx context.Context, inc *spec.Include, sp *stepParams) error
...
-			}
-		}
-
-		relDst, err := safeRelPath(p.Pos, as)
-		if err != nil {
-			return err
+		as := p.Val
+		if len(asPaths) > 0 {
+			as = asPaths[i].Val
 		}
```
sethvargo, line 65: Is this correct? Can we guarantee that `len(asPaths) >= len(incPaths)`? Or should this be `len(asPaths) > i`?

templates/commands/render/action.go
```diff
@@ -141,6 +141,29 @@ func templateAndCompileRegexes(regexes []model.String, scope *common.Scope) ([]*
 	return compiled, merr
 }
 
+// processPaths processes a list of input String paths for go templating, relative paths,
+// OS-specific slashes, and (soon) file globbing.
+func processPaths(paths []model.String, scope *common.Scope) ([]model.String, error) {
```
sethvargo, line 146: Can we add a test that ensures the given arguments are not modified? Basically make a copy of `paths`, call the function, ensure paths has not changed.

templates/commands/render/action.go
kevya-google: Changed to append to empty slice.

templates/commands/render/action.go
kevya-google: Changed to append to empty slice without modifying in place.

templates/commands/render/action_include.go
kevya-google: Updated comment assertion. See validation function in spec.go for IncludePath.

templates/commands/render/action.go
kevya-google: Added test for this + some extras for processPaths.

templates/commands/render/action.go
kevya-google: This is for verifying that safeRelPath processes a path with the OS-specific correct paths, they could be combined into one function as suggested by @drevell but I think it might be ok to keep as-is. What do you think about combining?

drevell (approved):

sethvargo (approved):

PR #209, opened 2023-09-14
Add boilerplate for nextjs react template

chloechien:
Add a boilderplate for nextjst react template. More input params (e.g. auth0 integration) incoming.

- Template initially generated by npx create-next-app@latest
- Tailwind css is replaced with plain css + MUI - instead of designing Tailwind+MUI base with freedom, we wanna stay with Material UI
- Next Eslint is replaced with GTS eslint/prettier


mattrandallbecker (approved):
Looks good, one thing in a followup PR should be linting + running tests workflows in the .github directory, then it's even more value add

+maybe something that nags if they use dangerouslysetinnerhtml without dompurify

chloechien:
Yes we'll need the content security policy :)

rachelcao-g (approved):

PR #222, opened 2023-10-04
Add golden test render command implementations.

chloechien:
Will add test.yaml version check in a separated PR.

### DIFF
--- a/templates/commands/goldentest/record.go
+++ b/templates/commands/goldentest/record.go
@@ -75,7 +75,6 @@ func (c *RecordCommand) Run(ctx context.Context, args []string) error {
 	if err != nil {
 		return fmt.Errorf("failed to create temporary directory: %w", err)
 	}
-
 	var merr error
 	for _, tc := range testCases {
 		merr = errors.Join(merr, renderTestCase(c.flags.Location, tempDir, tc))
@@ -83,11 +82,11 @@ func (c *RecordCommand) Run(ctx context.Context, args []string) error {
 	if merr != nil {
 		return fmt.Errorf("failed to render golden tests: %w", merr)
 	}
-
 	if err := os.RemoveAll(tempDir); err != nil {
 		return fmt.Errorf("failed to remove temporary directory: %w", err)
 	}
 
+	// Render outputs in template golden test directories.
 	logger := logging.DefaultLogger()
 	for _, tc := range testCases {
 		if err := renderTestCase(c.flags.Location, c.flags.Location, tc); err != nil {
--- a/templates/commands/goldentest/record.go
+++ b/templates/commands/goldentest/record.go
@@ -71,10 +71,12 @@ func (c *RecordCommand) Run(ctx context.Context, args []string) error {
 	// Create a temporary directory to validate golden tests rendered with no
 	// error. If any test fails, no data should be written to file system
 	// for atomicity purpose.
-	tempDir, err := getTempDir()
+	tempDir, err := os.MkdirTemp("", "abc-test-*")
 	if err != nil {
 		return fmt.Errorf("failed to create temporary directory: %w", err)
 	}
+	defer os.RemoveAll(tempDir)
+
 	var merr error
 	for _, tc := range testCases {
 		merr = errors.Join(merr, renderTestCase(c.flags.Location, tempDir, tc))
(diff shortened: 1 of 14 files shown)


templates/commands/goldentest/record.go
```diff
@@ -60,9 +63,37 @@ func (c *RecordCommand) Run(ctx context.Context, args []string) error {
...
+		return fmt.Errorf("failed to create temporary directory: %w", err)
+	}
+	var merr error
+	for _, tc := range testCases {
+		merr = errors.Join(merr, renderTestCase(c.flags.Location, tempDir, tc))
+	}
+	if merr != nil {
+		return fmt.Errorf("failed to render golden tests: %w", merr)
+	}
+	if err := os.RemoveAll(tempDir); err != nil {
```
drevell, line 85: Should this os.Remove() be in a `defer` so that it runs even if there's an error?

templates/commands/goldentest/record.go
```diff
@@ -60,9 +63,37 @@ func (c *RecordCommand) Run(ctx context.Context, args []string) error {
...
+		return fmt.Errorf("failed to remove temporary directory: %w", err)
+	}
+
+	// Render outputs in template golden test directories.
+	logger := logging.DefaultLogger()
```
drevell, line 90: Perhaps the logging style should be consistent with other places in the abc cli, see https://github.com/search?q=repo%3Aabcxyz%2Fabc%20logging&type=code

templates/commands/goldentest/record.go
```diff
@@ -60,9 +63,37 @@ func (c *RecordCommand) Run(ctx context.Context, args []string) error {
...
+	for _, tc := range testCases {
```
drevell, line 91: Rather than running each test twice, how about just running each one once and committing the output into the final destination once they've all succeeded? We could use the existing `copyRecursive()` function for this.

templates/commands/goldentest/record_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/abcxyz/abc/templates/common"
+	"github.com/abcxyz/pkg/logging"
+	"github.com/abcxyz/pkg/testutil"
+	"github.com/google/go-cmp/cmp"
+)
+
+//nolint:paralleltest // This test cannot be run in parallel
```
drevell, line 29: Could you add more explanation to this comment explaining why it can't run in parallel?

templates/commands/goldentest/record_test.go
```diff
@@ -0,0 +1,166 @@
...
+				filepath.Join("test2", "a.txt"):     "file A content",
+			},
+		},
+		{
+			name: "oudated_golden_file_removed",
+			filesContent: map[string]string{
+				"spec.yaml":                        specYaml,
+				"a.txt":                            "file A content",
+				"testdata/golden/test/test.yaml":   testYaml,
+				"testdata/golden/test/oudated.txt": "oudated file",
```
drevell, line 86: oudated -> outdated, throughout this file

templates/commands/goldentest/test_funcs.go
```diff
@@ -105,3 +113,71 @@ func parseTestConfig(path string) (*goldentest.Test, error) {
 	}
 	return test, nil
 }
+
+// getTempDir returns a temporary directory for golden tests usage.
+func getTempDir() (string, error) {
```
drevell, line 118: Could we use os.MkdirTemp() for this? If not, could you add a comment explaining why?

templates/commands/goldentest/test_funcs.go
```diff
@@ -105,3 +113,71 @@ func parseTestConfig(path string) (*goldentest.Test, error) {
...
+	}
+
+	args := []string{"--dest", testDir, "--force-overwrite"}
+	for _, input := range tc.TestConfig.Inputs {
+		args = append(args, "--input")
+		args = append(args, fmt.Sprintf("%s=%s", input.Name.Val, input.Value.Val))
+	}
+	args = append(args, templateDir)
+
+	r := &render.Command{}
```
drevell, line 171: I'm not sure we should be using render.Command for this. Maybe we should factor out the rendering logic as a library, so we can call it like `render.Render(dest, location, inputs)`, without having to go through flags. Maybe this could be a TODO.

drevell (changes requested):

templates/commands/goldentest/test_funcs.go
chloechien: Oh nice suggestion! I simply don't know this func :)

templates/commands/goldentest/record.go
chloechien: Done.

templates/commands/goldentest/record_test.go
chloechien: Removed - now it can be paralleled. Previously they use the same temp dir.

templates/commands/goldentest/record_test.go
chloechien: Thank you

templates/commands/goldentest/record.go
chloechien: Done.

templates/commands/goldentest/test_funcs.go
chloechien: I've leaved a TODO

templates/commands/goldentest/record.go
chloechien: Good idea. Taking a look at copyRecursive we'll need to refactor a lot of rendering pkg to extract a shared renderFS into common folder. Leave a todo and will do refactor after this PR.

drevell: SGTM

PR #226, opened 2023-10-05
Use recursive file copy to render golden testdata to destination inst…

chloechien:
…ead of running rendering comments twice

### DIFF
--- a/templates/commands/goldentest/record.go
+++ b/templates/commands/goldentest/record.go
@@ -21,8 +21,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
+	"path/filepath"
 
+	"github.com/abcxyz/abc/templates/common"
 	"github.com/abcxyz/pkg/cli"
 	"github.com/abcxyz/pkg/logging"
 )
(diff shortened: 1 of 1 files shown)


templates/commands/goldentest/record.go
```diff
@@ -85,14 +88,33 @@ func (c *RecordCommand) Run(ctx context.Context, args []string) error {
...
+		if err := clearTestDir(testDir); err != nil {
+			return fmt.Errorf("failed to clear test directory: %w", err)
+		}
+
+		visitor := func(relToAbsSrc string, de fs.DirEntry) (common.CopyHint, error) {
+			if !de.IsDir() {
+				logger.InfoContext(ctx, "recording", "testname", tc.TestName, "testdata", relToAbsSrc)
+			}
+			return common.CopyHint{
+				Overwrite: true,
```
drevell, line 105: Is there an edge case here? What if the template coincidentally outputs a file named test.yaml? Would that conflict with the goldentest test.yaml?

It's not really specific to this PR, I just realized this while reviewing.

drevell (approved):

templates/commands/goldentest/record.go
chloechien: We're adopting the simplest solution at the moment - reserve test.yaml same as spec.yaml.

Or another way is
testdata/golden/testname/test.yaml
testdata/golden/testname/data/....

WDYT?

drevell: I think it's different than spec.yaml. spec.yaml is reserved in the template *source* directory, not in the output. The user can still output a file named spec.yaml if they want to, they'll just have to do something like `include renamed_spec.yaml as spec.yaml`. With test.yaml, it's different, because there's no workaround for a reserved filename in the template output, it's just impossible to output a ./test.yaml file.

Your idea for a `data` directory sounds good to me 👍

chloechien: Good catch in input/output difference! I'll update to data/ structure :)

Issue #220, opened 2023-09-21
Conditional execution of steps with `if`


drevell:
### TL;DR

We could add the ability to conditionally execute one or more template steps using a CEL predicate.

+ @verbanicm 

### Detailed design

```markdown
There are two possibilities we can think of for how this might work. There could be an `if` field on every action, or there could be a new action type `if`.

Design 1: an `if` field on every action


inputs:
  - name: 'include_auth0'

steps:
  - action: 'include'
    if: 'bool(include_auth0)'
    desc: 'conditionally include auth0 integration'
    params: 
      paths: ['auth0']
```

Design 2: a new action type for `if` that has a list of steps:

```
inputs:
  - name: 'include_auth0'

steps:
 - action: `if`
   params:
     then:
       - action: 'include'
          expr: 'bool(include_auth0)'
          desc: 'conditionally include auth0 integration'
          params: 
            paths: ['auth0']
```
```


### Alternatives considered

_No response_

### Additional information

_No response_

PR #228, opened 2023-10-10
Support multiple api_versions in the same codebase

drevell:
Apologies for the large PR, but I think in this case it's important to present all the pieces at the same time.

This PR creates a system for supporting YAML files using different `api_version`s from a single binary. This is implemented by having a separate set of structs for each `(kind,api_version)` pair. Each api_version can change the semantics or existence of YAML fields. The structs are stored in `templates/model/$KIND/$APIVERSION`.

Instructions for adding a new `api_version` are in `templates/model/README.md`, that might be a good starting point for review.

Old versions are automatically converted to new versions before being used: old structs define an Upgrade() method that define how to upgrade them.

There's a registry of `api_version`s in decode.go that defines which versions exist and which structs are associated with each.

To demonstrate that this all actually works, this PR creates a new API version `v1beta1`. The only change in this new version is adding the optional `if` field for each step in spec.yaml (#220).

Alternatives considered to try to avoid this huge copy-pasting of structs:
- Struct embedding to reduce copy-pasta: this would expose a more complicated API to the rest of the code that has to use the structs. It makes creating a new version more painful. It also triggers https://github.com/golang/go/issues/15924 since we use `reflect.StructOf`.
- Use http://github.com/mitchellh/mapstructure to avoid so much double-parsing: we'd lose YAML position information, and our users really appreciate getting line numbers on their error messages.
- Use `switch(apiVersion)` within a single struct instead of copy-pasting structs: high risk of accidentally changing the behavior of old API versions and generally creating fragile code.
 
I've marked files that are mostly copy-pasted so you can mostly skip reviewing them


templates/model/spec/v1beta1/spec.go
drevell, line 1: This file was copied from v1alpha1/spec.go. It just adds the "if" field in `type Step`.

templates/model/spec/v1beta1/spec.go
```diff
@@ -0,0 +1,560 @@
...
+	return model.NotZeroModel(&i.Pos, i.Rule, "rule")
+}
+
+// Step represents one of the work steps involved in rendering a template.
+type Step struct {
+	// Pos is the YAML file location where this object started.
+	Pos model.ConfigPos `yaml:"-"`
+
+	Desc   model.String `yaml:"desc"`
+	If     model.String `yaml:"if"`
```
drevell, line 112: This is the line that was added.

templates/model/spec/v1beta1/upgrade.go
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1beta1
+
+import "github.com/abcxyz/abc/templates/model"
+
+func (s *Spec) Upgrade() (model.ValidatorUpgrader, error) {
+	return nil, nil // This is the most recent version
```
sethvargo, line 20: It might be helpful to output a debug log in this case.

templates/model/unmarshal.go
```diff
@@ -102,3 +102,17 @@ func DecodeAndValidate(r io.Reader, filename string, outPtr Validator) error {
...
+
+// ValidatorUpgrader is the interface implemented by every kind of YAML struct
+// (templates, golden tests, manifests, etc).
+type ValidatorUpgrader interface {
+	Validator
+
+	// Upgrade converts an old version of a YAML struct into newer version for
+	// example, from api_version v1 to v2. If the struct is already the most
+	// recent version, it returns nil. An error means that the model cannot be
+	// converted because either (1) something weird happened or (2) the YAML
```
sethvargo, line 114: Nit: please format comment as a markdown list since modern godoc will convert it to an ol.

templates/model/spec/v1alpha1/upgrade.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"fmt"
+
+	"github.com/abcxyz/abc/templates/model"
+	"github.com/abcxyz/abc/templates/model/spec/v1beta1"
+	"github.com/jinzhu/copier"
+)
+
+func (s *Spec) Upgrade() (model.ValidatorUpgrader, error) {
+	out := &v1beta1.Spec{}
```
sethvargo, line 26: Style nit:

```suggestion
	var out v1beta1.Spe
```

and then `&out`

templates/model/decode/decode.go
```diff
@@ -0,0 +1,185 @@
...
+	if err := vu.Validate(); err != nil {
+		return nil, fmt.Errorf("validation failed in %s: %w", filename, err)
+	}
+
+	for {
+		upgraded, err := vu.Upgrade()
+		if err != nil {
+			return nil, fmt.Errorf("internal error: YAML model couldn't be upgraded from api_version %s: %w", apiVersion, err)
+		}
+		if upgraded == nil {
```
sethvargo, line 138: I worry about the sentinel value "nil". Would it be better to have a specific error type that we can check instead? This is similar to how io returns `io.EOF` when it reaches the end of the file.

templates/model/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+In any of the following cases:
+
+- You're adding a new field to an existing YAML struct
+- You're changing the semantics or interpretation of an existing yaml field
+- You're adding a new type of YAML file
+
+Every api_version represents a distinct configuration "language", and we never
+want to change the semantics of an existing api_version that has already been
+release.
```
verbanicm, line 15:
```suggestion
released.
```

templates/model/decode/decode.go
```diff
@@ -0,0 +1,185 @@
...
+			KindGoldenTest: &goldentestv1alpha1.Test{},
+			KindManifest:   &manifestv1alpha1.Manifest{},
+		},
+	},
+	{
+		apiVersion: "cli.abcxyz.dev/v1beta1",
+		kinds: map[string]model.ValidatorUpgrader{
+			KindTemplate:   &specv1beta1.Spec{},
+			KindGoldenTest: &goldentestv1alpha1.Test{},
+			KindManifest:   &manifestv1alpha1.Manifest{},
```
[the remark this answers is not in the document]
verbanicm: Maybe these should be separate structs, in different files and we only add individual api version changes? This would make it clearer about which versions are supported by which `Kinda`. This would also make it "easier" to automate later by targeting a specific file per kind.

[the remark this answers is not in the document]
verbanicm: Spoke offline, the current code seems like the best solution currently.

templates/model/spec/v1beta1/upgrade.go
drevell: Done.

templates/model/spec/v1alpha1/upgrade.go
drevell: Done.

templates/model/README.md
drevell: Done.

templates/model/decode/decode.go
drevell: Good idea, implemented.

templates/model/unmarshal.go
drevell: Done.

sethvargo (approved):

PR #230, opened 2023-10-11
Suggest upgrading api_version when it would help

drevell:
I'm trying to help users be less confused when they try to use a feature that's not supported in the api_version that they're using. When a YAML file fails parsing or validation, we'll suggest that the user upgrade to the newest api_version, with one condition. We'll only make this suggestion if the file actually does parse and validate successfully under the newest api_version.

For example, suppose you try to use the new `if` feature in a YAML file whose `api_version` is `v1alpha`. The error message will say (paraphrasing) `your YAML file isn't valid, but if you change the api_version field to v1beta1 instead it will be valid.`

### DIFF
--- a/templates/model/decode/decode.go
+++ b/templates/model/decode/decode.go
@@ -71,10 +71,12 @@ var apiVersions = []apiVersionDef{
 	},
 }
 
-// Decode parses the given YAML contents of r into a struct and returns it. The given filename
-// is used only for error messages. The type of struct to return is determined by the "kind" field
-// in the YAML. If the given requireKind is non-empty, then we'll also validate that the "kind"
-// of the YAML file matches requireKind, and return error if not.
+// Decode parses the given YAML contents of r into a struct and returns it. The
+// given filename is used only for error messages. The type of struct to return
+// is determined by the "kind" field in the YAML. If the given requireKind is
+// non-empty, then we'll also validate that the "kind" of the YAML file matches
+// requireKind, and return error if not. This also calls Validate() on
+// the returned struct and returns error if invalid.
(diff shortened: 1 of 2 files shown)


templates/model/decode/decode.go
```diff
@@ -107,16 +109,28 @@ func Decode(r io.Reader, filename, requireKind string) (model.ValidatorUpgrader,
...
+	// their YAML file doesn't support that new feature. To detect this, we'll
+	// speculatively try to parse the YAML file with a newer api_version and see
+	// if that version would have been valid. If so, we inform the user that
+	// they should change the api_version field in their YAML file.
+	attemptAPIVersion := apiVersions[len(apiVersions)-1].apiVersion
+	if attemptAPIVersion == apiVersion {
+		return nil, "", err // api_version upgrade isn't possible, they're already on the latest.
+	}
+	if _, attemptErr := decodeFromVersionKind(filename, attemptAPIVersion, cf.Kind, buf); attemptErr == nil {
+		return nil, "", fmt.Errorf("file %s sets api_version %q but does not parse and validate successfully under that version. However, it will be valid if you change the api_version to %q. The error was: %w",
```
chloechien, line 129:
```suggestion
		return nil, "", fmt.Errorf("file %s sets api_version %q but does not parse and validate successfully under that version. However, it will be valid if you upgrade the api_version to %q. The error was: %w",
```

chloechien (approved):

sethvargo (approved):

templates/model/decode/decode.go
drevell: Just so I don't lose my readability approval, I'll merge as-is and send another one for this change.

PR #236, opened 2023-10-12
Set a default go version for the protos template

shan4993:


t/proto_template/.github/workflows/ci.yml
```diff
@@ -29,7 +30,7 @@ jobs:
       - name: 'Setup Go'
         uses: 'actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753' # ratchet:actions/setup-go@v4
         with:
-          go-version: '1.21'
+          go-version: ${{ env.DEFAULT_GO_VERSION }}
```
verbanicm, line 33: Wrap with single quotes

drevell (dismissed):

t/proto_template/protos/v1/person_service/person.proto
```diff
@@ -2,7 +2,7 @@ syntax = "proto3";
 
 package person.v1;
 
-option go_package = "github.com/github_org_name/github_repo_name/v1/person;personpb";
+option go_package = "v1/person;personpb";
```
shan4993, line 5: Im removing the prefix here because this causes a redundancy when importing the proto from the client application

verbanicm (approved):

PR #237, opened 2023-10-13
Add upload workflow to proto template

shan4993:


t/proto_template/.github/workflows/upload.yml
```diff
@@ -0,0 +1,70 @@
...
+      contents: 'read'
+      id-token: 'write'
+
+  deploy_go:
+    needs:
+      - 'ci'
+    runs-on: 'ubuntu-latest'
+    permissions:
+      contents: 'read'
+      id-token: 'write'
```
verbanicm, line 35: Seems these are the same permissions between jobs? You can move to workflow level.

t/proto_template/.github/workflows/ci.yml
verbanicm, line 1: nit: this workflow should define permissions too (best practice)

t/proto_template/.github/workflows/upload.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: 'Setup Go'
+        uses: 'actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753' # ratchet:actions/setup-go@v4
+        with:
+          go-version: '${{ env.GO_VERSION }}'
+
+      - name: 'Upload to GAR (go)'
+        env:
+          # version must have format vX.Y.Z https://cloud.google.com/artifact-registry/docs/go/store-go#package_and_upload_your_module
+          VERSION: 'v0.0.0-${{ github.sha }}'
```
verbanicm, line 61: As a follow up maybe, will we allow custom versioning or is there a reason why we dont?

shan4993: i can make the prefix a variable and default to this value. In general this pattern for the version works. This version value was the source of all the pain atleast for go.

verbanicm: I feel like people will want to version their protos, can do in a follow up PR after some testing.

verbanicm (approved):

t/proto_template/.github/workflows/upload.yml
```diff
@@ -0,0 +1,71 @@
...
+          install_components: 'package-go-module'
+
+      - name: 'Setup Go'
+        uses: 'actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753' # ratchet:actions/setup-go@v4
+        with:
+          go-version: '${{ env.GO_VERSION }}'
+
+      - name: 'Upload to GAR (go)'
+        env:
+          # version must have format vX.Y.Z https://cloud.google.com/artifact-registry/docs/go/store-go#package_and_upload_your_module
```
drevell, line 60: The comment `version must have the format vX.Y.Z` seems to contradict the actual version number which has a dash and a sha

drevell (changes requested):

t/proto_template/.github/workflows/upload.yml
drevell: Maybe we should have people avoid v0 version numbers since Go does weird special stuff with them?

shan4993: ack i should specify the comment applies to the prefix. The commit sha suffix is for tracking purposes (as a user i can confirm which artifact has which changes corresponding to the repo). We start off with a basic person proto so ideally when they swap out for their own protos they can upgrade the major version.

drevell (approved):

PR #239, opened 2023-10-18
Add git CLI wrappers for cloning and listing tags

drevell:
This is PR 1/N for supporting semantic-versioned templates. Eventually, we'll replace the go-getter library with our own git CLI wrapper (see internal design doc).

The operations we need to support are:
  - Download a template (git clone)
  - Discover what versions are available (git ls-remote); versions are git tags

### DIFF
--- a/templates/common/git/git.go
+++ b/templates/common/git/git.go
@@ -22,9 +22,13 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/abcxyz/abc/templates/common"
 )
+
+// How long to wait for git operations. This was chosen arbitrarily.
+var gitTimeout = time.Minute
 
 // Clone checks out the given branch or tag from the given repo. It uses the git
 // CLI already installed on the system.
@@ -34,7 +38,9 @@ import (
 // "remote" may be any format accepted by git, such as
 // https://github.com/abcxyz/abc.git or git@github.com:abcxyz/abc.git .
 func Clone(ctx context.Context, remote, branchOrTag, outDir string) error {
-	_, _, err := common.Exec(ctx, "git", "clone", "--depth", "1", "--branch", branchOrTag, remote, outDir)
+	ctx, cancel := context.WithTimeout(ctx, gitTimeout)
+	defer cancel()
+	_, _, err := common.Run(ctx, "git", "clone", "--depth", "1", "--branch", branchOrTag, remote, outDir)
 	if err != nil {
 		return err //nolint:wrapcheck
 	}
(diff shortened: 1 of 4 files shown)


templates/common/git/git.go
```diff
@@ -0,0 +1,98 @@
...
+// https://github.com/abcxyz/abc.git or git@github.com:abcxyz/abc.git .
+func Clone(ctx context.Context, remote, branchOrTag, outDir string) error {
+	cmd := exec.CommandContext(ctx, "git", "clone", "--depth", "1", "--branch", branchOrTag, remote, outDir)
+	cmd.Stderr = &bytes.Buffer{}
+	cmd.Stdout = &bytes.Buffer{}
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("git exec of %v failed: %w\nstdout: %s\nstderr: %s", cmd.Args, err, cmd.Stdout, cmd.Stderr)
+	}
+
+	// Make sure there are no symlinks.
```
sethvargo, line 44: Nit: could this be a helper for easier testing?

templates/common/git/git.go
```diff
@@ -0,0 +1,98 @@
...
+//
+// "remote" may be any format accepted by git, such as
+// https://github.com/abcxyz/abc.git or git@github.com:abcxyz/abc.git .
+func Tags(ctx context.Context, remote string) ([]string, error) {
+	cmd := exec.CommandContext(ctx, "git", "ls-remote", "--tags", remote)
+	stdout, stderr := &bytes.Buffer{}, &bytes.Buffer{}
+	cmd.Stdout, cmd.Stderr = stdout, stderr
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf("git exec of %v failed: %w\nstdout: %s\nstderr: %s", cmd.Args, err, cmd.Stdout, cmd.Stderr)
+	}
```
sethvargo, line 81: Since we're probably going to be doing quite a few shell commands, we might want to extract this into a helper/struct. For example, should we kill this process after a timeout? Should we propagate signals to the child process? How do we prevent zombie processes?

sethvargo (approved):

templates/common/git/git_test.go
```diff
@@ -0,0 +1,230 @@
...
+
+// To actually run the tests in this file, you'll need to set this environment
+// variable.
+//
+// For example:
+//
+//	$ ABC_TEST_NON_HERMETIC=true go test ./...
+const envName = "ABC_TEST_NON_HERMETIC"
+
+func skipUnlessEnvEnabled(t *testing.T) {
```
yolocs, line 36: Is this for manual testing only or there is plan to run this in CI?

yolocs (approved):

templates/common/git/git_test.go
drevell: Currently just for manual testing. No plans to automate yet.

PR #242, opened 2023-10-21
Add parser and downloader for git-hosted templates

drevell:
This is PR 2/N to replace go-getter with our own git downloader. In this PR we:

 - resolve template locations like "github.com/myorg/myrepo@latest"
 - attempt multiple ways of parsing a location, so we can try treating a given template source as either git or local and keep the first one that parses successfully

Anticipated comments:
Q: Why is gitSourceParser so parameterizable and indirect? It's a singleton, it doesn't need to be so flexible.
A: I'm going to add another instance of gitSourceParser in a future PR that will parse go-getter-style URLs like `github.com/abcxyz/abc.git//t/react_template?ref=latest`. So it won't be a singleton anymore.


templates/common/saferelpath_test.go
```diff
@@ -0,0 +1,101 @@
+// Copyright 2023 The Authors (see AUTHORS file)
```
drevell, line 1: This file is just moved code with no changes.

templates/common/saferelpath.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2023 The Authors (see AUTHORS file)
```
drevell, line 1: This file is just moved code with no changes.

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/abcxyz/abc
 go 1.21
 
 require (
+	github.com/Masterminds/semver/v3 v3.2.1
```
sethvargo, line 6: Can you share more about the decision to use this library over https://pkg.go.dev/golang.org/x/mod/semver?

templates/commands/render/templatesource/git.go
```diff
@@ -0,0 +1,218 @@
...
+	// Example: `${subdir}`
+	subdirExpansion string
+	// Example: `${version}`
+	versionExpansion string
+}
+
+func (g *gitSourceParser) sourceParse(ctx context.Context, src, protocol string) (templateDownloader, bool, error) {
+	match := g.re.FindStringSubmatchIndex(src)
+	if match == nil {
+		return nil, false, nil
```
sethvargo, line 59: Do we want to return an error here?

templates/commands/render/templatesource/source.go
```diff
@@ -0,0 +1,99 @@
...
+		if ok {
+			dl = downloader
+			break
+		}
+	}
+	if dl == nil {
+		return nil, fmt.Errorf("template source %q isn't something that we know how to download", source)
+	}
+
+	return dl, nil
```
sethvargo, line 98: Could this be simplified to:

```suggestion
	for _, sp := range srcParsers {
		downloader, ok, err := sp.sourceParse(ctx, source, protocol)
		if err != nil {
			return nil, err //nolint:wrapcheck
		}
		if ok {
		  return downloader
		}
	}
	
	return nil, fmt.Errorf("template source %q isn't something that we know how to download", source)
```

sethvargo (dismissed):

templates/commands/render/templatesource/git.go
```diff
@@ -0,0 +1,218 @@
...
+	return nil
+}
+
+// resolveBranchOrTag returns the latest release tag if branchOrTag is "latest", and otherwise
+// just returns the input branchOrTag after validating it. The return value always begins
+// with "v" (unless there's an error).
+func resolveBranchOrTag(ctx context.Context, t tagser, remote, branchOrTag string) (string, error) {
+	logger := logging.FromContext(ctx).With("logger", "resolveBranchOrTag")
+
+	if branchOrTag != "latest" {
```
yolocs, line 146: What's the behavior of empty `branchOrTag`? From the code, it seems still treated as "latest"? Just confirming it's intended.

templates/commands/render/templatesource/source.go
```diff
@@ -0,0 +1,99 @@
...
+	// This source parser recognizes template sources like
+	// "github.com/myorg/myrepo@v1.2.3" (and variants thereof).
+	&gitSourceParser{
+		re: regexp.MustCompile(
+			`^` + // Anchor the start, must match the entire input
+				`(?P<host>github\.com|gitlab\.com)` + // The domain names of known git hosting services
+				`/` +
+				`(?P<org>[a-zA-Z0-9_-]+)` + // the github org name, e.g. "abcxyz"
+				`/` +
+				`(?P<repo>[a-zA-Z0-9_-]+)` + // the github repo name, e.g. "abc"
```
yolocs, line 59: It's not very clear to me why these sub matching are needed. Not saying we need to but curious about what stops us from supporting something like `https://${other_git_server}@latest`

go.mod
drevell: Good question. I could go either way on this. Here's my reasoning:

 - The Masterminds one is more popular (1200ish imports vs 700ish)
 - The x/mod one has a 0.x version number, which indicates to me that it might not be a stable foundation to build on. The Masterminds one has a 3.x version number.
 - The Masterminds one has an API that seems more sensible to me. It exposes the parsed form of a semver as a struct. The x/mod library which deals with strings and reparses them into a unexported struct every time they're seen (so a sort comparator has to re-parse the semver for every comparison). IMO there's a nice improvement to readability and maintainability to have a SemVer struct that uses the Go type system to guarantee that this semver string is known to be valid. In contrast, the x/mod one just returns empty string when parsing fails (which is the "in-band error" antipattern)

templates/commands/render/templatesource/git.go
drevell: If branchOrTag is empty, then we'll enter this "if" block (because "" != "latest"). We'll then hit the ```fmt.Errorf(`the template source version %q must start with "v", like "v1.2.3"`, branchOrTag)``` a few lines below.

templates/commands/render/templatesource/source.go
drevell: I'll send you the design doc in chat 😀

drevell: Actually, I can just answer here. One of the things we're doing in this series of PRs is creating a namespace for versioned templates that is separate from the transport details (https or ssh). This lets us do things like "template X depends on template Y", and we can refer to template Y without being overly specific about the details of the transport protocol that we use to fetch template Y. It's similar to `go get` in that way.

templates/commands/render/templatesource/git.go
yolocs: But should we treat empty as "latest"?

drevell: Perhaps so... @sethvargo do you have an opinion on whether `@latest` should be implicit if no version is given? It seems to work fine for `go get`.

templates/commands/render/templatesource/source.go
yolocs: Understood. But what I was asking about was the host/org/repo part. Why do we need to know about them? From your description, it doesn't sound like we need them for the template dependency?

drevell: It's just to verify that the structure is what we expect. If somebody gives a template source of `github.com/foo` (org name with no repo name), this lets us detect that it's not a valid remote git repo, because it doesn't match the regex. We can continue with other sourceParsers and try treating it as a local file or $FUTURE_TEMPLATE_SOURCE_TYPE.

We could create one big capturing group for "host plus org plus repo but not including subdirectory or version", but that might be harder to understand because that's a confusing union of concepts for the reader to grok.

go.mod
sethvargo: Those all sound like really compelling reasons! I'm struggling to think of a good place to capture this decision somewhere in the code.

templates/commands/render/templatesource/git.go
sethvargo: The upgrade behavior for Go modules is documented by confusing. I think getting latest makes sense.

sethvargo (approved):

templates/commands/render/templatesource/source.go
yolocs: I see. I mainly question whether we really need to care about the format. Would it still work if it's not github/gitlab format?

drevell: If it's not github/gitlab, then it won't work, but that's by design. In that case, we'll continue with the next sourceParser and attempt to treat the input as a local file rather than a git repo.

go.mod
drevell: Yeah... I couldn't think of a good way to document it that would actually be helpful to anyone.

PR #243, opened 2023-10-23
Untangle tech debt re file separators in tests

drevell:
There are a bunch of tests that deal with converting file path separators, but this can all be cleaned way up by moving it down into the helper functions that are already being used.

Also, make `WriteAll*` call `t.Fatalf` instead of returning error, to be consistent with `LoadAll*` and save boilerplate in tests.


sethvargo (approved):
So long fielepath.Join. You served us well 🖖 .

PR #245, opened 2023-10-24
Add new flag arg for input file config

stytchiz:


templates/commands/render/render.go
```diff
@@ -586,6 +600,24 @@ func executeOneStep(ctx context.Context, stepIdx int, step *spec.Step, sp *stepP
...
 
+func loadInputFile(ctx context.Context, fs common.FS, path string) (map[string]string, error) {
+	f, err := fs.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("error opening input file: ReadFile(): %w", err)
+	}
+	defer f.Close()
+
+	m := make(map[string]string)
+	data, err := io.ReadAll(f)
```
drevell, line 611: I think this could be simplified a bit using https://pkg.go.dev/os#ReadFile

templates/commands/render/flags.go
```diff
@@ -91,6 +94,14 @@ func (r *RenderFlags) Register(set *cli.FlagSet) {
 		Usage:   "The key=val pairs of template values; may be repeated.",
 	})
 
+	f.StringVar(&cli.StringVar{
+		Name:    "input-file",
+		Example: "/my/git/abc-inputs.yaml",
+		Default: "",
+		Target:  &r.InputFile,
+		Usage:   "The file with the key=val pairs of template values.",
```
drevell, line 102: It might be misleading to say "key=val" pairs since it's a YAML file

templates/commands/render/render.go
```diff
@@ -586,6 +600,24 @@ func executeOneStep(ctx context.Context, stepIdx int, step *spec.Step, sp *stepP
 	}
 }
 
+func loadInputFile(ctx context.Context, fs common.FS, path string) (map[string]string, error) {
```
drevell, line 603: Should there be a test for this?

drevell (changes requested):

templates/commands/render/flags.go
stytchiz: Done

templates/commands/render/render.go
stytchiz: Tests added.

templates/commands/render/render.go
stytchiz: Done

templates/commands/render/flags.go
```diff
@@ -91,6 +94,14 @@ func (r *RenderFlags) Register(set *cli.FlagSet) {
 		Usage:   "The key=val pairs of template values; may be repeated.",
 	})
 
+	f.StringVar(&cli.StringVar{
```
drevell, line 97: This can wait for a future PR, but I wonder if we should accept multiple --input-file arguments, and they would be merged?

drevell (approved):

templates/commands/render/flags.go
stytchiz: No strong opinion. The customer workaround would be to concatenate the input files into a single tmp file (i.e. `cat file1.yaml file2.yaml > /tmp/inputs.yaml`). However if it turns out to be a common use case, I'm willing to go ahead and support it.

PR #247, opened 2023-10-24
Add a sourceParser that reads a template from a local dir

drevell:
This is part of the PR series that adds semantic versioning of templates. As part of that, we're replacing the go-getter library with our own template downloader.

### DIFF
--- a/templates/commands/render/templatesource/localsource.go
+++ b/templates/commands/render/templatesource/localsource.go
@@ -1,0 +1,75 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


templates/commands/render/templatesource/localsource.go
```diff
@@ -0,0 +1,75 @@
...
+	// false negatives (e.g. you have a directory named "github.com/..."). Instead,
+	// we'll just check if the given path actually exists, and if so, then treat
+	// src as a local directory name.
+	//
+	// This sourceParser should run after the sourceParser that recognizes remote
+	// git repos, so this code won't run if the source looks like a git repo.
+
+	_, err := os.Stat(src)
+	if err != nil {
+		_, isPathError := err.(*fs.PathError) //nolint:errorlint // we'd like to use errors.Is(), but it doesn't work with *fs.PathError
```
sethvargo, line 47: Why doesn't it work with PathError?

drevell: Just based on trying it. When I tried to understand why, I think I figured it out.

From the docs for errors.Is: `An error is considered to match a target if it is equal to that target or if it implements a method Is(error) bool such that Is(target) returns true.` So there are two cases that could make it work, neither of which is satisfied here.

1. Plain equality. [PathError](https://pkg.go.dev/io/fs#PathError) contains some fields like `Path` and `Op`, so plain equality doesn't work when comparing against a zero `PathError{}`.
2. PathError doesn't have an `Is()` method.

drevell: Added more info to the code comment to explain this to future maintainers.

sethvargo (approved):

PR #252, opened 2023-10-25
Final replacement of go-getter with our own templatesource

drevell:
This is the final PR of the series that replaces go-getter with our own template downloading and copying logic.


sethvargo (dismissed):

templates/commands/render/render_test.go
```diff
@@ -719,26 +708,33 @@ steps:
...
 	}
-	return out
+	switch len(matches) {
+	case 0:
+		return "", false
+	case 1:
+		return matches[0], true
+	}
+	t.Fatalf("got %d matches for glob %q, wanted 1: %s", len(matches), glob, matches)
+	panic("unreachable")
```
sethvargo, line 737: How would we get here?

sethvargo (approved):

templates/commands/render/render_test.go
drevell: Some background here: this is used to find one directory within a temp directory contaning several directories, while ignoring a random cookie at the end of the name. For example, suppose there's a temp directory containing two subdirectories, `template-copy-192837` and `scratch-977245`. If you want to find the template directory, you pass a glob of `template-copy-*` and you'd find `template-copy-192837`.

So the Fatalf covers the case where the caller's glob wasn't specific enough to match a single directory. This could happen if the glob was just `*` or `*-*`.

sethvargo: Sorry, I'm not questioning the fatalf. I'm questioning the panic after it. Fatal should stop the test immediately - I'm not clear how we'd ever hit L737 - it's unreachable code.

yolocs: Same question

templates/commands/render/render_test.go
```diff
@@ -719,26 +708,33 @@ steps:
...
-// Since os.MkdirTemp adds an extra random token, we strip it back out to get
-// determistic results. Input map keys should use forward slash separator.
-func stripFirstPathElem(m map[string]string) map[string]string {
-	out := map[string]string{}
-	for k, v := range m {
-		// Panic in the case where k has no slashes; this is just a test helper.
-		elems := strings.Split(k, "/")
-		newKey := path.Join(elems[1:]...)
-		out[newKey] = v
+func mustGlob(t *testing.T, glob string) (string, bool) {
```
yolocs, line 723: https://github.com/abcxyz/readability/wiki/Go-Best-Practices#test-helper We often prefix test helpers with "test" prefix, e.g. `testMustGlob`, for the purpose of identifying test helpers more easily

templates/commands/render/render_test.go
drevell: Ah, I see. The compiler complains about a missing return statement because the compiler doesn't know that t.Fatalf() never returns. The panic is the standard way of handling that.

drevell: Added a comment.

templates/commands/render/render_test.go
drevell: Done.

templates/commands/render/render_test.go
```diff
@@ -748,26 +737,33 @@ steps:
...
-// Since os.MkdirTemp adds an extra random token, we strip it back out to get
-// determistic results. Input map keys should use forward slash separator.
-func stripFirstPathElem(m map[string]string) map[string]string {
-	out := map[string]string{}
-	for k, v := range m {
-		// Panic in the case where k has no slashes; this is just a test helper.
-		elems := strings.Split(k, "/")
-		newKey := path.Join(elems[1:]...)
-		out[newKey] = v
+func testMustGlob(t *testing.T, glob string) (string, bool) {
```
sethvargo, line 752: Prefer `testing.TB` interface instead (sorry I missed this in the first pass)

sethvargo (approved):

yolocs (approved):

templates/commands/render/render_test.go
drevell: Will follow up in a separate PR, this one is blocking many downstream PRs so I'm eager to get it merged.

PR #263, opened 2023-11-03
Add Flutter template

chloechien:
A boilerplate generated from Flutter CLI + clean UX + a huge amount of linters


mattrandallbecker (approved):

ruizhg (approved):

bradegler (commented):
Do we have the ability to render a template over the top of an existing directory? It would be nice to apply just the value added stuff and let the normal generator do the rest of the work since that might change frequently and we'd have to keep this template up to date.

It would be better if the workflow for this template was more like:

```
flutter create -t app --project-name myapp --platforms ios,android --org dev.abcxyz --description "my cool app" .;
abc template render <template path.../t/flutter_template> myapp;
```

What are your thoughts?

chloechien:
I'd say that's brilliant idea! The only concern is we don't have control on flutter CLI, any CLI version updates might (even by very low chance) break current template. If we lock down the CLI version then we still don't get new template updates. We can discuss this further. At present it might be simple to have this template as-it, and improve it in future.

PR #266, opened 2023-11-06
Opt-out TestParseSource from parallel testing

drevell:
Previously we opted out the subtests individually from parallel testing, but on further thought we actually need to opt-out the entire test. Since we call os.Chdir() in this test, we need to guarantee that we're not running at the same time as another test which might possibly also call os.Chdir().


pdewilde (approved):

sethvargo:
Is there a way we can feed the tempdir in for testing instead of relying on the global state of os.GetCwd?

yolocs (dismissed):

pdewilde (approved):
I agree it would be better if these tests could be isolated, though I don't want the perfect to be the enemy of the good.

drevell:
> Is there a way we can feed the tempdir in for testing instead of relying on the global state of os.GetCwd?

Yeah, we could, but I think it's not worth it. Here's why:
1. What we're trying to test is that templates can be loaded from the CWD. We could rewrite this code to remove the dependency on the CWD and just pass in the directory, but that would just move the logic elsewhere.
2. It would add a bunch of boilerplate parameter passing to no real benefit.

yolocs (approved):

sethvargo:
> > Is there a way we can feed the tempdir in for testing instead of relying on the global state of os.GetCwd?
> 
> Yeah, we could, but I think it's not worth it. Here's why:
> 
> 1. What we're trying to test is that templates can be loaded from the CWD. We could rewrite this code to remove the dependency on the CWD and just pass in the directory, but that would just move the logic elsewhere.

Sorry but I disagree. Taking a dependency on global state like cwd is a design anti-pattern.

> 2. It would add a bunch of boilerplate parameter passing to no real benefit.

It's one parameter, and the benefit is that you are not depending on global state. https://github.com/abcxyz/abc/pull/279 is how I would accomplish this. You can avoid leaking the extra param in the public API if you make `ParseSourceWithWorkingDir` private.

drevell:
> Sorry but I disagree. Taking a dependency on global state like cwd is a design anti-pattern.

SGTM, will continue the discussion in #279.

PR #269, opened 2023-11-07
Polish Flutter template and add User Avatar.

chloechien:


t/flutter_template/pubspec.yaml
```diff
@@ -62,10 +62,9 @@ flutter:
   # the material Icons class.
   uses-material-design: true
 
-  # To add assets to your application, add an assets section, like this:
-  # assets:
-  #   - images/a_dot_burr.jpeg
-  #   - images/a_dot_ham.jpeg
+  # To add assets to your application, add in assets section below.
+  assets:
+    - assets/bp_logo.png
```
bradegler, line 67: We should quote all strings in this doc since it is yaml and bad things can happen if you don't quote them. I know that `flutter create` doesn't quote anything but if this is our template we should set a better example.

bradegler (approved):

t/flutter_template/pubspec.yaml
chloechien: Good question! I've set this file `# yamllint disable-file` main concern is, it's been maintained by Flutter CLI. e.g. install a new dependencies. That meaning if we format the file, whenever a user runs `flutter pub add`, an extra manual step is needed. However in another file, `[analysis_options.yaml](https://github.com/abcxyz/abc/blob/main/t/flutter_template/analysis_options.yaml)`, b/c we have full control over that, we quote all the dependencies which is different from `flutter create`

bradegler: Yeah, that makes sense!

bradegler (approved):

PR #272, opened 2023-11-07
Cleanups to input handling

drevell:
1. Replace mutable state with immutable state: c.flags.input is the map containing the --input flag values. We were mutating it in-place with the contents of the --input-file files, which was somewhat unclean.
2. The above change lets us change some methods into stateless functions.
3. Previously we didn't define what would happen when multiple --input-files try to define the same input. We now define this to be an error, and document that fact.
4. Use some library functions for union and subtraction rather than writing our own for loops.


README.md
```diff
@@ -56,9 +56,15 @@ The `<template_location>` parameter is one of these two things:
...
-  must have a string `val` (i.e. no nesting and no arrays). If `key` exists in the 
-  file but is also provided as an input parameter, the input value takes precedence.
+- `--input-file=file`: provide a YAML file with input(s) to the template. The
+  file must contain a YAML object whose keys and values are strings. If a key
+  exists in the file but is also provided as an `--input`, the `--input` value
+  takes precedence.
+
+  This flag may be repeated, like
+  `--input-file=some-inputs.yaml --input-file=more-inputs.yaml`. When there are
+  multiple input files, they must not have any overlapping keys.
```
stytchiz, line 66: What's the reason for not supporting overlapping keys vs. allowing later files to override keys? Seems easier to implement the latter (unless there's a design principle I'm missing here).

e.g. `--input-file=defaults.yaml --input-file=overrides.yaml`

stytchiz (dismissed):

README.md
drevell: If this happens, I think it's likely to be a mistake and I want to help the user not do something they don't intend. I don't think the "defaults plus overrides" pattern will occur much in real use cases (do you have any concrete use cases in mind?)

stytchiz: No strong opinion. Like I mentioned, it's a design decision that I wanted to understand.

PR #279, opened 2023-11-09
Allow parallelizing ParseSource tests

sethvargo:
This removes the global dependency on $PWD.

- Refs https://github.com/abcxyz/abc/pull/266

### DIFF
--- a/templates/commands/render/templatesource/git.go
+++ b/templates/commands/render/templatesource/git.go
@@ -30,6 +30,8 @@ import (
 	"golang.org/x/exp/slices"
 )
 
+var _ sourceParser = (*gitSourceParser)(nil)
+
 // gitSourceParser implements sourceParser for downloading templates from a
 // remote git repo.
 type gitSourceParser struct {
@@ -60,7 +62,7 @@ type gitSourceParser struct {
 	warning string
 }
 
-func (g *gitSourceParser) sourceParse(ctx context.Context, src, protocol string) (Downloader, bool, error) {
+func (g *gitSourceParser) sourceParse(ctx context.Context, cwd, src, protocol string) (Downloader, bool, error) {
 	logger := logging.FromContext(ctx).With("logger", "gitSourceParser.sourceParse")
 
 	match := g.re.FindStringSubmatchIndex(src)
--- a/templates/commands/render/templatesource/localsource.go
+++ b/templates/commands/render/templatesource/localsource.go
@@ -21,16 +21,19 @@ import (
 	"fmt"
 	"io/fs"
 	"os"
+	"path/filepath"
 
 	"github.com/abcxyz/abc/templates/common"
 	"github.com/abcxyz/pkg/logging"
 )
 
+var _ sourceParser = (*localSourceParser)(nil)
+
 // localSourceParser implements sourceParser for reading a template from a local
 // directory.
 type localSourceParser struct{}
 
-func (l *localSourceParser) sourceParse(ctx context.Context, src, protocol string) (Downloader, bool, error) {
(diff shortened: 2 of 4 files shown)


templates/commands/render/templatesource/source.go
```diff
@@ -86,17 +87,17 @@ var realSourceParsers = []sourceParser{
 	},
 }
 
-// ParseSource maps the input template source to a particular kind of source
-// (e.g. git) and returns a downloader that will download that source.
+// ParseSourceWithWorkingDir maps the input template source to a particular kind
```
drevell, line 90: Can this be unexported (lower case)?

Maybe also consider "CWD" or "WD" (in the vein of os.Getwd()) instead of "WorkingDir", it's quite a long function name.

templates/commands/render/templatesource/localsource.go
```diff
@@ -42,18 +45,26 @@ func (l *localSourceParser) sourceParse(ctx context.Context, src, protocol strin
...
+			"originalSrc", originalSrc,
+			"src", src)
+	}
+
+	if _, err := os.Stat(src); err != nil {
 		var pathErrPtr *fs.PathError
 
 		if errors.Is(err, fs.ErrNotExist) || errors.Is(err, fs.ErrInvalid) || errors.As(err, &pathErrPtr) {
-			logger.DebugContext(ctx, "won't treat src as a local path because that path doesn't exist", "src", src)
+			logger.DebugContext(ctx, "ignoring src as a local path because the path does not exist", "src", src)
```
drevell, line 61: The word "ignoring" seems likely to be misinterpreted by the user. This particular function is indeed ignoring it, but other code may not ignore it. It's probably asking too much of the user to understand that distinction.

drevell (changes requested):

templates/commands/render/templatesource/git.go
```diff
@@ -60,7 +62,7 @@ type gitSourceParser struct {
 	warning string
 }
 
-func (g *gitSourceParser) sourceParse(ctx context.Context, src, protocol string) (Downloader, bool, error) {
+func (g *gitSourceParser) sourceParse(ctx context.Context, _cwd, src, protocol string) (Downloader, bool, error) {
```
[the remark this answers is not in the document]
sethvargo: Good catch. I was just writing in the wrong language 😄. Underscore variables in Rust are a thing.

templates/commands/render/templatesource/source_test.go
```diff
@@ -16,19 +16,17 @@ package templatesource
...
 	"github.com/abcxyz/abc/templates/common"
 	"github.com/abcxyz/pkg/testutil"
 	"github.com/google/go-cmp/cmp"
 )
 
-func TestParseSource(t *testing.T) { //nolint:paralleltest
-	// We can't use t.Parallel() here because we use os.Chdir. We don't want
-	// multiple goroutines messing with the working directory at the same time.
-	// The test is fast enough that the peformance impact is negligible.
+func TestParseSource(t *testing.T) {
```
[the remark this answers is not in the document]
sethvargo: Don.

templates/commands/render/templatesource/source.go
sethvargo: `parseSourceWithCwd` it is!

templates/commands/render/templatesource/localsource.go
sethvargo: Reverted to "will not" (contractions are bad for non-native speakers).

drevell (approved):
Thanks for contributing this

PR #281, opened 2023-11-10
Add SHA256 hashing in CopyRecursive

drevell:
This isn't used yet, but will soon be used for saving a hash of each template output file in a manifest. This prepares for the upcoming template upgrade feature.


pdewilde (dismissed):

templates/common/fs_test.go
```diff
@@ -393,6 +461,20 @@ func TestCopyRecursive(t *testing.T) {
...
 			}
+
+			if diff := cmp.Diff(hashes, tc.wantHashes); diff != "" {
+				t.Errorf("hashes were not as expected: (-got,+want): %s", diff)
+			}
 		})
 	}
 }
+
+func mustHexDecode(t *testing.T, hexStr string) []byte {
```
sethvargo, line 472: Prefer tb

templates/common/fs_test.go
```diff
@@ -114,6 +120,19 @@ func TestCopyRecursive(t *testing.T) {
...
 		},
+		{
+			name: "dry_run_should_calculate_hashes",
+			srcDirContents: map[string]ModeAndContents{
+				"file1.txt": {Mode: 0o600, Contents: "file1 contents"},
+			},
+			hash:   sha256.New(),
+			dryRun: true,
+			wantHashes: map[string][]byte{
+				"file1.txt": mustHexDecode(t, "226e7cfa701fb8ba542d42e0f8bd3090cbbcc9f54d834f361c0ab8c3f4846b72"),
```
sethvargo, line 131: Nit: it might be cleaner if `wantHashes` is `map[string]string` and you perform the hex decoding in the test function before comparison.

templates/common/fs.go
```diff
@@ -109,6 +110,18 @@ type CopyParams struct {
...
+	// If Hash and OutHashes are not nil, then each copied file will be hashed
+	// and the raw binary hash will be saved in OutHashes. If a file is
+	// "skipped" (CopyHint.Skip==true) then the hash will not be computed. In
+	// dry run mode, the hash will be computed normally. OutHashes always uses
+	// forward slashes as path separator, regardless of OS.
+	//
+	// Note to maintainers: a hash.Hash is stateful and therefore not
+	// thread-safe. If we ever add concurrency inside CopyRecursive, we'll need
+	// to change this.
+	Hash      hash.Hash
```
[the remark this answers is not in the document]
drevell: Yeah, I was on the fence about that. Done.

templates/common/fs_test.go
drevell: I was on the fence about that too... given your vote, I'll go with hex. Done.

templates/common/fs_test.go
drevell: Function is now gone with the switch to returning hex from copyRecursive

templates/common/fs_test.go
drevell: I ended up just returning the hashes as hex.

templates/common/fs.go
```diff
@@ -109,6 +111,14 @@ type CopyParams struct {
 	// visitor is an optional function that will be called for each file in the
 	// source, to allow customization of the copy operation on a per-file basis.
 	Visitor CopyVisitor
+
+	// If Hasher and OutHashes are not nil, then each copied file will be hashed
```
sethvargo, line 115: What are the situations in which we expect them to be nil?

sethvargo (approved):

templates/common/fs.go
drevell: We use copyRecursive in a few places as a general directory copying tool: (1) copying files from the template dir to the scratch dir in an include action (2) copying the temp dir to the output dir in goldentest recording (3) committing the scratch dir to the destination dir. We only want hashing for case 3.

PR #283, opened 2023-11-13
Implement "canonical template sources"

drevell:
A canonical template source is one that isn't tied to the machine where it's running. For example, a template source like
`github/abcxyz/abc/t/rest_server` is canonical because anyone can reference it. A template source like "/home/dave/my_unpacked_tarball" is not canonical because that directory only exists on my machine. As a complicated special case, a path like `/home/dave/git/myrepo` may or may not be canonical, depending on whether the destination is in the same git repo as the source. See the internal design doc (ping me for a link).

Also, we add a new common function `IsStatNotExist()` to factor out the common pattern of checking whether we Stat()ed a path that doesn't exist.

### DIFF
--- a/templates/commands/render/render.go
+++ b/templates/commands/render/render.go
@@ -152,7 +152,7 @@ func (c *Command) realRun(ctx context.Context, rp *runParams) (outErr error) {
 		outErr = errors.Join(outErr, err)
 	}()
 
-	_, templateDir, err := c.copyTemplate(ctx, rp)
+	_, templateDir, err := c.downloadTemplate(ctx, rp)
 	if templateDir != "" { // templateDir might be set even if there's an error
 		tempDirs = append(tempDirs, templateDir)
 	}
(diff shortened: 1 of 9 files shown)


templates/commands/render/render.go
```diff
@@ -638,28 +638,31 @@ func loadSpecFile(ctx context.Context, fs common.FS, templateDir string) (*spec.
 // Downloads the template and returns the name of the temp directory where it
```
capri-xiyue, line 638: Nits: update the comment?

capri-xiyue (approved):

templates/commands/render/render.go
drevell: Fixed.

sethvargo:
@drevell I'm probably missing something obvious, but why isn't a canonical source a function of the downloader?

drevell:
> @drevell I'm probably missing something obvious, but why isn't a canonical source a function of the downloader?

Perhaps that would make more sense. As it is, it seems like it belongs equally well in ParsedSource or Downloader. But I suppose if it were in the Downloader, we could do smarter things like notice redirects. Did you have some rationale in mind?

sethvargo:
> Perhaps that would make more sense. As it is, it seems like it belongs equally well in ParsedSource or Downloader. But I suppose if it were in the Downloader, we could do smarter things like notice redirects. Did you have some rationale in mind?

Redirects were the exact thing I had in mind. I also thought, since the download is already an interface, we could easily check if it's "canonical" by having a private `isCanonical` function on a different interface.

drevell:
> Redirects were the exact thing I had in mind. I also thought, since the download is already an interface, we could easily check if it's "canonical" by having a private `isCanonical` function on a different interface.

Sounds good, done.

capri-xiyue (approved):

PR #285, opened 2023-11-15
Ignore unknown inputs in input file

stytchiz:


templates/commands/render/render.go
```diff
@@ -295,10 +299,6 @@ func (c *Command) resolveInputs(ctx context.Context, fs common.FS, spec *spec.Sp
 	// Order matters: values from --input take precedence over --input-file.
 	inputs := sets.UnionMapKeys(c.flags.Inputs, fileInputs)
```
drevell, line 300: I think we want to go a step further with this. If there's an input that's set in a file, then that input should not be exposed to the rest of the template logic. So only the inputs declared in the spec.Spec should be present in the `inputs` map.

And perhaps once we implement that, there should be a unit test like "an input that's not declared in the "inputs" section of the spec fail should not be able to be referenced in template expressions." So in a spec file `{{.my_ignored_input}}` should be an error even if the --input-file set `my_ignored_input`.

drevell (changes requested):

templates/commands/render/render.go
stytchiz: Great catch! Fixed by taking the intersection of map keys between file inputs and spec inputs. PTAL.

drevell (approved):

templates/commands/render/render.go
drevell: LGTM, thanks.

PR #290, opened 2023-11-18
chore: allow versions with branch names or SHA

### LINKED ISSUE #288: Allow templates to be installed from a named branch or tag, not just semver tags
### TL;DR

In the 0.3 release, a new requirement was added that templates installed from a remote git repo must come from a tagged semver release like `github.com/foo/bar@v1.2.3`. The goal was to avoid users accidentally installing unreleased templates. This was probably overzealous. We can also allow templates to be installed from a specific named branch or tag, to avoid forcing template owners to set up tags for releases.

"As a template user, I want to be able to specify @main (or @mybranch in general) to install the template from that repo."

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

gjonathanhong:
abc CLI `v0.3.0` introduced a constraint to only allowing versions with `latest` or `semver` tag. We no longer want to force the tagged semver constraint on users so this change will allow versioning with branch names, including `@main`, or long commit SHA's. Short commit SHA's are not supported and will be tried as branch name or tag.

Resolves #288

### DIFF
--- a/templates/commands/render/templatesource/git.go
+++ b/templates/commands/render/templatesource/git.go
@@ -203,6 +203,8 @@ func resolveVersion(ctx context.Context, t tagser, remote, version string) (stri
 	}
 }
 
+// resolveLatest retrieves the tags from the remote repository and returns the
+// highest semver tag. An error is thrown if no semver tags are found.
 func resolveLatest(ctx context.Context, t tagser, remote, version string) (string, error) {
 	logger := logging.FromContext(ctx).With("logger", "resolveVersion")
 
--- a/templates/commands/render/templatesource/git.go
+++ b/templates/commands/render/templatesource/git.go
@@ -30,10 +30,7 @@ import (
 	"github.com/abcxyz/pkg/logging"
 )
 
-var (
-	_  sourceParser = (*gitSourceParser)(nil)
-	re              = regexp.MustCompile("^[a-zA-Z0-9/_-]+$")
-)
+var _ sourceParser = (*gitSourceParser)(nil)
 
 // gitSourceParser implements sourceParser for downloading templates from a
 // remote git repo.
@@ -190,8 +187,8 @@ func (g *gitDownloader) CanonicalSource(context.Context, string, string) (string
 }
 
 // resolveVersion returns the latest release tag if version is "latest", and otherwise
-// just returns the input version after validating it. The return value always begins
-// with "v" (unless there's an error).
+// just returns the input version. The return value is either a branch, tag, or
+// a long commit SHA (unless there's an error).
 func resolveVersion(ctx context.Context, t tagser, remote, version string) (string, error) {
 	logger := logging.FromContext(ctx).With("logger", "resolveVersion")
 
--- a/templates/commands/render/templatesource/git.go
+++ b/templates/commands/render/templatesource/git.go
@@ -192,20 +192,20 @@ func (g *gitDownloader) CanonicalSource(context.Context, string, string) (string
 func resolveVersion(ctx context.Context, t tagser, remote, version string) (string, error) {
 	logger := logging.FromContext(ctx).With("logger", "resolveVersion")
 
-	if version != "latest" {
-		if !strings.HasPrefix(version, "v") {
-			return "", fmt.Errorf(`the template source version %q must start with "v", like "v1.2.3"`, version)
-		}
-		ersion := version[1:] // trim off "v" prefix
-		if _, err := semver.StrictNewVersion(ersion); err != nil {
-			return "", fmt.Errorf("the template source requested git tag %q, which is not a valid format for a semver.org version", version)
-		}
+	switch version {
+	case "":
+		return "", fmt.Errorf("the template source version cannot be empty")
+	case "latest":
+		return resolveLatest(ctx, t, remote, version)
+	default:
 		logger.DebugContext(ctx, `using user provided version, no need to look up remote tags`, "version", version)
 		return version, nil
 	}
+}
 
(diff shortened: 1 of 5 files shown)


templates/commands/render/templatesource/git.go
```diff
@@ -186,21 +186,33 @@ func (g *gitDownloader) CanonicalSource(context.Context, string, string) (string
...
+				logger.DebugContext(ctx, `using user provided branchOrTag, no need to look up remote tags`, "branchOrTag", branchOrTag)
+				return branchOrTag, nil
+			}
 		}
-		version := branchOrTag[1:] // trim off "v" prefix
-		if _, err := semver.StrictNewVersion(version); err != nil {
-			return "", fmt.Errorf("the template source requested git tag %q, which is not a valid format for a semver.org version", branchOrTag)
+
+		// Exclude "." to avoid matching an invalid semver, e.g. "v0..12" and "v1.01.2"
+		re := regexp.MustCompile("^[a-zA-Z0-9/_-]+$")
```
yolocs, line 208: For regex, we usually define it on the package level so we don't panic during runtime.

templates/commands/render/templatesource/git.go
```diff
@@ -186,21 +186,33 @@ func (g *gitDownloader) CanonicalSource(context.Context, string, string) (string
...
+		matches := re.MatchString(branchOrTag)
+		if matches {
```
yolocs, line 210: nit
```suggestion
		if re.MatchString(branchOrTag) {
```

templates/commands/render/templatesource/git.go
```diff
@@ -186,21 +186,33 @@ func (g *gitDownloader) CanonicalSource(context.Context, string, string) (string
...
+			logger.DebugContext(ctx, `using user provided branchOrTag, no need to look up remote tags`, "branchOrTag", branchOrTag)
```
yolocs, line 211:
```suggestion
			logger.DebugContext(ctx, "using user provided branchOrTag and skipping remote tags lookup", "branchOrTag", branchOrTag)
```

templates/commands/render/templatesource/git.go
```diff
@@ -186,21 +186,33 @@ func (g *gitDownloader) CanonicalSource(context.Context, string, string) (string
...
+			return branchOrTag, nil
 		}
-		logger.DebugContext(ctx, `using user provided branchOrTag, no need to look up remote tags`, "branchOrTag", branchOrTag)
-		return branchOrTag, nil
+
+		return "", fmt.Errorf("the template source version %q is not a valid format. must be one of semver, branch name, or SHA", branchOrTag)
```
yolocs, line 215:
```suggestion
		return "", fmt.Errorf("invalid template source version %q; must be one of semver, branch name, or SHA", branchOrTag)
```

templates/commands/render/templatesource/git.go
drevell, line 1: I think more changes may be needed for this feature to be complete. Even though resolveBranchOrTag has been changed to return a SHA, that SHA would then be passed to Clone() (https://github.com/abcxyz/abc/blob/main/templates/common/git/git.go#L36) which according to its docs requires a branch or tag name, not a SHA. So I suspect there needs to be a change in templates/common/git/git.go to allow cloning a SHA. There would be some minor design decisions to be made here: would we add another function like CloneSHA()? Or combine the new functionality into Clone()?

You can check if I'm mistaken. You can try running this PR from the command line and seeing whether it works with a SHA. An example command you could try would be `go run cmd/abc/abc.go templates render github.com/abcxyz/abc/t/rest_server@826e4582554f156665191fae040da2e79a64f3a5`. If this command already works, then I'm mistaken. If it fails, you'll know where to start fixing.

drevell (changes requested):

templates/commands/render/templatesource/git.go
sailorlqh: I think what Chen means here is declaring the variable at the package level outside of the function. Somelike
```
import xxxxx

//under the import, add a var section
var (
   _ sourceParser = (*gitSourceParser)(nil)
  re = regexp.MustCompile("^[a-zA-Z0-9/_-]+$")
)
```

sailorlqh: This way, we will be able to caught if the string is valid

templates/commands/render/templatesource/git.go
```diff
@@ -187,25 +187,25 @@ func (g *gitDownloader) CanonicalSource(context.Context, string, string) (string
...
+		return "", fmt.Errorf("the template source version cannot be empty")
+	case "latest":
+		return resolveLatest(ctx, t, remote, version)
+	default:
+		logger.DebugContext(ctx, "using user provided version and skipping remote tags lookup", "version", version)
 		return version, nil
 	}
+}
+
+func resolveLatest(ctx context.Context, t tagser, remote, version string) (string, error) {
```
sailorlqh, line 206: nit: add function doc.

templates/common/git/git_test.go
```diff
@@ -63,55 +63,54 @@ func TestClone(t *testing.T) {
...
 
 	cases := []struct {
-		name        string
-		remote      string
-		branchOrTag string
-		wantErr     string
+		name    string
+		remote  string
+		version string
+		wantErr string
```
sailorlqh, line 69: I think it's worth to add a test case when the sha is not valid, this is to add coverage for the regexp. Please correct me if it's tested elsewhere.

sailorlqh (approved):

drevell (approved):

Issue #62, opened 2023-06-26
Implement `templates describe <url>`


drevell:
### TL;DR

We want users to be able to find out information about a template, like:

 - The description
 - The set of inputs: which are required, what are their descriptions, what are their names

One possible design: we add a new subcommand `templates describe` that downloads a template, parses it's spec file, and prints out the metadata about the template in a human-friendly format.

Another idea: formalize the convention that "README.md must exist and contain usage instructions for the template, expected to be a few hundred words long at least" 

CC @verbanicm who mentioned this.

drevell:
Edited description to add this comment: "Another idea: formalize the convention that "README.md must exist and contain usage instructions for the template, expected to be a few hundred words long at least""

drevell:
Now that we support git tags as template versions, the `describe` command should also print a list of versions (or maybe just the latest version). This can use the `Tags()` function in git.go along with some parsing code like https://github.com/abcxyz/abc/blob/9314a9adf480cf8014d70ced3f4b4caa56de3133/templates/commands/render/templatesource/git.go#L180-L197 .

drevell:
I think we should consider adding docs for this before calling it complete

sailorlqh:
> I think we should consider adding docs for this before calling it complete

Oh yeah, I will update docs.

PR #293, opened 2023-11-29
refactor: Move downloadTemplate from templates/command/render.go to templates/common/templatesource.

sailorlqh:
Context: to fix issue https://github.com/abcxyz/abc/issues/62, we need to download the template first, so we need to move the [downloadTemplate function](https://github.com/abcxyz/abc/blob/main/templates/commands/render/render.go#L655)  from `templates/command/render.go` to somewhere under `templates/common`.

But if we only move this function, and make a `download.go` under package common, this function returns `templatesource.Downloader` makes common depends on `templateSource`, and `templateSource` already depends on `common`, therefore a circular dependency will occur. 

Therefore, I moved the whole `templatesource` folder from `templates/commands/render` to `templates/common/templatesource`.


templates/commands/render/render.go
```diff
@@ -153,12 +152,12 @@ func (c *Command) realRun(ctx context.Context, rp *runParams) (outErr error) {
 		outErr = errors.Join(outErr, err)
 	}()
 
-	_, templateDir, err := c.downloadTemplate(ctx, rp)
+	_, templateDir, err := templatesource.DownloadTemplate(ctx, rp.fs, rp.tempDirBase, c.flags.Source, c.flags.GitProtocol)
```
drevell, line 155: Maybe we can remove "Template" from the function name since it's already in the package name. https://google.github.io/styleguide/go/decisions#package-vs-exported-symbol-name

templates/common/templatesource/download.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/abcxyz/pkg/logging"
+)
+
+const (
+	// These will be used as part of the names of the temporary directories to
+	// make them identifiable.
+	templateDirNamePart = "template-copy-"
+)
+
+type RunParams struct {
```
drevell, line 35: This should be removed since the last refactor, right?

drevell (changes requested):

templates/common/templatesource/download.go
sailorlqh: removed. Sorry that I missed this.

templates/commands/render/render.go
sailorlqh: changed function name to `Download`

drevell:
LGTM. Perhaps remove "DO NOT MERGE" from the PR title though

drevell (approved):

sailorlqh:
> LGTM. Perhaps remove "DO NOT MERGE" from the PR title though

Yeah, just changed title to refactor.

PR #296, opened 2023-12-08
feat: implement describe subcommand

### LINKED ISSUE #62: Implement `templates describe <url>`

sailorlqh:
fix #62


templates/commands/describe/describe.go
```diff
@@ -0,0 +1,174 @@
...
+		fs:     fSys,
+		stdout: c.Stdout(),
+	})
+}
+
+// readRun provides fakable interface to test Run.
+func (c *Command) realRun(ctx context.Context, rp *runParams) (rErr error) {
+	var tempDirs []string
+	defer func() {
+		var rErr error
```
sethvargo, line 102: I don't think you want this line

templates/commands/describe/describe.go
```diff
@@ -0,0 +1,174 @@
...
+		for _, d := range tempDirs {
+			rErr = errors.Join(rErr, rp.fs.RemoveAll(d))
+		}
+	}()
+	_, templateDir, err := templatesource.Download(ctx, rp.fs, rp.describeTempDirBase, c.flags.Source, c.flags.GitProtocol)
```
sethvargo, line 107:
```suggestion

	_, templateDir, err := templatesource.Download(ctx, rp.fs, rp.describeTempDirBase, c.flags.Source, c.flags.GitProtocol)
```

templates/commands/describe/describe.go
```diff
@@ -0,0 +1,174 @@
...
+			defaultStr := i.Default.Val
+			if defaultStr == "" {
+				// When empty string is the default, print it differently so
+				// the user can actually see what's happening.
+				defaultStr = `""`
+			}
+			fmt.Fprintf(tw, "\nDefault:\t%s", defaultStr)
+		}
+		fmt.Fprintf(tw, "\n")
+		tw.Flush()
```
sethvargo, line 153: This logic seems like it could be extracted into a helper for more independent testing.

templates/commands/describe/describe.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return fmt.Errorf("error reading template spec file: %w", err)
+	}
+
+	spec, ok := specI.(*spec.Spec)
+	if !ok {
+		return fmt.Errorf("internal error: spec file did not decode to spec.Spec")
+	}
+	tw := tabwriter.NewWriter(rp.stdout, 8, 0, 2, ' ', 0)
+	fmt.Fprintf(tw, "\nTemplate:\t")
```
drevell, line 132: I'm not sure this line is necessary, it doesn't really convey any information

drevell (changes requested):

templates/commands/describe/describe.go
sethvargo: The function signature already has this as a named return parameter, so this is creating a local shadow of the same variable name.

drevell: Ah, right.

templates/commands/describe/describe.go
sailorlqh: I was thinking this can work as something like a Header

templates/commands/describe/describe.go
```diff
@@ -0,0 +1,174 @@
...
+		if i.Default != nil {
+			defaultStr := i.Default.Val
+			if defaultStr == "" {
+				// When empty string is the default, print it differently so
+				// the user can actually see what's happening.
+				defaultStr = `""`
+			}
+			fmt.Fprintf(tw, "\nDefault:\t%s", defaultStr)
+		}
+		fmt.Fprintf(tw, "\n")
```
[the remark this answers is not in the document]
sailorlqh: Yeah, I thought the same, I was planing on doing that in the next PR. I want to implement the feature first so this PR don't get blocked by the refactoring PR while you are out. 
I can work on the refactor PR and rebase this PR on that.

sailorlqh:
Hold until another PR for refactoring is merged.

templates/commands/describe/describe.go
drevell: Yeah, I see. In this case I think it might end up being confusing to the user. I was a bit confused when I saw it, like "is there supposed to be something after the colon? I already know this is a template."

templates/commands/describe/describe.go
```diff
@@ -0,0 +1,174 @@
...
+	return "describe a template base on it's spec.yaml"
+}
+
+func (c *Command) Help() string {
+	return `
+Usage: {{ COMMAND }} [options] <source>
+
+The {{ COMMAND }} command describe the given template.
+
+The "<source>" is the location of the template to be instantiated. Many forms
```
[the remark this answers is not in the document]
sailorlqh: I changed the following line to match the updated text after your fix, it this ok?

templates/commands/describe/describe_test.go
```diff
@@ -0,0 +1,268 @@
...
+- desc: 'Include some files and directories'
+  action: 'include'
+  params:
+    paths:
+      - paths: ['file1.txt', 'dir1', 'dir2/file2.txt']
+`
+
+	cases := []struct {
+		name             string
+		templateContents map[string]string
```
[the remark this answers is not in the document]
sailorlqh: I think this way is better, it can use the existing `common.WriteAllDefaultMode` to write the file to tempDir, so we don't need to write another function.

templates/commands/describe/describe.go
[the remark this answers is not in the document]
drevell: LGTM

templates/commands/describe/describe_test.go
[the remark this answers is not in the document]
drevell: I just realized that your way also lets us write a test case for spec.yaml being nonexistent. Could you add that (if you agree)?

templates/common/specutil/spec.go
```diff
@@ -0,0 +1,101 @@
...
+	OutputInputDefaultValueKey = "Default"
+	OutputInputRuleKey         = "Rule"
+)
+
+// ParseSpecToList parses a spec.Spec into a list with
+// the keys and attribute pairs.
+//
+// Example:
+// ["Description", "example description", "Input Name", "example name"].
+func ParseSpecToList(spec *spec.Spec) []string {
```
drevell, line 42: I'm not sure about the name of this function. Particularly the word "Parse". Usually parsing means going from text to an object. This is going in the other direction (object -> text).

Also, it doesn't really describe the whole spec (it doesn't include the "steps"). So maybe a name like SpecFieldsForDescribe()?

Also, it doesn't seem to be used except for in the `describe` subcommand, so maybe it doesn't belong in common?

templates/common/specutil/spec.go
```diff
@@ -0,0 +1,101 @@
...
+	l := make([]string, 0)
```
drevell, line 43: This might be little more understandable if it was a [][]string instead of a []string. Then each line of text could be it's own list, rather than using the assumption that every index multiple of 2 starts a new line.

templates/common/specutil/spec.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	return l
+}
+
+// FormatAttrList formats the attribute list for output
+//
+// Example output:
+//
+// Description:  A template for the ages
+
```
drevell, line 81: I think these empty lines should have `//` otherwise this won't be seen as one big comment and it won't all be included in the automatically generated godocs.

templates/commands/describe/describe.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return fmt.Errorf("error reading template spec file: %w", err)
+	}
+
+	spec, ok := specI.(*spec.Spec)
+	if !ok {
+		return fmt.Errorf("internal error: spec file did not decode to spec.Spec")
+	}
+
+	c.attrList = specutil.ParseSpecToList(spec)
```
drevell, line 127: Does this really need to be c.attrList? Could it be a local variable instead (just `attrList`)? Generally we should try to avoid storing state where it's not needed, which increases complexity.

drevell (changes requested):

templates/commands/describe/describe.go
sailorlqh: the purpose of this is for doing diff in the tests. It doesn't need to be c.attrList, but this makes tests easier.

templates/commands/describe/describe_test.go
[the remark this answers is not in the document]
sailorlqh: Test case added

templates/common/specutil/spec.go
sailorlqh: Done.

drevell (approved):

PR #298, opened 2023-12-08
Fix lint found by new linters

drevell:
This includes one real bug where we returned nil instead of err in templates/model/spec/v1beta1/spec.go !


sailorlqh (approved):
Thanks! I was just working on this!

sethvargo (approved):

PR #300, opened 2023-12-08
Refactor: move commonly used flags in to common/flags

sailorlqh:


templates/common/flags/flags.go
```diff
@@ -0,0 +1,14 @@
+// package flags contains flags that are commonly used by several commands
+package flags
+
+type CommonFlags struct {
```
sailorlqh, line 4: I am not sure which one is better
method 1:
implement a `func (r *CommonFlags) Register` here, so the RegisterFlags of RenderFlags,
it can just call `r.CommonFlags.Register(set)`

method 2, which is the current method:
no Register function in commonFlags, instead do this in each subcommand.
```
	g.StringVar(&cli.StringVar{
		Name:    "git-protocol",
		Target:  &r.CommonFlags.GitProtocol,
	})
```

drevell (approved):

PR #301, opened 2023-12-11
Add verify golden test content functionalities

chloechien:
This introduces 2 dependencies (sergi/go-diff, fatih/color), both been adopted by Google opensource.
Screenshot 2023-12-10 at 3 58 16 PM


templates/commands/goldentest/verify.go
```diff
@@ -60,9 +68,127 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
...
+					failureText := red(fmt.Sprintf("-- [%s] expected, however missing", goldenFile))
+					err := fmt.Errorf(failureText)
+					tcErr = errors.Join(tcErr, err)
+					continue
+				} else {
+					return fmt.Errorf("failed to read (%s): %w", tempFile, err)
+				}
+			}
+
+			diffs := dmp.DiffMain(string(tempContent), string(goldenContent), false)
```
drevell, line 132: I'm not sure what the final parameter "false" means here, and the diffmatchpatch docs don't explain. Could you clarify?

templates/commands/goldentest/verify.go
```diff
@@ -60,9 +68,127 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
...
+			}
+
+			tempContent, err := os.ReadFile(tempFile)
+			if err != nil {
+				if errors.Is(err, os.ErrNotExist) {
+					failureText := red(fmt.Sprintf("-- [%s] expected, however missing", goldenFile))
+					err := fmt.Errorf(failureText)
+					tcErr = errors.Join(tcErr, err)
+					continue
+				} else {
```
drevell, line 127: Since the previous `if` block has a continue, the convention is to avoid the unnecessary "else" here. So instead of `else {foo}` it can just be `foo` . Same at line 116.

templates/commands/goldentest/verify.go
```diff
@@ -60,9 +68,127 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
...
+
+	if merr != nil {
+		return fmt.Errorf("golden test verification failure:\n %w", merr)
+	}
+
+	return nil
+}
+
+// addTestFile collects file paths generated in a golden test.
+func addTestFile(fileSet *map[string]struct{}, testDataDir string) error {
```
drevell, line 164: Perhaps addTestFile -> addTestFiles (plural)?

templates/commands/goldentest/verify.go
```diff
@@ -60,9 +68,127 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
...
+			return err
+		}
+		if err := addTestFile(&fileSet, tempDataDir); err != nil {
+			return err
+		}
+
+		dmp := diffmatchpatch.New()
+
+		var tcErr error
+		for relPath := range fileSet {
```
drevell, line 104: In Go, the iteration order of a map is undefined, which means the order of the file comparisons is undefined. We probably want to do something deterministic here, so the test output is the same every time. Perhaps alphabetical (which would implicitly also be depth-first if we include the directory name as part of the path)?

templates/commands/goldentest/verify.go
```diff
@@ -60,9 +68,127 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
...
+
+	for _, tc := range testCases {
+		goldenDataDir := filepath.Join(c.flags.Location, goldenTestDir, tc.TestName, testDataDir)
+		tempDataDir := filepath.Join(tempDir, goldenTestDir, tc.TestName, testDataDir)
+
+		fileSet := make(map[string]struct{})
+		if err := addTestFile(&fileSet, goldenDataDir); err != nil {
+			return err
+		}
+		if err := addTestFile(&fileSet, tempDataDir); err != nil {
```
drevell, line 97: Maybe if we used LoadDirWithoutMode we wouldn't need addTestFile?

https://github.com/abcxyz/abc/blob/eb69483b155ffb205abe88448047f27a6d1d90f4/templates/common/test_utils.go#L143

drevell (changes requested):

templates/commands/goldentest/verify.go
drevell: Oops, never mind, I just remembered that LoadDirWIthoutMode requires a *testing.T.

templates/commands/goldentest/verify.go
```diff
@@ -60,9 +68,127 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
...
 
-	if _, err := parseTestCases(ctx, c.flags.Location, c.flags.TestName); err != nil {
-		return fmt.Errorf("failed to parse golden test: %w", err)
+	testCases, err := parseTestCases(ctx, c.flags.Location, c.flags.TestName)
+	if err != nil {
+		return fmt.Errorf("failed to parse golden tests: %w", err)
+	}
+
+	// Create a temporary directory to render golden tests
+	tempDir, err := renderTestCases(testCases, c.flags.Location)
```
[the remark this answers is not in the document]
chloechien: I see, I was thinking of making golden test simple, but since we already have this patter we may consider adopt that in future.

templates/commands/goldentest/verify.go
```diff
@@ -60,9 +68,127 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
...
+	defer os.RemoveAll(tempDir)
+	if err != nil {
+		return fmt.Errorf("failed to render test cases: %w", err)
 	}
 
-	return fmt.Errorf("unimplemented")
+	var merr error
+	// Highlight error message color, given diff text might be hundreds lines long.
+	red := color.New(color.FgRed).SprintFunc()
```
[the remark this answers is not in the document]
chloechien: Good catch! I haven't thought about that, updated.

templates/commands/goldentest/verify.go
```diff
@@ -60,9 +68,127 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
...
+	green := color.New(color.FgGreen).SprintFunc()
+	resultReport := "\nTest Report:\n"
+
+	for _, tc := range testCases {
+		goldenDataDir := filepath.Join(c.flags.Location, goldenTestDir, tc.TestName, testDataDir)
```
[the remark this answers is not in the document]
chloechien: It seems that both a.. and ..a are interpreted as common folder name in filepath.join, and `..` is illegal to be a folder name. So we're fine here without extra trimming. But kindly lemme know if I'm wrong.

templates/commands/goldentest/verify.go
```diff
@@ -60,9 +68,127 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
...
+		tempDataDir := filepath.Join(tempDir, goldenTestDir, tc.TestName, testDataDir)
+
+		fileSet := make(map[string]struct{})
+		if err := addTestFile(&fileSet, goldenDataDir); err != nil {
```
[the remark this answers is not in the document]
chloechien: Got that! The magic of go

templates/commands/goldentest/verify.go
chloechien: Done

templates/commands/goldentest/verify.go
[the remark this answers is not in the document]
drevell: The thing I'm worried about is the edge case where somebody has a TestName of `..`. As long as we have a test for that, or we're sure that will work OK, then I'm satisfied.

templates/commands/goldentest/verify.go
chloechien: Good idea! Added.

templates/commands/goldentest/verify.go
chloechien: Done.

templates/commands/goldentest/verify.go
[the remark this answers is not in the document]
chloechien: I've tested with a.., ..a, a..a, both tests passed. And .. is not an illegal folder name so can't be retrieved as a testname here.

templates/commands/goldentest/verify.go
```diff
@@ -60,9 +70,148 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
...
+
+	// Highlight error message color, given diff text might be hundreds lines long.
+	// Only color the text when the result is to displayed at a terminal
+	var red, green func(a ...interface{}) string
+	useColor := c.Stdout() == os.Stdout && isatty.IsTerminal(os.Stdout.Fd())
+	if useColor {
+		red = color.New(color.FgRed).SprintFunc()
+		green = color.New(color.FgGreen).SprintFunc()
+	} else {
+		red = func(a ...interface{}) string {
```
drevell, line 95: Could this be simplified to

```
red = fmt.Sprint
green = fmt.Sprint
```

templates/commands/goldentest/verify.go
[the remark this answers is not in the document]
drevell: As long as it fails cleanly in the case where testName==".." that's all I care about.

drevell (approved):

templates/commands/goldentest/verify.go
chloechien: Brilliant!

drevell (approved):

templates/commands/goldentest/verify.go
drevell: Did this get fixed already? I think I'm still seeing the old version where the order is undefined

chloechien: Yes this is an older snapshots. Being fixed in the latest one.

PR #302, opened 2023-12-12
Feat: Add test cases for common/specutil

sailorlqh:


templates/common/specutil/spec_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+func TestSingleSpecInputVarForDescribe(t *testing.T) {
+	t.Parallel()
+	spec := &spec.Spec{
+		Desc: model.String{Val: "Test Description"},
+		Inputs: []*spec.Input{
+			{
+				Name:    model.String{Val: "name1"},
+				Desc:    model.String{Val: "desc1"},
+				Default: &model.String{Val: "."},
```
drevell, line 96: Should we add a test for an input that has no default?

drevell (changes requested):

templates/common/specutil/spec_test.go
sailorlqh: Nice catch, didn't think of this.

drevell (approved):
LGTM except for the lint error

drevell (approved):

PR #307, opened 2023-12-13
Give a more helpful error when the input is not a template

drevell:
The old error message was:

```
$ abc templates render .
error opening template spec: ReadFile(): open /tmp/template-copy-2622731642/spec.yaml: no such file or directory
```

The new error message is:

```
$ abc templates render .
couldn't find spec.yaml in that directory, the provided template location "." might be incorrect
```

As part of this, we refactor some duplicated code into a common package.

Also clean up some repetitive names and out-of-date comments in specutil/spec.go.

### DIFF
--- a/templates/common/specutil/spec.go
+++ b/templates/common/specutil/spec.go
@@ -16,10 +16,15 @@
 package specutil
 
 import (
+	"context"
 	"fmt"
 	"io"
+	"os"
+	"path/filepath"
 	"text/tabwriter"
 
+	"github.com/abcxyz/abc/templates/common"
+	"github.com/abcxyz/abc/templates/model/decode"
 	spec "github.com/abcxyz/abc/templates/model/spec/v1beta1"
 )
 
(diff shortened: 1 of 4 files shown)


templates/common/specutil/spec.go
```diff
@@ -102,3 +111,28 @@ func FormatAttrList(w io.Writer, attrList [][]string) {
...
+	defer f.Close()
+
+	specI, err := decode.DecodeValidateUpgrade(ctx, f, SpecFileName, decode.KindTemplate)
+	if err != nil {
+		return nil, fmt.Errorf("error reading template spec file: %w", err)
+	}
+
+	spec, ok := specI.(*spec.Spec)
+	if !ok {
+		return nil, fmt.Errorf("internal error: spec file did not decode to spec.Spec")
```
sailorlqh, line 134: question, would changing `spec.Spec` to `v1beta1.Spec` provide more context? As this files is also `spec.go`, might cause confusion.

drevell: I don't think we want that. We want to be able to change the spec version number in the future without updating this error message.

templates/common/specutil/spec.go
```diff
@@ -34,28 +39,32 @@ const (
...
+func Describe(spec *spec.Spec) [][]string {
 	l := make([][]string, 0)
 	l = append(l, []string{OutputDescriptionKey, spec.Desc.Val})
 	return l
 }
 
-// AllSpecInputVarForDescribe parses all spec.Input values in the spec.
-func AllSpecInputVarForDescribe(spec *spec.Spec) [][]string {
+// DescribeAllInputs describes all spec.Input values in the spec.
+func DescribeAllInputs(spec *spec.Spec) [][]string {
```
sailorlqh: Thank you! This sure is a better naming.

templates/common/specutil/spec.go
sailorlqh: Thanks!

sailorlqh (approved):

capri-xiyue (approved):

PR #310, opened 2023-12-13
feat: add readme for describe

sailorlqh:
add readme for describe command.

Part of the fix for #62

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,9 @@ template input params.`
 
 ### For `abc templates describe`
 
+The describe command downloads the template and prints out its description,
+and describes the inputs that it accepts.
+
 Usages:
 - `abc templates describe <template_location>`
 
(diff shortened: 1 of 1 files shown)


chloechien (approved):

README.md
```diff
@@ -124,6 +124,17 @@ For every test case, it is expected that
   - a testdata/golden/<test_name>/test.yaml exists to define
 template input params.`
 
+### For `abc templates describe`
```
drevell, line 127: We should say what it does. Something like "The describe command downloads the template and prints out its description, and describes the inputs that it accepts."

It would also be nice to show a fully executed example, including the command output.

sailorlqh: Updated the description, also added example output.

README.md
```diff
@@ -124,6 +124,46 @@ For every test case, it is expected that
...
+
+Examples:
+- `abc templates describe github.com/abcxyz/gcp-org-terraform-template@latest`
+- `abc templates describe /my/template/dir`
+
+The `<template_location>` takes the same value as the [render](#for-abc-templates-render) command.
+
+Output Example:
+
+For template at github.com/abcxyz/guardian/abc.templates/default-workflows@v0.1.0-alpha12:
```
drevell, line 143: I'd suggest including the exact command you ran, like:

```
$ abc templates describe github.com/abcxyz/guardian/abc.templates/default-workflows@v0.1.0-alpha12
Description: ...
```

drevell (changes requested):

drevell (approved):

PR #313, opened 2023-12-14
Refactor clean-up of passing around canonical source

drevell:
There was some awkwardness around passing around a Downloader object that had already been used for downloading, just for the purpose of getting some metadata about the download. This cleanup just returns the download metadata from the Download() function, and that's passed around instead.

This avoids:
 - Undesirable statefulness in the Downloader
 - Leaking the Downloader to places that shouldn't download
 - Overly complicated tests that have to mock a Downloader despite not doing any downloading

### DIFF
--- a/templates/commands/render/manifest.go
+++ b/templates/commands/render/manifest.go
@@ -31,6 +31,7 @@ import (
 	"gopkg.in/yaml.v3"
 
 	"github.com/abcxyz/abc/templates/common"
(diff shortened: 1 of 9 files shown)


yolocs (dismissed):

drevell:
Rebased and force-pushed; this was based on a previous unsubmitted branch that has now been submitted. There are no new diffs.

yolocs (approved):

PR #317, opened 2023-12-15
Update golden-test cmd param orders.

chloechien:
- Update to `abc templates golden-test record --test-name=<test_name> <location>`
- Testname is now repeatable

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,18 +106,19 @@ The valid values for `ABC_LOG_LEVEL` are `debug`, `info`, `notice`, `warning`,
 ### For `abc templates golden-test`
 
 Usages:
-- `abc templates golden-test record --location=<template_location> <testname>`
-- `abc templates golden-test verify --location=<template_location> <testname>`
+- `abc templates golden-test record --test-name=<test_name> <location>`
+- `abc templates golden-test verify --test-name=<test_name> <location>`
 
 Examples:
-- `abc templates golden-test record --location=examples/templates/render/hello_jupiter example_test`
-- `abc templates golden-test record --location=examples/templates/render/hello_jupiter`
-- `abc templates golden-test verify --location=examples/templates/render/hello_jupiter`
+- `abc templates golden-test record --test-name=example_test examples/templates/render/hello_jupiter`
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -106,18 +106,19 @@ The valid values for `ABC_LOG_LEVEL` are `debug`, `info`, `notice`, `warning`,
 ### For `abc templates golden-test`
 
 Usages:
-- `abc templates golden-test record --location=<template_location> <testname>`
-- `abc templates golden-test verify --location=<template_location> <testname>`
+- `abc templates golden-test record --test-name=<test_name> <location>`
```
drevell, line 109: I think --test-name is optional, so the convention would be to put square brackets around it, like `[--test-name=<test_name>]`. Same on the following line.

templates/commands/goldentest/test_funcs.go
```diff
@@ -52,33 +52,35 @@ const (
...
-		if err != nil {
-			return nil, err
-		}
-		return []*TestCase{
-			{
+	testCases := []*TestCase{}
+
+	if len(testNames) > 0 {
+		for _, testName := range testNames {
+			testConfig := filepath.Join(testDir, testName, configName)
```
drevell, line 66: This block of code appears twice, here and at line 89. Can it be factored out to avoid duplication?

templates/commands/goldentest/verify.go
```diff
@@ -46,15 +46,15 @@ func (c *VerifyCommand) Desc() string {
 
 func (c *VerifyCommand) Help() string {
 	return `
-Usage: {{ COMMAND }} --location=<location> <test_name>
+Usage: {{ COMMAND }} --test-name=<test-name-1>,<test-name-2> <location>
```
drevell, line 49: Square brackets here too I think

README.md
chloechien: Thank you!

templates/commands/goldentest/test_funcs.go
chloechien: Done. Did not save lines but looks cleaner now :)

drevell (approved):

PR #318, opened 2023-12-15
Add Include Action file globbing

kevya-google:
Adds simple file globbing to the Include Action (specifically: "Paths" and "Skip", with special "As" behavior for expanding globs). Used as an UpgradeFeature in v1beta2 and v1beta3. Example usage written in README.

Long-awaited continuation of https://github.com/abcxyz/abc/pull/196.

co-authored by: @drevell


templates/commands/render/action_include.go
```diff
@@ -36,92 +37,110 @@ func actionInclude(ctx context.Context, inc *spec.Include, sp *stepParams) error
...
-			RFS:     sp.fs,
-			SrcRoot: absSrc,
-			Visitor: func(relToAbsSrc string, de fs.DirEntry) (common.CopyHint, error) {
-				if _, ok := skipNow[relToAbsSrc]; ok {
-					return common.CopyHint{
-						Skip: true,
-					}, nil
+			as := relMatchedPath
+			if len(asPaths) > 0 { // As provided
+				if len(matchedPaths) != 1 || matchedPath.Val != filepath.Join(fromDir, p.Val) {
```
drevell, line 101: Maybe factor out an `isGlob()` function to try to reduce the complexity of this function a bit

drevell (changes requested):
I'm thinking about whether we should make this change as part of a new api_version, since we're changing the interpretation of filenames containing glob characters, and changing the interpretation of `skip` to be relative to the root.

I think that would require us to support both non-globbing and globbing in the same code base, depending on the api_version. If this is horribly complex we can likely get away with not doing it, because it's unlikely to break anyone in practice and we care about feature velocity. I think nobody is using `skip` and *probably* no one is naming files in their spec.yaml that have special glob characters to be treated literally.

I think it would look like this: there would be a bool on `spec.Include` named `doGlob` that would be set if api_version is greater than v1beta1, and otherwise false. Then the code in action_include.go would behave differently depending on this bool.

Thoughts?

kevya-google:
> I'm thinking about whether we should make this change as part of a new api_version, since we're changing the interpretation of filenames containing glob characters, and changing the interpretation of `skip` to be relative to the root.

> I think it would look like this: there would be a bool on `spec.Include` named `doGlob` that would be set if api_version is greater than v1beta1, and otherwise false. Then the code in action_include.go would behave differently depending on this bool.

That sounds pretty reasonable, I'll start on implementing this and try to see how complex it ends up being.

README.md
```diff
@@ -473,15 +473,27 @@ directory. It's similar to the `COPY` command in a Dockerfile.
 Params:
 
 - `paths`: a list of files and/or directories to copy. These may use template
-  expressions (e.g. `{{.my_input}}`). Directories will be crawled recursively
+  expressions or simple file globs (e.g. `{{.my_input}}`, `*.txt`). Directories will be crawled recursively
```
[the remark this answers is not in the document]
kevya-google: prettier changed everything from single to double quotes, I think it might've been https://marketplace.visualstudio.com/items?itemName=stkb.rewrap (I had this installed from one of our pair programming sessions)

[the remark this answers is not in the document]
kevya-google: re-formatted.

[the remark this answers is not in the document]
kevya-google: Removed "simple"

README.md
```diff
@@ -473,15 +473,27 @@ directory. It's similar to the `COPY` command in a Dockerfile.
...
 - `as`: as list of output locations relative to the output directory. This can
   be used to make the output location(s) different than the input locations. If
   `as` is present, its length must be equal to the length of `paths`; that is,
-  each path must be given an output location.
+  each path must be given an output location. 
+  These may use template expressions or simple file globs (e.g. `{{.my_input}}`, `*.txt`).
+  If file globs are used in `paths`, the corresponding `as` inputs will be treated as directories:
+  ```
+  - paths: ["*.txt", "*.md", "nonglob.json"]
+    as: ["dir1", "dir2", "newname.json"]
```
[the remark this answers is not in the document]
kevya-google: Separated markdown blocks.

templates/commands/render/action_include.go
```diff
@@ -36,92 +37,110 @@ func actionInclude(ctx context.Context, inc *spec.Include, sp *stepParams) error
...
+	skipPaths, err := processGlobs(ctx, unglobbedSkipPaths, fromDir)
 	if err != nil {
 		return err
 	}
 	skip := make(map[string]struct{}, len(inc.Skip))
 	for _, s := range skipPaths {
-		skip[s.Val] = struct{}{}
+		relSkipPath, err := filepath.Rel(fromDir, s.Val)
+		if err != nil {
+			return fmt.Errorf("error making relative skip path: %w", err)
```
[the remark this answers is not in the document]
kevya-google: Added internal to errors

templates/commands/render/action_include.go
```diff
@@ -36,92 +37,110 @@ func actionInclude(ctx context.Context, inc *spec.Include, sp *stepParams) error
...
+						// scratch directory. This doesn't affect whether files in
+						// the final *destination* directory will be overwritten;
+						// that comes later.
+						Overwrite: true,
+					}, nil
+				},
+			}
+			if err := common.CopyRecursive(ctx, p.Pos, params); err != nil {
+				return p.Pos.Errorf("copying failed: %w", err)
+			}
```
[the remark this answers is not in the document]
kevya-google: Factored out getRelDst helper, made better var names.

templates/commands/render/action_include.go
kevya-google: I put this logic into a helper function, I think the comment should be sufficient to understand.

templates/commands/render/action_include.go
```diff
@@ -35,93 +34,133 @@ func actionInclude(ctx context.Context, inc *spec.Include, sp *stepParams) error
...
-
-				abs := filepath.Join(absSrc, relToAbsSrc)
-				relToFromDir, err := filepath.Rel(fromDir, abs)
-				if err != nil {
-					return common.CopyHint{}, fmt.Errorf("filepath.Rel(%s,%s)=%w", fromDir, abs, err)
-				}
-				if !de.IsDir() && inc.From.Val == "destination" {
-					sp.includedFromDest = append(sp.includedFromDest, relToFromDir)
-				}
+			relDst := getRelDst(asPaths, i, p, matchedPaths, relSrc, absSrc, fromDir)
```
yolocs, line 158: I actually think keeping the logic here would make it more readable.


```go
relDst := relSrc
if len(asPaths) > 0 {
  relDst = asPaths[i].Val
  if isGlob(...) {
    relDst = filepath.Join(relDst, relSrc)
  }
}
```

However, I did find the logic to determine whether it's a glob weird - this part:

```go
if len(matchedPaths) != 1 || absSrc != filepath.Join(fromDir, p.Val) {
```

When would `absSrc != filepath.Join(fromDir, p.Val)` happen?

templates/commands/render/action_include.go
```diff
@@ -35,93 +34,133 @@ func actionInclude(ctx context.Context, inc *spec.Include, sp *stepParams) error
...
-			},
-		}
-		if err := common.CopyRecursive(ctx, p.Pos, params); err != nil {
-			return p.Pos.Errorf("copying failed: %w", err)
+			if err := copyToDst(ctx, sp, skip, matchedPath.Pos, absDst, absSrc, relSrc, inc.From.Val, fromDir); err != nil {
+				return err
+			}
 		}
 	}
 	return nil
```
yolocs, line 166: nit
```suggestion
			return copyToDst(ctx, sp, skip, matchedPath.Pos, absDst,
			  absSrc, relSrc, inc.From.Val, fromDir)
```

kevya-google: oh yeah, nice one

kevya-google: wait, I think this actually fails because when err == nil, it stops the matchedPath loop.

templates/commands/render/action_include.go
kevya-google: Refactored to isGlob. 

`absSrc != filepath.Join(fromDir, p.Val)` occurs when the original path expands.

For example, `*.txt` only matches one file and becomes `file.txt`

kevya-google: Here's a good example:
image

kevya-google:
> I think that would require us to support both non-globbing and globbing in the same code base.

Added in the form of UpgradeFeatures.SkipGlobs in v1beta3 spec, paired on this together offline.

kevya-google:
Please merge https://github.com/abcxyz/abc/pull/339 first for UpgradeFeatures change.

templates/commands/render/action_include.go
```diff
@@ -44,9 +42,111 @@ func actionInclude(ctx context.Context, inc *spec.Include, sp *stepParams) error
...
 
-func includePath(ctx context.Context, inc *spec.IncludePath, sp *stepParams) error {
+func createSkipMap(ctx context.Context, inc *spec.IncludePath, sp *stepParams, fromDir string) (map[string]struct{}, error) {
+	skip := make(map[string]struct{}, len(inc.Skip))
+
+	unglobbedSkipPaths, err := processPaths(inc.Skip, sp.scope)
+	if err != nil {
+		return nil, err
+	}
+	skipPaths, err := processGlobs(ctx, unglobbedSkipPaths, fromDir, sp.features.SkipGlobs)
```
drevell, line 52: I think this might be less efficient than it could be. This code as written expans the skip glob into a list of files, and then checks that list for each file that gets crawled by copyParams. The first step isn't actually necessary: we can check each filename against each glob using filepath.Match(). That's the approach used at https://github.com/abcxyz/abc/blob/12290a2cf560efcb2c5b3aad606d9de1a08f0240/templates/commands/render/action_include.go#L164.

drevell (changes requested):

README.md
```diff
@@ -482,15 +482,32 @@ directory. It's similar to the `COPY` command in a Dockerfile.
...
+  expressions or file globs (e.g. `{{.my_input}}`, `*.txt`). Directories will be
+  crawled recursively and every file underneath will be processed. By default,
+  the output location of each file is the same as its location in the template
+  directory.
 - `as`: as list of output locations relative to the output directory. This can
   be used to make the output location(s) different than the input locations. If
   `as` is present, its length must be equal to the length of `paths`; that is,
-  each path must be given an output location.
+  each path must be given an output location. These may use template expressions
+  or file globs (e.g. `{{.my_input}}`, `*.txt`). If file globs are used in
```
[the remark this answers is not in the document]
kevya-google: whoops, fake news

templates/commands/render/action.go
```diff
@@ -137,7 +141,11 @@ func templateAndCompileRegexes(regexes []model.String, scope *common.Scope) ([]*
 
 // processGlobs processes a list of input String paths for simple file globbing.
 // Used after processPaths where applicable.
-func processGlobs(ctx context.Context, paths []model.String, fromDir string) ([]model.String, error) {
+func processGlobs(ctx context.Context, paths []model.String, fromDir string, skipGlobs bool) ([]model.String, error) {
+	if skipGlobs {
+		return paths, nil
```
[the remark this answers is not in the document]
kevya-google: Added comment

templates/commands/render/action_include.go
kevya-google: Isn't the current approach more efficient? Imagine we have S number of skip globs and P number of paths.

If we expand all of the skip globs and put them into the map then check if each path is in the map as they're processed (current behavior), I think this is S + P operations.

If we don't expand the skip globs but keep them as a list, for each path we encounter, we must check filepath.Match for all of the skip globs to see if it needs to be skipped. I think this is S * P operations?

drevell: Yeah, I see what you mean. There is some unpleasantness in doing S*P match operations. But those are almost free compared to the cost of filesystem operations. We can avoid S calls to filepath.Glob(), which is a system call that does IO, which will be far more expensive than an in-memory Match() operation.

kevya-google: Ok, that sounds good to me! I didn't realize that filepath.Glob was so expensive.

kevya-google: Use filepath.Match, removed skip processGlobs.

templates/commands/render/action.go
```diff
@@ -137,7 +141,12 @@ func templateAndCompileRegexes(regexes []model.String, scope *common.Scope) ([]*
 
 // processGlobs processes a list of input String paths for simple file globbing.
 // Used after processPaths where applicable.
-func processGlobs(ctx context.Context, paths []model.String, fromDir string) ([]model.String, error) {
+func processGlobs(ctx context.Context, paths []model.String, fromDir string, skipGlobs bool) ([]model.String, error) {
+	// This supports older api_versions which didn't have glob support.
+	if skipGlobs {
+		return paths, nil
```
drevell, line 147: I'm having second thoughts about this now. IIUC, this function will return relative paths of skipGlobs==true, otherwise absolute paths. Would it be possible to make it always return absolute paths? That would be less confusing for future maintainers. That would also avoid the need for the filepath.Join at line 64.

If you take this route, let's also document the absolute vs relative return values in the doc comment for this function.

drevell (changes requested):
Nice, this came out looking pretty clean, considering how complex inclusion logic is

templates/commands/render/action.go
kevya-google: Changed to convert rel -> abs

drevell (approved):

PR #319, opened 2023-12-15
Add new work-in-progress unreleased api_version v1beta2

drevell:
This is intended to be the version where @sqin2019 and @tyroneclay add their new features. When they're ready, we'll actually release this version. I added TODOs so you know where to start hacking.

I'm following the instructions in https://github.com/abcxyz/abc/blob/main/templates/model/README.md.

### DIFF
--- a/templates/model/spec/v1beta2/spec.go
+++ b/templates/model/spec/v1beta2/spec.go
@@ -1,0 +1,565 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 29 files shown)


templates/model/spec/v1beta2/spec.go
drevell, line 1: This file is copied with only TODOs added.

templates/model/spec/v1beta2/spec_test.go
drevell, line 1: File is copied without changes.

sqin2019 (approved):

yolocs (approved):

PR #320, opened 2023-12-18
feat: implement ignore feature

### LINKED ISSUE #158: Exclude special filetypes from Include via Ignore section in spec.yaml
### TL;DR

An idea was proposed to add a separate .specignore file for excluding special system files (like .DS_Store) or other private files that should not be included in template renders. Having a separate file for this would require template readers to have to go back and forth between spec.yaml and .specignore to understand how the render should behave. Instead, this could be included as a separate optional "ignore" section in the spec.yaml. Adding either a .specignore or an ignore section in spec.yaml would also support nice-to-haves such as a future IDE plugin that could detect and gray out such files.

It was decided to hold off on this feature until we have a customer who needs this functionality.

An example of this might look like:
```
kind: 'Template'
desc: 'Example template'

ignore: ['.DS_Store, '.bin', '.ssh']

inputs:
steps:
...
```

[proposal for this feature in Include Enhancements doc](https://docs.google.com/document/d/12iN2g1Idz6fW5ZqaokTmSFZf5kHXaWeewULWy2q5GfE/edit#heading=h.2wia9q1hlzad)

Idea from @sethvargo.

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

sqin2019:
fix #158


templates/commands/render/render.go
```diff
@@ -39,10 +39,12 @@ import (
 	"github.com/abcxyz/abc/templates/common"
 	"github.com/abcxyz/abc/templates/common/specutil"
 	"github.com/abcxyz/abc/templates/common/templatesource"
+	"github.com/abcxyz/abc/templates/model"
 	spec "github.com/abcxyz/abc/templates/model/spec/v1beta2"
 	"github.com/abcxyz/pkg/cli"
 	"github.com/abcxyz/pkg/logging"
 	"github.com/abcxyz/pkg/sets"
+	ign "github.com/monochromegane/go-gitignore"
```
sqin2019, line 47: I see this is used internally, but still not sure if it is ok to use this library here, if not, should we implement our own in pkg?

drevell: I'm leaning toward not using it. It looks unmaintained and kinda sketchy, and there's not much actual code thereI think we'd be better writing these few lines of code ourselves.

sqin2019: Sure, let me hold this PR until we implemented our own.

drevell (changes requested):

README.md
```diff
@@ -843,13 +843,35 @@ Params:
...
   It's analogous to the `steps` field at the top level of the spec file.
 
+### Ignore (Optional)
+
+We use gitignore-like path matching to match the file and directory paths that
+should be ignored when being included from template directory to destination
+directory. This `ignore` feature is similiar to `skip` in `include` action, the
+difference here is ignore is global and it applies to every `include` action.
+
+This section is optional, but there are some patterns will always be ignored
```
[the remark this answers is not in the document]
sqin2019: Possibility 2 sounds more natural to me, what do you think?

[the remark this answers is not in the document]
drevell: Yeah, possibility 2 SGTM.

README.md
```diff
@@ -843,13 +843,35 @@ Params:
...
+such as `.DS_Store`, `.bin`, and `.ssh`.
+
+An example of ignoring all script files in `my_dir` and its subdirectories:
+
+```yaml
+ignore:
+  - '**/*.ssh'
```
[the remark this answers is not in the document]
sqin2019: Maybe we can breakdown our tasks here, how about:
1. in this PR, we should include some basic file globing (as in [processGlob](https://github.com/abcxyz/abc/blob/1373d5297359a414b1992a924faa078af5bc9308/templates/commands/render/action.go#L140)).
2. later, we can add more and move towards gitignore if needed.
3. move abc's gitignore implemention to pkg to be shared by other projects.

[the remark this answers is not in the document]
drevell: Sounds good to me, except for #3. I don't think we should move to pkg until we've implemented something very similar to gitignore (with extended globs, `!` patterns, compatibility tests, etc). Basically I think as long as we're implementing our own thing that isn't really similar to gitignore, we shouldn't call it gitignore or imply that it's like gitignore.

templates/commands/render/render.go
sqin2019: Switched to use filepath glob.

README.md
```diff
@@ -843,13 +843,45 @@ Params:
 - `steps`: a list of steps/actions to execute in the scope of the for_each loop.
   It's analogous to the `steps` field at the top level of the spec file.
 
+### Ignore (Optional)
+
+We use [filepath Glob](https://pkg.go.dev/path/filepath#Glob) to match the file
+and directory paths that should be ignored if included/copied from template
+directory to destination directory, and should not be overwritten if included
```
sqin2019, line 850: Please correct me if it is not the desired behavior, an alternative is to remove the ignored destination files, but that seems risky.

templates/commands/render/action.go
```diff
@@ -167,6 +167,34 @@ func processGlobs(ctx context.Context, paths []model.String, fromDir string) ([]
 	return out, nil
 }
 
+// TODO (#158): There could be some refactor we can do to combine processGlobs
```
sqin2019, line 170: I am deferring the refactor in later PRs since we might want to handle "glob <glob> did not match any files" error differently, as in ignore, this error should be ignored.

README.md
```diff
@@ -843,13 +843,45 @@ Params:
...
+
+This section is optional, if not provided, a default ignore list is used:
+`.DS_Store`, `.bin`, and `.ssh`. To set your custom ignore list, please check
+accepted patterns [here](https://pkg.go.dev/path/filepath#Match).
+
+Example:
+
+```yaml
+ignore:
+  - '*/*.ssh'
```
drevell, line 863: I think we want `*/.ssh` rather than `*/*.ssh`. The intention with ignoring these is that we want to ignore the specific directory name `.ssh`, and not ignore all files ending with `.ssh`.

README.md
```diff
@@ -843,13 +843,45 @@ Params:
...
+steps:
+  - desc: 'Include some files and directories'
+    action: 'include'
+    params:
+      # ignore script files in `src_dir` but not its sub-directories.
```
drevell, line 868: I'm not sure what's meant by "script files" here

README.md
```diff
@@ -843,13 +843,45 @@ Params:
...
+      paths: ['src_dir']
+  - desc: 'Include some files and directories from destination'
+    action: 'include'
+    params:
+      # skip overwriting script files in `dest_dir` but not its sub-directories.
```
drevell, line 873: Ditto here, I think "skip overwriting" might be confusing. Just "don't include" should work.

templates/commands/render/action.go
drevell: I think it's a pretty simple refactor, and perhaps simple enough to implement now, rather than copy-pasting this function with a small change.

The return type of `processGlobs` could change from `([]model.String, error)` to `([]model.String, bool, error)`. The boolean would be `true` if there were results, or `false` if not. Each caller of this function would decide whether `false` is an error.

In general, I think copy-pasting code with minor changes creates lots of problems and should be strenuously avoided.

templates/commands/render/action_include.go
```diff
@@ -126,3 +150,32 @@ func includePath(ctx context.Context, inc *spec.IncludePath, sp *stepParams) err
 	}
 	return nil
 }
+
+// TODO(#158): add support for ignoring file name. For example, "/foo" ignores
+// a single file named foo in the root, but "foo" ignores all files named foo
+// anywhere.
+func ignorePaths(ctx context.Context, sp *stepParams, fromDir string) (map[string]struct{}, error) {
```
drevell, line 157: There's a different way this could work. I'm not sure it would be better, but we should consider it.

The way this code works, as written, is:
 1. We expand the ignore globs into a list of files using filepath.Glob(). This involves doing a bunch of filesystem operations.
 2. When we're deciding whether to ignore each file, we check if that filename is in the list of files returned from filepath.Glob()
 
This approach works, but arguably slower and more complicated than it needs to be. Step 1 doesn't actually need to happen and those slow-ish filesystem operations can be avoided.

Instead, we could use [filepath.Match()](https://pkg.go.dev/path/filepath#Match) instead of filepath.Glob(). Inside the Visitor function, we already have the filename of each file, and we're considering if it should be ignored. We can do that by iterating over each ignore glob and calling `filepath.Match(ignoreGlob, filename)`, ignoring the file if any glob returns true.

On the other hand, there is some nice simplicity in sharing code with processGlobs(), which argues in favor of the existing algorithm.

I'm leaning towards the filepath.Match() approach because I want this code to stay fast, and file IO is kinda slow.

drevell (changes requested):

README.md
sqin2019: Make sense, updating.

README.md
sqin2019: This is an example to add custom ignore list, I will use `*/*.txt` instead since it is confusing.

drevell: I see, +1 to using .txt instead

README.md
sqin2019: Updated to use txt files since it is confusing, pls let me if any concerns!

templates/commands/render/action.go
sqin2019: I thought about that approach, I was thinking to return empty list instead. I hesitated because:
1. we won't be able to report the information of the position of the pattern unless we return more information.
2. we have a default ignore list that is []string and does not have position information, unless we convert []string to []model.String with empty positions but it doesn't make sense to me.

based on 2, I think processGlobsString is needed, what do you think?

templates/commands/render/action_include.go
sqin2019: I see, filepath.Match() does sound better, let me try it. Thanks!

templates/commands/render/action_include.go
```diff
@@ -102,6 +110,22 @@ func includePath(ctx context.Context, inc *spec.IncludePath, sp *stepParams) err
 					}, nil
 				}
 
+				pathIncParent := filepath.Join(p.Val, relToAbsSrc)
```
drevell, line 113: I don't think we need to compute pathIncParent. The value we want is already being computed below as `relToFromDir`, and we can simply reuse it and call `checkIgnore(sp.ignorePatterns, relToFromDir)`.

So it would look like:

```
			Visitor: func(relToAbsSrc string, de fs.DirEntry) (common.CopyHint, error) {
				if _, ok := skipNow[relToAbsSrc]; ok {
					return common.CopyHint{
						Skip: true,
					}, nil
				}

				abs := filepath.Join(absSrc, relToAbsSrc)
				relToFromDir, err := filepath.Rel(fromDir, abs)
				if err != nil {
					return common.CopyHint{}, fmt.Errorf("filepath.Rel(%s,%s)=%w", fromDir, abs, err)
				}
				matched, err := checkIgnore(sp.ignorePatterns, relToFromDir)
				if err != nil {
					return common.CopyHint{},
						fmt.Errorf("failed to match path(%q) with ignore patterns: %w", relToFromDir, err)
				}
				if matched {
					logger.DebugContext(ctx, "path ignored", "path", relToFromDir)
					return common.CopyHint{
						Skip: true,
					}, nil
				}

				if !de.IsDir() && inc.From.Val == "destination" {
					sp.includedFromDest = append(sp.includedFromDest, relToFromDir)
				}

				return common.CopyHint{
					// Allow later includes to replace earlier includes in the
					// scratch directory. This doesn't affect whether files in
					// the final *destination* directory will be overwritten;
					// that comes later.
					Overwrite: true,
				}, nil
			},
```

templates/commands/render/action_include.go
```diff
@@ -126,3 +150,35 @@ func includePath(ctx context.Context, inc *spec.IncludePath, sp *stepParams) err
...
 	return nil
 }
+
+// TODO(#158): add support for ignoring file name. For example, "/foo" ignores
+// a single file named foo in the root, but "foo" ignores all files named foo
+// anywhere.
+// checkIgnore checks the given path against the given patterns, if given
+// patterns is not provided, a default list of patterns is used.
+func checkIgnore(patterns []model.String, path string) (bool, error) {
+	if len(patterns) > 0 {
```
drevell, line 160: I see the motivation here for copy-pasting this code block, but I'm strongly opposed to copy-pasting code most of the time, because I've seen the problems that it creates in the long run. The different copies tend to diverge and accumulate separate bugs.

In this case, I think it's OK to just declare the the defaultIgnorePatterns as a list of model.String. There's already some special-case logic for handling the case where the position is undefined: https://github.com/abcxyz/abc/blob/c9d99ef4f1b4ea314235734b442463865a5132bb/templates/model/pos.go#L58, so it's kinda OK to have a model.String with an undefined position.

So, to summarize, my recommendation is:

Declare the defaultIgnorePatterns as `[]model.String`:
```
// defaultIgnorePatterns to be used if ignore is not provided.
var defaultIgnorePatterns = []model.String{
	{Val: ".DS_Store"},
	{Val: ".bin"},
	{Val: ".ssh"},
}
```

And simplify checkIgnore:
```
func checkIgnore(patterns []model.String, path string) (bool, error) {
	if len(patterns) == 0 {
		patterns = defaultIgnorePatterns
	}
	for _, p := range patterns {
		matched, err := filepath.Match(filepath.FromSlash(p.Val), path)
		if err != nil {
			return false,
				p.Pos.Errorf("failed to match path (%q) with pattern (%q): %w", path, p.Val, err)
		}
		if matched {
			return true, nil
		}
	}
	return false, nil
}
```

templates/commands/render/action_include.go
```diff
@@ -126,3 +150,35 @@ func includePath(ctx context.Context, inc *spec.IncludePath, sp *stepParams) err
...
+		for _, p := range patterns {
+			matched, err := filepath.Match(filepath.FromSlash(p.Val), path)
+			if err != nil {
+				return false,
+					p.Pos.Errorf("failed to match path (%q) with pattern (%q): %w", path, p.Val, err)
+			}
+			if matched {
+				return matched, nil
```
drevell, line 168: Our convention would be to say `return true` if we already know that the value of `matched` is `true`.

drevell (changes requested):

templates/commands/render/action.go
sqin2019: Removed as not needed anymore.

templates/commands/render/action_include.go
sqin2019: Nice catch, thanks!

templates/commands/render/action_include.go
sqin2019: Got it, was not sure about using nil position. looks much cleaner!

drevell (approved):

PR #327, opened 2023-12-21
feat: add template version to the manifest file

drevell:
Users are likely to want to know what version of a template was rendered, and this provides it. This allows things like (1) render a template at the same version as an already-rendered template (2) grep for renderings of templates at a specific version.

Most of the complexity here is in dealing with git to retrieve info about tags and SHAs. We already have a git clone of the template repo (usually), we just have to inspect it to find out what version number we have.

There's a interesting design decision in here. To find the template version, we'll examine the git workspace and the HEAD commit. If there's a tag pointing to HEAD, then we'll use that as the version. If there are multiple such tags, we'll take the most recent in semver order. If the tags aren't semver, we tiebreak using reverse-alphabetical order. If there are no tags, we'll just use the full SHA as the version number. We never use branch names for versions because branches are too mutable.

### DIFF
--- a/templates/common/exec.go
+++ b/templates/common/exec.go
@@ -39,11 +39,9 @@ const DefaultRunTimeout = time.Minute
 // If the incoming context doesn't already have a timeout, then a default
 // timeout will be added (see DefaultRunTimeout).
 //
-// exitCode is -1 if the command could not be executed.
-//
 // If the command fails, the error message will include the contents of stdout
 // and stderr. This saves boilerplate in the caller.
-func Run(ctx context.Context, args ...string) (stdout, stderr string, exitCode int, _ error) {
+func Run(ctx context.Context, args ...string) (stdout, stderr string, _ error) {
 	if _, ok := ctx.Deadline(); !ok {
 		var cancel context.CancelFunc
 		ctx, cancel = context.WithTimeout(ctx, DefaultRunTimeout)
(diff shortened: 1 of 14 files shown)


templates/common/git/git.go
```diff
@@ -122,3 +126,190 @@ func Tags(ctx context.Context, remote string) ([]string, error) {
...
+// Workspace looks for the presence of a .git directory in parent directories
+// to determine the root directory of the git workspace containing "path".
+// Returns false if the given path is not inside a git workspace.
+//
+// The input path need not actually exist yet. For example, suppose "/a/b" is a
+// git workspace, which means that "/a/b/.git" is a directory that exists.
+// Calling Workspace("/a/b/c") will return "/a/b" whether or not "c" actually
+// exists yet. This supports the case where the user is rendering into a
+// directory that doesn't exist yet but will be created by the render operation.
+func Workspace(ctx context.Context, path string) (string, bool, error) {
```
drevell, line 139: Note to reviewers: this function is just moved, not new or changed.

templates/common/exec.go
```diff
@@ -39,9 +39,11 @@ const DefaultRunTimeout = time.Minute
 // If the incoming context doesn't already have a timeout, then a default
 // timeout will be added (see DefaultRunTimeout).
 //
+// exitCode is -1 if the command could not be executed.
+//
 // If the command fails, the error message will include the contents of stdout
 // and stderr. This saves boilerplate in the caller.
-func Run(ctx context.Context, args ...string) (stdout, stderr string, _ error) {
+func Run(ctx context.Context, args ...string) (stdout, stderr string, exitCode int, _ error) {
```
sethvargo, line 46: I'm not sure about this approach. I think it would be better to implement our own error type that embeds the exit code in the returned error. I also see `RunAllowNonZero` added, but it doesn't appear to be used anywhere?

drevell: Removed, looks like this was a vestigial bit of code that wasn't used in the final revision.

sethvargo (approved):

PR #330, opened 2023-12-28
Add the skeleton of the upgrade command

drevell:
There's no interesting upgrade logic yet. This PR just:

 - factors out some flags that were formerly specific to the `render` command that are now shared with the `upgrade` command, into the `common/flags` package.
 - Adds the `abc templates upgrade` common as a no-op. It's in "hidden" mode so it won't show up in the help text. It just loads the manifest file from the filesystem and does nothing with it.
 - Adds usage instructions.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -14,6 +14,7 @@
             "args": [
                 "templates",
                 "render",
+                "--manifest",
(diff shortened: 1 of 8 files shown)


sethvargo (approved):

sethvargo (approved):

sethvargo (approved):
:stamp:

PR #332, opened 2023-12-28
feat: add debug step diffs flag

sqin2019:
fix part of #121: Offer the ability to show a diff of changes to each file due to each action

### DIFF
--- a/templates/commands/render/action_foreach.go
+++ b/templates/commands/render/action_foreach.go
@@ -42,7 +42,7 @@ func actionForEach(ctx context.Context, fe *spec.ForEach, sp *stepParams) error 
 
 	for _, keyVal := range values {
 		subStepParams := sp.WithScope(map[string]string{key: keyVal})
-		if err := executeSteps(ctx, fe.Steps, subStepParams, ""); err != nil {
+		if err := executeSteps(ctx, fe.Steps, subStepParams); err != nil {
 			return err
 		}
 	}
--- a/templates/commands/render/action_foreach.go
+++ b/templates/commands/render/action_foreach.go
@@ -42,7 +42,7 @@ func actionForEach(ctx context.Context, fe *spec.ForEach, sp *stepParams) error 
 
 	for _, keyVal := range values {
 		subStepParams := sp.WithScope(map[string]string{key: keyVal})
-		if err := executeSteps(ctx, fe.Steps, subStepParams); err != nil {
+		if err := executeSteps(ctx, fe.Steps, subStepParams, ""); err != nil {
 			return err
 		}
 	}
--- a/templates/commands/render/render.go
+++ b/templates/commands/render/render.go
@@ -25,7 +25,6 @@ import (
 	"io"
 	"io/fs"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"sort"
 	"strings"
@@ -200,9 +199,6 @@ func (c *Command) realRun(ctx context.Context, rp *runParams) (outErr error) {
 		debugDir, err := rp.fs.MkdirTemp(rp.tempDirBase, debugDirNamePart)
 		if err != nil {
 			return fmt.Errorf("failed to create temp directory for debug directory: %w", err)
-		}
-		if err := rp.fs.MkdirAll(debugDir, common.OwnerRWXPerms); err != nil {
-			return fmt.Errorf("failed to create debug directory: MkdirAll(): %w", err)
 		}
 		sp.debugDir = debugDir
 	}
@@ -514,14 +510,13 @@ func executeSteps(ctx context.Context, steps []*spec.Step, sp *stepParams) error
 			// Make debug dir a git repository with a detached work tree in scratch
 			// dir, meaning it will track the file changes in scratch dir without
 			// affecting the scratch dir.
-			strings.Fields(fmt.Sprintf(
-				"--git-dir=%s --work-tree=%s init", sp.debugDir, sp.scratchDir)),
+			{"git", "--git-dir", sp.debugDir, "--work-tree", sp.scratchDir, "init"},
 
 			// Set git user name and email, required for ubuntu and windows os.
-			{"config", "user.name", "abc cli"},
-			{"config", "user.email", "abc@abcxyz.com"},
+			{"git", "--git-dir", sp.debugDir, "config", "user.name", "abc cli"},
+			{"git", "--git-dir", sp.debugDir, "config", "user.email", "abc@abcxyz.com"},
 		}
-		if err := runGitCmds(argsList, sp.debugDir); err != nil {
+		if err := runCmds(ctx, argsList); err != nil {
 			return err
 		}
 	}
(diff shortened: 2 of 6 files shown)


templates/commands/render/render.go
```diff
@@ -495,10 +510,40 @@ func checkInputsMissing(spec *spec.Spec, inputs map[string]string) []string {
 func executeSteps(ctx context.Context, steps []*spec.Step, sp *stepParams) error {
 	logger := logging.FromContext(ctx).With("logger", "executeSteps")
 
+	if sp.flags.DebugStepDiffs {
+		argsList := [][]string{
+			// Make debug dir a git repository with a detached work tree in scratch
+			// dir, meaning it will track the file changes in scratch dir without
+			// affecting the scratch dir.
+			strings.Fields(fmt.Sprintf(
+				"--git-dir=%s --work-tree=%s init", sp.debugDir, sp.scratchDir)),
```
sqin2019, line 519: git -C <pathA> init <pathB> will make <pathA> a git repo, not "pathB as git repo but .git stored in pathA".

templates/commands/render/render.go
```diff
@@ -193,6 +196,18 @@ func (c *Command) realRun(ctx context.Context, rp *runParams) (outErr error) {
 		templateDir:    templateDir,
 		ignorePatterns: spec.Ignore,
 	}
+
+	if c.flags.DebugStepDiffs {
+		debugDir, err := rp.fs.MkdirTemp(rp.tempDirBase, debugDirNamePart)
+		if err != nil {
+			return fmt.Errorf("failed to create temp directory for debug directory: %w", err)
+		}
+		if err := rp.fs.MkdirAll(debugDir, common.OwnerRWXPerms); err != nil {
```
drevell, line 205: It's not clear to me why we need both MkdirTmp and MkdirAll. I would have though that MkdirAll would be unnecessary because the directory already exists after running MkdirTemp.

templates/commands/render/render.go
```diff
@@ -495,10 +510,40 @@ func checkInputsMissing(spec *spec.Spec, inputs map[string]string) []string {
 func executeSteps(ctx context.Context, steps []*spec.Step, sp *stepParams) error {
 	logger := logging.FromContext(ctx).With("logger", "executeSteps")
 
+	if sp.flags.DebugStepDiffs {
+		argsList := [][]string{
+			// Make debug dir a git repository with a detached work tree in scratch
+			// dir, meaning it will track the file changes in scratch dir without
+			// affecting the scratch dir.
+			strings.Fields(fmt.Sprintf(
```
drevell, line 518: Rather than use strings.Fields, I think it would be a little cleaner to just declare the args as a string slice, like

```
{"--git-dir", sp.debugDir, "--work-tree", sp.scratchDir}
```

This avoids some possible bugs if debugDir or scratchDir somehow contain whitespace.

templates/commands/render/render.go
```diff
@@ -697,3 +755,16 @@ func destOK(fs fs.StatFS, dest string) error {
 
 	return nil
 }
+
+func runGitCmds(argsList [][]string, workingDir string) error {
```
drevell, line 759: Can we use common.Run() for this? https://github.com/abcxyz/abc/blob/main/templates/common/exec.go#L44

drevell (changes requested):

templates/commands/render/render.go
sqin2019: Make sense, thanks!

templates/commands/render/render.go
sqin2019: Updating...

templates/commands/render/render.go
```diff
@@ -508,6 +548,17 @@ func executeSteps(ctx context.Context, steps []*spec.Step, sp *stepParams) error
...
 		}
 	}
+
+	if sp.flags.DebugStepDiffs {
+		// Use default log level.
+		logger.WarnContext(
+			ctx,
+			fmt.Sprintf(
+				"Please navigate to '%s' or use 'git --git-dir=%s <command>' to see commits/diffs for each step",
+				sp.debugDir, sp.debugDir,
```
drevell, line 558: Should we include the scratch dir in this message? I'm not sure, but I think the user will need to cd into the scratch dir and use --git-dir=debugDir. A full working example command might be helpful (perhaps `git log`).

templates/commands/render/render.go
```diff
@@ -551,6 +602,8 @@ type stepParams struct {
 	scratchDir  string
 	stdout      io.Writer
 	templateDir string
+	// Temporary directory to hold debug information when debug is enabled.
```
drevell, line 605: Could we note in the comment that this field is empty when debugging is not enabled?

templates/commands/render/render.go
```diff
@@ -551,6 +602,8 @@ type stepParams struct {
...
+	debugDir string
```
drevell, line 606: Rather than making this a field of stepParams, could we make it a parameter to executeSteps? My reasoning is that stepParams is provided to all the `action*` functions (like `actionInclude()` and `actionPrint()`) and I want to avoid providing any extraneous values to those functions that they're not supposed to touch.

drevell (changes requested):
Some more opinions occurred to me, sorry for not noting these things on the first round of review.

sqin2019:
> Some more opinions occurred to me, sorry for not noting these things on the first round of review.

No worries!

templates/commands/render/render.go
sqin2019: Will add a full command and in readme too, scratch dir is already removed though with only debug-step-diffs flag, do you mean to have --keep-temp-dirs enabled too when --debug-step-diffs is enabled which might cause confusions though.

templates/commands/render/render.go
sqin2019: Will try your suggestion and see!

templates/commands/render/render.go
drevell: Oh right, I see what you mean about the scratch dir. So let's just include a full command in this log message, like you said.

sqin2019: For your concern, I tested this flag locally, it does not require me to cd into the scratch dir to use --git-dir flag.

drevell: SGTM.

templates/commands/render/action_foreach.go
```diff
@@ -42,7 +42,7 @@ func actionForEach(ctx context.Context, fe *spec.ForEach, sp *stepParams) error
 
 	for _, keyVal := range values {
 		subStepParams := sp.WithScope(map[string]string{key: keyVal})
-		if err := executeSteps(ctx, fe.Steps, subStepParams); err != nil {
+		if err := executeSteps(ctx, fe.Steps, subStepParams, ""); err != nil {
```
sqin2019, line 45: Should we pass the debugDir to substeps like in here? if yes, it will create commits for each substep in addition to the parent step, maybe not necessary.

drevell: Oh, good point! We should move the `git init` outside of executeSteps.

sqin2019: Done!

sqin2019: Changed back to have separate commits for substeps, one concern is the commit message does not have all the info (i.e.: parent step), see example:
Screenshot 2024-01-02 at 4 40 51 PM

I thought about making step a linked list, so that we can get the info of all parent steps (we will need to pass step around instead of step.ForEach), but that could be a separate topic.

Please let me know if you have any better ideas!

drevell: For now, how about we just omit the step number from the commit message and call it good?

drevell (approved):

PR #335, opened 2024-01-02
Remove superfluous call to MkdirAll for temp dir

drevell:
This was a leftover from an earlier time when the temp dir had a separate name generation step and creation step.

### DIFF
--- a/templates/commands/render/render.go
+++ b/templates/commands/render/render.go
@@ -177,9 +177,6 @@ func (c *Command) realRun(ctx context.Context, rp *runParams) (outErr error) {
 	if err != nil {
 		return fmt.Errorf("failed to create temp directory for scratch directory: %w", err)
 	}
-	if err := rp.fs.MkdirAll(scratchDir, common.OwnerRWXPerms); err != nil {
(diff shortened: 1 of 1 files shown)


sqin2019 (approved):

sethvargo (approved):
Approving since Mike is OOO, but keep in mind that this _is_ subtly differently. IIRC, the current implementation will change permissions to 0777 whereas removing this will [leave them at 0700](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/os/tempfile.go;l=94).

drevell:
> Approving since Mike is OOO, but keep in mind that this _is_ subtly differently. IIRC, the current implementation will change permissions to 0777 whereas removing this will [leave them at 0700](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/os/tempfile.go;l=94).

Actually, I think it has always been mode 0700 and will remain so. Here's the mode bits we were using previously:

https://github.com/abcxyz/abc/blob/d0acad8a477bb42771f4e43912269c5a4467ed86/templates/common/fs.go#L33

PR #339, opened 2024-01-04
Add upgradeFeatures to v1beta2 and v1beta3 specs in preparation for adding Include Globbing

kevya-google:


templates/model/spec/v1beta2/upgrade.go
```diff
@@ -32,5 +32,10 @@ func (s *Spec) Upgrade(ctx context.Context) (model.ValidatorUpgrader, error) {
 	if err := copier.Copy(&out, s); err != nil {
 		return nil, fmt.Errorf("internal error: failed upgrading spec from v1beta2 to v1beta3: %w", err)
 	}
+
+	out.UpgradeFeatures = &v1beta3.UpgradeFeatures{
+		SkipGlobs: false,
```
drevell, line 37: I think this logic needs to be more complex. In the case where the spec.yaml declares api_version v1beta1 or older, then this Upgrade() method will run after the v1beta1 Upgrade() method, which means SkipGlobs will already be set to true. We shouldn't throw that away.

So something like:

```
// Only set upgradefeatures if not already set by an earlier Upgrade() method
if s.Features == nil { 
  s.Features = &v1beta3.UpgradeFeatures {} // We can rely on the default of false
}
```

drevell (changes requested):

templates/model/spec/v1beta2/upgrade.go
kevya-google: Ah, ok. Added if.

templates/model/spec/v1beta2/upgrade.go
```diff
@@ -32,5 +32,13 @@ func (s *Spec) Upgrade(ctx context.Context) (model.ValidatorUpgrader, error) {
 	if err := copier.Copy(&out, s); err != nil {
 		return nil, fmt.Errorf("internal error: failed upgrading spec from v1beta2 to v1beta3: %w", err)
 	}
+
+	// if earlier Upgrade() already had Features, preserve them. Otherwise, use defaults.
```
drevell, line 36: Nice comment

drevell (approved):

PR #342, opened 2024-01-05
feat: ignore file name recursively

### LINKED ISSUE #158: Exclude special filetypes from Include via Ignore section in spec.yaml

sqin2019:
fix #158


templates/commands/render/action_include_test.go
```diff
@@ -443,30 +443,33 @@ func TestActionInclude(t *testing.T) {
...
+						Paths: modelStrings([]string{"."}),
 					},
 					{
 						Paths: modelStrings([]string{"."}),
 						From:  model.String{Val: "destination"},
 					},
 				},
 			},
-			ignorePatterns: modelStrings([]string{"folder1/folder2", "file2.txt"}),
+			ignorePatterns: modelStrings([]string{"folder1/folder2", "file1.txt", "*.cfg"}),
```
drevell, line 454: I don't think we have a test for ignoring a single named file without globs that's in the root directory (not a subdirectory). Could you add one?

drevell (changes requested):

templates/commands/render/action_include_test.go
sqin2019: Added line 457, let me know if you prefer separating them into separate test cases.

drevell: I think we might need still more tests. From your TODO:

```
// TODO(#158): add support for ignoring file name. For example, "/foo" ignores
// a single file named foo in the root, but "foo" ignores all files named foo
// anywhere.
```

Things we probably want tests for:
 - `/subdir/file` (file in subdir with leading slash)
 - `file` ignores file in subdir and in root
 - `/file` ignores file in root but not in subdir
 - `*.cfg` ignores files in root and in subdir
 - `/*.cfg` ignores files in root but not in subdir
 - `/*/*.cfg` ignores files in subdir but not in root and not in some other subdir like `subdir1/subdir2/foo.cfg`
 - `subdir1/subdir2` ignores `othersubdir/subdir1/subdir2`
 - `/subdir1/subdir2` ignores `/subdir1/subdir2` but not `othersubdir/subdir1/subdir2`
 
 So.. basically the combinations of (1) with/without leading slash (2) file vs directory (3) with/without globs (4) file/dir being ignored is in root vs in subdirectory

sqin2019: (1) and (4) are the same? One thing is missing in the TODO is that currently we are matching files/dirs based on their path relative to the template dir or destination dir if it is from destination, I am thinking not supporting leading slash is fine?

templates/commands/render/action_include.go
```diff
@@ -161,7 +161,15 @@ func checkIgnore(patterns []model.String, path string) (bool, error) {
 		patterns = defaultIgnorePatterns
```
sqin2019, line 161: Note: remove TODO

templates/commands/render/action_include_test.go
sqin2019: There is one edge case we cannot handle if user want to only ignore a file under destination dir without accidentally remove files with the same name in subdirs if we are using all relative path patterns. Maybe we do need to support leading slash for this, but I am still thinking if there are other workarounds...

sqin2019: Updated to support leading slash.

templates/commands/render/action_include.go
```diff
@@ -151,17 +151,25 @@ func includePath(ctx context.Context, inc *spec.IncludePath, sp *stepParams) err
...
 	for _, p := range patterns {
-		matched, err := filepath.Match(filepath.FromSlash(p.Val), path)
+		var matched bool
+		var err error
+		if filepath.Base(p.Val) == p.Val {
+			// Match file name if the pattern value is file name instead of path.
+			matched, err = filepath.Match(p.Val, filepath.Base(path))
+		} else if p.Val[0] == '/' {
+			// Match absolute path if the pattern start with a leading slash.
+			matched, err = filepath.Match(filepath.Join(fromDir, filepath.FromSlash(p.Val)), filepath.Join(fromDir, path))
```
drevell, line 168: Check my math on this, but I think this line can be simplified to:

```
matched, err = filepath.Match(filepath.FromSlash(p.Val[1:]), path)
```

drevell (changes requested):
Only some nits

templates/commands/render/action_include.go
sqin2019: You are right, I thought it is going to match the subdirs too but I was wrong. updating, thanks!

drevell (approved):

PR #344, opened 2024-01-05
Don't sign debug commits

drevell:
When the user enables `--debug-step-diffs` (debugging of per-step diffs using a temporary git repo on the side), don't sign the git commits in that repo. For people like me that have a security device requiring a touch for each signature, this means that many extra touches are required.


sqin2019 (approved):

yolocs (approved):

drevell:
This is obsolete now, the equivalent change was just submitted as https://github.com/abcxyz/abc/blob/09dcf160a776931feec0920b5947197f10fa3e0d/templates/common/render/render.go#L289. Basically this code got moved in a different PR, and while moving it I went ahead and added this flag.

PR #345, opened 2024-01-05
Refactor rendering into a Render() function instead of a command

drevell:
Template rendering has become a subroutine inside other commands. For example, the `golden-test` command already runs rendering internally, somewhat hackishly by constructing a render.Command{} struct. We're about to do something similar for the `upgrade` command. Therefore, now is the time to clean this up and make rendering a library function that can be cleanly invoked as part of any command. This resolves a TODO in golden-test that we should do exactly this.

This PR is essentially:
 - Most of the contents of **commands**/render/render.go are moved to **common**/render/render.go, and massaged a bit for clarity and cleanliness
 - The action_*.go files (like action_print.go) are likewise moved from commands/render to common/render

Still to do: migrate the tests from commands/render/render_test.go to common/render. As of this PR, all the tests for rendering are testing the command, not the library function, which isn't right. But this PR is already too big.


templates/commands/goldentest/test_funcs.go
```diff
@@ -153,20 +157,27 @@ func renderTestCase(ctx context.Context, templateDir, outputDir string, tc *Test
...
+		FS:      &common.RealFS{},
+		Inputs:  inputs,
+		Backups: false,
+		Source:  templateDir,
+		Stdout:  io.Discard, // Mute stdout from command runs.
+	}); err != nil {
+		return err //nolint:wrapcheck
+	}
+
 	return nil
```
yolocs, line 182: Can you do this or "nolint" will complian?
```suggestion
	return render.Render(ctx, &render.Params{
		Clock:   clock.New(),
		Cwd:     cwd,
		DestDir: testDir,
		FS:      &common.RealFS{},
		Inputs:  inputs,
		Backups: false,
		Source:  templateDir,
		Stdout:  io.Discard, // Mute stdout from command runs.
	}) //nolint:wrapcheck
```

templates/common/render/manifest.go
```diff
@@ -72,8 +72,8 @@ type writeManifestParams struct {
 	// in the destination directory.
 	outputHashes map[string][]byte
 
-	// The raw template source location, not necessarily canonical.
-	src string
+	// // The raw template source location, not necessarily canonical.
+	// src string
```
yolocs, line 76: Can't tell if this is intentionally commented out

yolocs (approved):

templates/commands/goldentest/test_funcs.go
drevell: Yeah, can do. I'll make this change in the following PR in the series so I don't lose my precious PR approval.

templates/common/render/manifest.go
drevell: Oops, will fix. I'll make this change in the following PR in the series so I don't lose my precious PR approval.

PR #348, opened 2024-01-05
Configure CI to run non-hermetic tests

### LINKED ISSUE #314: Configure CI to run non-hermetic tests
### TL;DR

There are some tests that don't run by default, because they are not hermetic, because they do networking:

https://github.com/abcxyz/abc/blob/6d89192ef124c4fa5bd785fff644e7d87b9e2328/templates/common/git/git_test.go#L35-L45

We should configure our CI tests to run these tests. This will help avoid us breaking something without knowing it.

This might be as simple as just seting the environment variable `ABC_TEST_NON_HERMETIC=true` in the github action YAML.

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

capri-xiyue:
fixed #314 
* Make CI to run non-hermetic tests
* Make non-skip as the default behavior for non-hermetic tests


.gitignore
drevell, line 1: This file should probably go in a separate PR

drevell (changes requested):

.gitignore
capri-xiyue: done

capri-xiyue:
Looks like https://github.com/abcxyz/pkg/blob/00353d11b4931aca33574bd674d85fafc547972c/.github/workflows/go-test.yml#L61 doesn't work with windows. I will try to see whether I can fix abcxyz/pkg to make it also support windows.

capri-xiyue:
> Looks like https://github.com/abcxyz/pkg/blob/00353d11b4931aca33574bd674d85fafc547972c/.github/workflows/go-test.yml#L61 doesn't work with windows. I will try to see whether I can fix abcxyz/pkg to make it also support windows.

I made the no-skip as the default behavior now as it probably takes longer time for me to fix https://github.com/abcxyz/pkg/blob/00353d11b4931aca33574bd674d85fafc547972c/.github/workflows/go-test.yml#L61, see https://github.com/abcxyz/pkg/pull/255

drevell:
> > Looks like https://github.com/abcxyz/pkg/blob/00353d11b4931aca33574bd674d85fafc547972c/.github/workflows/go-test.yml#L61 doesn't work with windows. I will try to see whether I can fix abcxyz/pkg to make it also support windows.
> 
> I made the no-skip as the default behavior now as it probably takes longer time for me to fix https://github.com/abcxyz/pkg/blob/00353d11b4931aca33574bd674d85fafc547972c/.github/workflows/go-test.yml#L61, see [abcxyz/pkg#255](https://github.com/abcxyz/pkg/pull/255)

I'm not sure this is the right thing to do. When a user runs `go test ./...` in the abc repo, I think it's a best practice to avoid depending on any external resources.

How about this alternative: look for the GITHUB_ACTIONS environment variable ([docs](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)). If that variable is present, then we'll go ahead and run the tests in question.

capri-xiyue:
> > > Looks like https://github.com/abcxyz/pkg/blob/00353d11b4931aca33574bd674d85fafc547972c/.github/workflows/go-test.yml#L61 doesn't work with windows. I will try to see whether I can fix abcxyz/pkg to make it also support windows.
> > 
> > 
> > I made the no-skip as the default behavior now as it probably takes longer time for me to fix https://github.com/abcxyz/pkg/blob/00353d11b4931aca33574bd674d85fafc547972c/.github/workflows/go-test.yml#L61, see [abcxyz/pkg#255](https://github.com/abcxyz/pkg/pull/255)
> 
> I'm not sure this is the right thing to do. When a user runs `go test ./...` in the abc repo, I think it's a best practice to avoid depending on any external resources.
> 
> How about this alternative: look for the GITHUB_ACTIONS environment variable ([docs](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)). If that variable is present, then we'll go ahead and run the tests in question.

Do you mean set the env like https://github.com/abcxyz/abc/pull/348/commits/dda5280a11f4076c59b88f817f34e9dcd37cb8d1, it won't work as it is a invalid use for a resuable workflow https://github.com/abcxyz/abc/actions/runs/7427742276/workflow?

drevell:
> Do you mean set the env like [dda5280](https://github.com/abcxyz/abc/commit/dda5280a11f4076c59b88f817f34e9dcd37cb8d1), it won't work as it is a invalid use for a resuable workflow https://github.com/abcxyz/abc/actions/runs/7427742276/workflow? And the GITHUB_ENV is what abcxyz/pkg is doing. I will try to make abcxyz/pkg works with windows. @drevell

I'm thinking we can leave the .github directory the same, and change https://github.com/abcxyz/abc/blob/5b467fd4e3f8897ac307e8dc41a7a7988cf76cba/templates/common/git/git_test.go#L37 from:

```
if os.Getenv(envName) == "" {
  t.Skipf("skipping test because env var %q isn't set", envName)
}
```

to

```
if os.Getenv("ABC_TEST_NON_HERMETIC) == "" && os.Getenv("GITHUB_ACTIONS") == "" {
  t.Skipf("skipping test because env var %q isn't set", envName)
}
```

I haven't tested, but I think this would work because github actions always set the GITHUB_ACTIONS env variable.

capri-xiyue:
> > Do you mean set the env like [dda5280](https://github.com/abcxyz/abc/commit/dda5280a11f4076c59b88f817f34e9dcd37cb8d1), it won't work as it is a invalid use for a resuable workflow https://github.com/abcxyz/abc/actions/runs/7427742276/workflow? And the GITHUB_ENV is what abcxyz/pkg is doing. I will try to make abcxyz/pkg works with windows. @drevell
> 
> I'm thinking we can leave the .github directory the same, and change
> 
> https://github.com/abcxyz/abc/blob/5b467fd4e3f8897ac307e8dc41a7a7988cf76cba/templates/common/git/git_test.go#L37
> 
> from:
> ```
> if os.Getenv(envName) == "" {
>   t.Skipf("skipping test because env var %q isn't set", envName)
> }
> ```
> 
> to
> 
> ```
> if os.Getenv("ABC_TEST_NON_HERMETIC) == "" && os.Getenv("GITHUB_ACTIONS") == "" {
>   t.Skipf("skipping test because env var %q isn't set", envName)
> }
> ```
> 
> I haven't tested, but I think this would work because github actions always set the GITHUB_ACTIONS env variable.

I've made abcxyz/pkg work with windows. Current the PR is ready to review now with the original approach we want to go.

drevell:
> I've made abcxyz/pkg work with windows. Current the PR is ready to review now with the original approach we want to go.

Excellent, thank you.

drevell (approved):

PR #350, opened 2024-01-08
Give a more helpful error if the user gives an invalid template location

drevell:
These are mistakes that the user is likely to make, and we want to print a more helpful error message:

Mistake 1: the user provides the path to the spec.yaml as the template path, where they're actually supposed to provide the path to the directory containing the spec.yaml. We add a very clear error message for this case.

Mistake 2: providing some other filename instead of a directory name: we provide a warning that filenames can't be parsed as local template paths.

Before this, the error message was very confusing, just `cannot overwrite a directory with a file of the same name`.

### DIFF
--- a/templates/common/fs.go
+++ b/templates/common/fs.go
@@ -157,7 +157,7 @@ func CopyRecursive(ctx context.Context, pos *model.ConfigPos, p *CopyParams) (ou
 		if err != nil {
 			return err // There was some filesystem error. Give up.
 		}
-		logger.DebugContext(ctx, "CopyRecursive handling directory entry",
+		logger.DebugContext(ctx, "handling directory entry",
 			"path", path)
 		relToSrc, err := filepath.Rel(p.SrcRoot, path)
 		if err != nil {
--- a/templates/common/fs.go
+++ b/templates/common/fs.go
@@ -157,6 +157,8 @@ func CopyRecursive(ctx context.Context, pos *model.ConfigPos, p *CopyParams) (ou
 		if err != nil {
 			return err // There was some filesystem error. Give up.
 		}
+		logger.DebugContext(ctx, "CopyRecursive handling directory entry",
+			"path", path)
 		relToSrc, err := filepath.Rel(p.SrcRoot, path)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


templates/common/fs.go
```diff
@@ -157,6 +157,8 @@ func CopyRecursive(ctx context.Context, pos *model.ConfigPos, p *CopyParams) (ou
 		if err != nil {
 			return err // There was some filesystem error. Give up.
 		}
+		logger.DebugContext(ctx, "CopyRecursive handling directory entry",
```
sethvargo, line 160: Nit: there's already a "logger" field (L152), so this is duplicating "CopyRecursive":

```json
{"logger":"CopyRecursive", "message":"CopyRecursive handling directory entry"}
```

templates/common/render/render.go
```diff
@@ -117,6 +117,7 @@ func Render(ctx context.Context, p *Params) (outErr error) {
 		outErr = errors.Join(outErr, tempRemover.maybeRemoveAll(ctx))
 	}()
 
+	logger.DebugContext(ctx, "render phase: downloading/copying template")
```
sethvargo, line 120: Same with these - they are already namespaced, so I think you can drop "render phase: ".

templates/common/render/render.go
```diff
@@ -152,10 +156,9 @@ func Render(ctx context.Context, p *Params) (outErr error) {
 	if err != nil {
 		return fmt.Errorf("failed to create temp directory for scratch directory: %w", err)
 	}
+	logger.DebugContext(ctx, "created temporary scratch directory")
```
sethvargo, line 159: I'm curious why remove the path?

templates/common/fs.go
drevell: Aha, thanks.

templates/common/render/render.go
drevell: I don't know why I did this, reverted.

templates/common/render/render.go
drevell: Done.

sethvargo (approved):

PR #354, opened 2024-01-10
Output git SHA and tag metadata from downloaders

drevell:
Output git SHA and tag metadata from downloaders

These will be used in a future PR to create template variables like `_git_sha` and `_git_tag`.

Also:
 - Move some code from the `common/git` package to the `templatesource/git`
   package. This was code that wasn't really reusable git code, it was
   specific to the template downloading and versioning logic.
 - Move tests for same
 - Break up the `BestTag` function into two functions, one in `common/git`
   and one in `templatesource/git`, for cleanliness. The sorting logic
   should be separate from the fetching logic.

### DIFF
--- a/templates/common/templatesource/source.go
+++ b/templates/common/templatesource/source.go
@@ -19,6 +19,11 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"sort"
+
+	"github.com/Masterminds/semver/v3"
+	"github.com/abcxyz/abc/templates/common/git"
+	"github.com/abcxyz/pkg/logging"
 )
 
 // sourceParser is implemented for each particular kind of template source (git,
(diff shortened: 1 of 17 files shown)


templates/common/templatesource/source.go
```diff
@@ -134,3 +139,94 @@ func ParseSource(ctx context.Context, params *ParseSourceParams) (Downloader, er
 	}
 	return parseSourceWithCwd(ctx, cwd, params)
 }
+
+// gitCanonicalVersion examines a template directory and tries to determine the
```
drevell, line 143: This code was moved from common/git/CanonicalVersion. It's not new code.

templates/common/templatesource/source.go
```diff
@@ -134,3 +139,94 @@ func ParseSource(ctx context.Context, params *ParseSourceParams) (Downloader, er
...
+	return sha, true, nil
+}
+
+// bestHeadTag returns the tag that points to the current HEAD SHA. If there are
+// multiple such tags, the precedence order is:
+//   - tags in decreasing order of semver (recent releases first)
+//   - other non-semver tags in reverse alphabetical order
+//
+// Returns false if there are no tags pointing to HEAD.
+func bestHeadTag(ctx context.Context, dir string) (string, bool, error) {
```
drevell, line 203: This code was factored out the former BestTag function. It's mostly old code.

sethvargo (approved):
Marie Kondo would be pleased.

PR #355, opened 2024-01-11
chore: clean up unused dest

capri-xiyue:
clean up unused dest


capri-xiyue:
I found `Dest` is not unused. Therefore I sent a PR to delete it. Let me know if I miss any context and delete it in the wrong way.

drevell (approved):
Thank you for taking the time for this cleanup.

PR #359, opened 2024-01-13
Streamline the tracking of features enabled in a given api_version

drevell:
Background: we have to keep the semantics of a given api_version stable even though we're constantly adding and changing features. We have a `Features` struct on the Spec model that tracks which features are enabled based on the api_version that's declared in the spec.yaml.

This PR:
 - Uses a single Features struct rather than creating separate Features structs per schema version. That was unnecessarily verbose and complex.
 - Fixes a bug in the v1beta2->v1beta3 upgrader that disregarded the value of OmitGlobs when the spec.yaml declared api_version=v1beta1.
 - Uses a non-pointer struct rather than a pointer struct. This saves a bunch of "initalize pointer to point to empty struct" code, especially in tests.

### DIFF
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -29,8 +29,6 @@ concurrency:
 jobs:
   go_lint:
     uses: 'abcxyz/pkg/.github/workflows/go-lint.yml@main' # ratchet:exclude
-    with:
-      go_version: '1.21'
 
   go_test:
     uses: 'abcxyz/pkg/.github/workflows/go-test.yml@main' # ratchet:exclude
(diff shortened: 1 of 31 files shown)


sethvargo (commented):
Why wouldn't features be tied to an API version?

verbanicm (approved):

drevell:
> Why wouldn't features be tied to an API version?

It's more like they *are* tied to an API version, but we don't need N copies of the feature flags struct. For old api_versions, the new feature flag will just be false by default, because of being initialized to the zero value.

Q: why is parsimony important here? A: because the api_version auto-upgrade code is already complex, and we can take this opportunity to have only one features struct instead of N of them, which is a nice simplification.

PR #363, opened 2024-01-16
Replace ParseSource with ParseSourceWithCwd

drevell:
Previously, we used ParseSource as the public interface for the package. It just injected the os.GetCwd() into parseSourceWithCwd. Over time, the surrounding system has evolved that the caller of ParseSource already has access to a fakeable Cwd, so the old shim system is just unnecessary complexity now.


sethvargo (approved):
Nit: if we're dropping `ParseSource`, why not rename `parseSourceWithCwd` to `ParseSource`?

drevell:
> Nit: if we're dropping `ParseSource`, why not rename `parseSourceWithCwd` to `ParseSource`?

Why not indeed!

Fixed.

sethvargo (approved):

Issue #362, opened 2024-01-16
Allow _git_* vars to be faked for testing


drevell:
As a template developer, I want to test that my spec.yaml involving vars like `_git_tag` renders as I expect. I don't want to have to create real git tags in my local git workspace to test this. Idea: have a flag like `--fake-builtin-vars=_git_tag=foo`

As a user of `abc templates golden-test`, I want to write tests for my template that hardcode a specific value of _git_sha and _git_tag that I can then assert on in the template output.

drevell:
cc @gjonathanhong

drevell:
Fixed in #369 #368 #367 #366

Issue #316, opened 2023-12-14
Add a `golden-test new-test` feature


drevell:
### TL;DR

Currently, to add a new golden test, the user must go create a file named testdata/golden/$testname/test.yaml

This has some UX issues:
 - The user has to know what directory structure is required
 - The user has to write a YAML file by hand, and know the field names

It would be more convenient if the user could create new tests from the command line:
 - Option 1: we can add flags to the existing `golden-test record` subcommand to give it the ability to create a test.yaml file. Example: `abc templates golden-test record --location . --test-name my-new-test --input service_account=abc123@foo.com --input somethingelse=bar`
 - Option 2: we add a new subcommand `abc templates golden-test new-test`. Example: `abc templates golden-test new-test --location . --test-name my-new-test --input service_account=abc123@foo.com --input somethingelse=bar`

We could also support prompting for inputs: `abc templates golden-test new-test --location . --test-name my-new-test --prompt`

cc @chloechien

capri-xiyue:
Hi @drevell,  I did a quick investigation, I prefer we use option 22 with a new subcommand `abc templates golden-test new-test` and at the same time we support prompting for inputs. I think option 2 is more clear to users as we separate the `new-test` and `record` feature and it satisfies the single responsibility principle. What do you think of it?

drevell:
> Hi @drevell, I did a quick investigation, I prefer we use option 2 with a new subcommand `abc templates golden-test new-test` and at the same time we support prompting for inputs. I think option 2 is more clear to users as we separate the `new-test` and `record` feature and it satisfies the single responsibility principle. What do you think of it?

@capri-xiyue Makes sense, sounds good to me.

drevell:
@capri-xiyue can we close this issue now? I think this might be finished.

capri-xiyue:
> @capri-xiyue can we close this issue now? I think this might be finished.

closed

PR #366, opened 2024-01-17
Add BuiltinVars to the GoldenTest YAML model

drevell:
- Creates a new version of the GoldenTest model in v1beta3 following
  templates/model/README.md.
- Add the new BuiltinVars field to the new v1beta3 goldentest model.
  This will be used to implement #362.
- Rename InputValue to VarValue; it used to be the case that all vars
  are template inputs, but that stopped being true after the
  introduction of builtin vars like `_git_tag.`


templates/commands/goldentest/test_funcs.go
```diff
@@ -28,7 +28,7 @@ import (
 	"github.com/abcxyz/abc/templates/common"
 	"github.com/abcxyz/abc/templates/common/render"
 	"github.com/abcxyz/abc/templates/model/decode"
-	goldentest "github.com/abcxyz/abc/templates/model/goldentest/v1alpha1"
+	goldentest "github.com/abcxyz/abc/templates/model/goldentest/v1beta3"
```
capri-xiyue, line 31: Curious to learn the version management in abc repo.  Looks like in `templates/model/goldentest`, we maintain different versions of goldentest for example v1alpha1 and v1beta3. But in `templates/commands/goldentest`, we use the latest version which is v1beta3. In that case, who will be the user of `templates/model/goldentest/v1alpha1`, I guess the users with older cli versions? Does the command `cli` has its own version?

templates/model/decode/decode_test.go
```diff
@@ -150,7 +151,7 @@ inputs:
   - name: 'foo'
```
capri-xiyue, line 151: Do we need to add `builtin_vars` in the test case to verify the latest goldentest works?

capri-xiyue:
Curious to learn the difference between BuiltinVars and Input. Is my understanding correct?
1. Input is defined the template spec but BuiltinVars is not?
2. Do we need to have any validation for BuiltinVars similar to Input? Looks like we don't have any specific validation besides non-zero.

In addition, curious to learn _git_*  vars CUJ in goldentest?

templates/commands/goldentest/test_funcs.go
drevell: The docs about api_version changes are a little rudimentary, but they're in templates/model/README.md (this was also linked in the PR description)

The fundamental design is that old `api_version`s are supported, they're just auto-upgraded to the latest api_version. The abc CLI app logic is written to use only the latest api_version. See the upgrade.go files and the `DecodeValidateUpgrade()` function to understand how api_version upgrading works.

templates/model/decode/decode_test.go
drevell: Yes, there's much more coming in future PRs.

drevell:
> 1. Input is defined the template spec but BuiltinVars is not?

Right.

> 2. Do we need to have any validation for BuiltinVars similar to Input? Looks like we don't have any specific validation besides non-zero.

Yes, that will be a future PR.

> In addition, curious to learn _git_* vars CUJ in goldentest?

I just sent you the internal design doc over chat.

> I'm thinking how this BuiltInVars would change the interface of #316? Like `abc templates golden-test new-test. Example: abc templates golden-test new-test --location . --test-name my-new-test --input service_account=abc123@foo.com --input somethingelse=bar --builtin_vars gitkey1=gitvalue1 --builtin_vars gitkey2=gitvalue2`, and the `--prompt` is just for input but NOT for builtin_vars as builtinVars is not defined in template?

I didn't think about, good question. Your suggestion sounds good to me. But I'd prefer `--builtin-var` (singular, no underscore) rather than (`--builtin_vars`) for consistency with the other flag names.

templates/commands/goldentest/test_funcs.go
capri-xiyue: Thanks for explanation!

capri-xiyue (approved):

PR #367, opened 2024-01-19
Allow golden-tests to override built-in vars

drevell:
We recently expanded the concept of built-in vars like `_git_tag` and `_flag_dest` (unreleased as yet). These vars need to be able to take on user-provided "fake" values in golden-tests, so the test is hermetic and deterministic; the template rendering operation inside the golden-test shouldn't change its output depending on my local git tags.

So how does this work?
 
The Render() function now takes a new parameter, `OverrideBuiltinVars`. This parameter is only used by golden-tests, and not by the other callers of Render().

The golden-test test.yaml has a top-level field `builtin_vars` that is passed through to the above parameter. This lets the user write something like:

```yaml
# For the purposes of unit testing this template, hardcode _git_tag to mytag
builtin_vars:
  - name: '_git_tag'
    value: 'mytag'
```
 
There's a bunch of new validation logic to ensure that there's no mixing of builtin variables with regular template inputs. They have distinct namespaces (underscore-prefixed for builtins) and have different ways of being set.


drevell:
Added @capri-xiyue as reviewer since this intersects with your current work.

sethvargo (approved):

PR #376, opened 2024-01-23
Improve error msg when CEL references non-overridden builtin in golden-test

drevell:
Part of #362 .

This includes:

 - Move `common/render.UnknownTemplateKeyError` to `common/errs.UnknownVar`. The package change is so it can be used by the CEL code. The struct name change is because (1) "Template" is not right if this error will be used by CEL and (2) "key" is obsolete terminology, we say "var" now.
 - Return an `UnknownVarErr` when CEL compilation fails with an "undeclared reference" error. Regrettably we have to do string parsing on compilation error, but we have to work with the error types it gives us.
 - Clean up the tests for CEL compilation and evaluation. For no good reason, we were bypassing the `CelCompileAndEval()` public API and calling `celCompile()` and `celEval()` internal functions, which added complexity to tests and missed testing an important function.
 - Clean up the error location reporting in cel.go. Rather than scatting `pos.Errorf()` around cel.go, we can just do it once in `CelCompileAndEval()` to wrap the errors coming out of `celCompile()` and `celEval()`.


templates/common/errs/unknownvar.go
```diff
@@ -0,0 +1,42 @@
...
+
+import "fmt"
+
+// UnknownVarError is an error that will be returned when a template
+// references a variable that's nonexistent.
+type UnknownVarError struct {
+	VarName       string
+	AvailableKeys []string
+	Wrapped       error
+}
```
sethvargo, line 28: Nit: Do these all need to be public? It's a tad odd to have `Wrapped` and `Unwrap`.

drevell: Since these errors are created in a few packages, we have two options:

 - Leave `Wrapped` public so it can be initialized
 - Make `Wrapped` private and provide a public constructor that initializes it.

I'm leaning towards the former just for simplicity.

sethvargo (approved):

PR #378, opened 2024-01-23
fix: template with no output should create an empty output dir

drevell:
The former behavior was that the output directory would not be created in the edge case where the template produced no output files. This was problematic for two reasons:

 - It broke `golden-test record`; it expected to find an output directory
 - Empty output is semantically valid. Suppose there's a template that has a bunch of optional steps that are all skipped.

Also:
 - Rename `expectGoldenContents` to `want` for consistency with convention
 - Remove an unnecessary call to os.RemoveAll(). This is a vestige of the old days before we recorded into a temp dir.


drevell:
Closing this, we're changing the way this is going to work.

PR #379, opened 2024-01-24
Add new test feature in goldentest

capri-xiyue:
1. Add new test feature 
2. builtin_vars support will be in another PR.

### DIFF
--- a/cmd/abc/abc.go
+++ b/cmd/abc/abc.go
@@ -1,4 +1,4 @@
-// Copyright 2023 The Authors (see AUTHORS file)
+// Copyright 2024 The Authors (see AUTHORS file)
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/cmd/abc/abc.go
+++ b/cmd/abc/abc.go
@@ -1,4 +1,4 @@
-// Copyright 2024 The Authors (see AUTHORS file)
+// Copyright 2023 The Authors (see AUTHORS file)
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/cmd/abc/abc.go
+++ b/cmd/abc/abc.go
@@ -53,6 +53,9 @@ var rootCmd = func() *cli.RootCommand {
 								Name:        "golden-test",
 								Description: "subcommands for validating template rendering with golden tests",
 								Commands: map[string]cli.CommandFactory{
+									"new-test": func() cli.Command {
+										return &goldentest.NewTestCommand{}
+									},
 									"record": func() cli.Command {
 										return &goldentest.RecordCommand{}
 									},
--- a/templates/commands/goldentest/new_test_cli.go
+++ b/templates/commands/goldentest/new_test_cli.go
@@ -16,13 +16,15 @@
 package goldentest
 
 // This file implements the "templates golden-test new-test" subcommand.
+
 import (
 	"context"
 	"errors"
 	"fmt"
-	"gopkg.in/yaml.v3"
 	"os"
 	"path/filepath"
+
+	"gopkg.in/yaml.v3"
 
 	"github.com/abcxyz/abc/templates/common"
 	"github.com/abcxyz/abc/templates/common/input"
@@ -46,7 +48,7 @@ func (c *NewTestCommand) Desc() string {
 
 func (c *NewTestCommand) Help() string {
 	return `
-Usage: {{ COMMAND }} [--test-name=<test-name-1>]
+Usage: {{ COMMAND }} [options] <test_name> <location>
 
 The {{ COMMAND }} create a new golden test.
 
@@ -92,6 +94,10 @@ func (c *NewTestCommand) Run(ctx context.Context, args []string) (rErr error) {
 		// using localSourceParser instead of remoteGitSourceParser.
 		GitProtocol: "",
 	})
+	if err != nil {
+		return err //nolint:wrapcheck
+	}
+
 	spec, err := specutil.Load(ctx, fs, templateDir, c.flags.Location)
 	if err != nil {
 		return err //nolint:wrapcheck
@@ -109,32 +115,19 @@ func (c *NewTestCommand) Run(ctx context.Context, args []string) (rErr error) {
 		return err //nolint:wrapcheck
 	}
 
-	testCase := goldentest.Test{
-		Inputs: mapToVarValues(resolvedInputs),
-	}
-
-	buf, err := yaml.Marshal(testCase)
+	buf, err := marshalTestCase(resolvedInputs)
 	if err != nil {
-		return fmt.Errorf("failed marshaling test case when writing: %w", err)
+		return fmt.Errorf("failed to marhsal test config data: %w", err)
 	}
-	header := map[string]string{
-		"api_version": decode.LatestAPIVersion,
-		"kind":        decode.KindGoldenTest,
-	}
-	headerYAML, err := yaml.Marshal(header)
-	if err != nil {
-		return fmt.Errorf("failed marshaling api_version: %w", err)
-	}
-
-	buf = append(headerYAML,
-		buf...)
 
 	testDir := filepath.Join(c.flags.Location, goldenTestDir, c.flags.NewTestName)
 	testConfigFile := filepath.Join(testDir, configName)
+
 	err = fs.MkdirAll(testDir, common.OwnerRWXPerms)
 	if err != nil {
 		return fmt.Errorf("failed creating %s directory to contain test yaml file: %w", testDir, err)
 	}
+	// file overriding is not allowed.
 	fh, err := fs.OpenFile(testConfigFile, os.O_CREATE|os.O_EXCL|os.O_WRONLY, common.OwnerRWPerms)
 	if err != nil {
 		return fmt.Errorf("OpenFile(%q): %w", testConfigFile, err)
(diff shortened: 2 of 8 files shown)


templates/commands/goldentest/new_test_cli.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	return nil
+}
+
+func marshalTestCase(resolvedInputs map[string]string) ([]byte, error) {
+	testCase := goldentest.Test{
+		Inputs: mapToVarValues(resolvedInputs),
+	}
+	buf, err := yaml.Marshal(testCase)
```
capri-xiyue, line 149: As I use yaml marshal here, the output yaml file won't have single or double quote as the existing example showed https://github.com/abcxyz/abc/blob/1c91e857127e78537695ec7a02efe9e1dc8ff2ad/examples/templates/render/for_each_dynamic/testdata/golden/multiple_envs/test.yaml#L5. But it is still a valid yaml file and I verified the `record` and `verify` still works without single quote or double quote. 

I googled it, looks like golang yaml package can't support marshal with quote?

capri-xiyue:
@drevell  It is ready for initial review as the functionality works. I will write the UT later in the same PR.

templates/commands/goldentest/new_test_cli.go
```diff
@@ -0,0 +1,165 @@
...
+	flags NewTestFlags
+}
+
+func (c *NewTestCommand) Desc() string {
+	return "create a new golden test"
+}
+
+func (c *NewTestCommand) Help() string {
+	return `
+Usage: {{ COMMAND }} [options] <test_name> <location>
```
drevell, line 51: Idea: maybe location should default to the current directory if not specified?

templates/commands/goldentest/new_test_cli.go
```diff
@@ -0,0 +1,165 @@
...
+
+The {{ COMMAND }} create a new golden test.
+
+The "<test_name>" is the name of the test.
+
+The "<location>" is the location of the template.
```
drevell, line 57: Maybe we should say here that the template location must be a local directory.

templates/commands/goldentest/new_test_cli.go
```diff
@@ -0,0 +1,165 @@
...
+	}()
+
+	wd, err := c.WorkingDir()
+	if err != nil {
+		return fmt.Errorf("failed to get working directory: %w", err)
+	}
+
+	fs := &common.RealFS{}
+	logger.DebugContext(ctx, "downloading/copying template")
+	_, templateDir, err := templatesource.Download(ctx, &templatesource.DownloadParams{
```
drevell, line 89: I think maybe the call to Download() is unnecessary. Since the template is already required to be in a local directory, we can load the spec file from that directory without needing to do a download.

templates/commands/goldentest/new_test_cli.go
```diff
@@ -0,0 +1,165 @@
...
+		Prompter: c,
+		Spec:     spec,
+	})
+	if err != nil {
+		return err //nolint:wrapcheck
+	}
+
+	buf, err := marshalTestCase(resolvedInputs)
+	if err != nil {
+		return fmt.Errorf("failed to marhsal test config data: %w", err)
```
drevell, line 120: marhsal -> marshal

templates/commands/goldentest/new_test_flags.go
```diff
@@ -0,0 +1,89 @@
...
+type NewTestFlags struct {
+	// Positional arguments:
+
+	// Location is the file system location of the template to be tested.
+	//
+	// Example: t/rest_server.
+	Location string
+
+	// NewTestName are the name of the test case to create.
+	NewTestName string
```
drevell, line 35: I think we should enforce that this name doesn't contain any slashes or backslashes. Those will cause problems with filesystem operations.

templates/commands/goldentest/new_test_cli.go
drevell: That seems fine to me, thanks.

templates/commands/goldentest/new_test_cli.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	testDir := filepath.Join(c.flags.Location, goldenTestDir, c.flags.NewTestName)
+	testConfigFile := filepath.Join(testDir, configName)
+
+	err = fs.MkdirAll(testDir, common.OwnerRWXPerms)
+	if err != nil {
+		return fmt.Errorf("failed creating %s directory to contain test yaml file: %w", testDir, err)
+	}
+	// file overriding is not allowed.
```
drevell, line 130: overriding -> overwriting

capri-xiyue, line 130: Maybe we can add forceOverwrite option to customize overwrite behavior?

templates/commands/goldentest/new_test_flags.go
```diff
@@ -0,0 +1,89 @@
...
+		// we prompted the user, they'd be tempted to enter something creative
+		// rather than realize that their playbook needs to be updated.
+		Default: false,
+
+		Usage: "Prompt the user for template inputs that weren't provided as flags.",
+	})
+
+	// Default NewTestName to the first CLI argument, if given
+	set.AfterParse(func(existingErr error) error {
+		r.NewTestName = strings.TrimSpace(set.Arg(0))
```
drevell, line 72: Is the call to TrimSpace necessary?

drevell (changes requested):
Looks great overall, just some small thoughts.

templates/commands/goldentest/new_test_cli.go
drevell: That sounds good to me.

templates/commands/goldentest/new_test_cli.go
capri-xiyue: Yep, will remove it.

templates/commands/goldentest/new_test_cli.go
```diff
@@ -0,0 +1,165 @@
+// Copyright 2023 The Authors (see AUTHORS file)
```
sethvargo, line 1: Nit: 2024 (throughout)

templates/commands/goldentest/new_test_cli.go
capri-xiyue: done. I've changed the location from positional argument to optional argument. Let me know if it is what you were thinking?

templates/commands/goldentest/new_test_cli.go
capri-xiyue: done

templates/commands/goldentest/new_test_cli.go
capri-xiyue: done

templates/commands/goldentest/new_test_flags.go
capri-xiyue: done

templates/commands/goldentest/new_test_cli.go
capri-xiyue: done

templates/commands/goldentest/new_test_cli.go
capri-xiyue: done

templates/commands/goldentest/new_test_cli.go
capri-xiyue: will do it in another PR

cmd/abc/abc.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2023 The Authors (see AUTHORS file)
+// Copyright 2024 The Authors (see AUTHORS file)
```
sethvargo, line 1: Actually, don't update this. See go/copyright, but you only change the copyright for a totally _new_ file.

templates/model/goldentest/v1beta3/goldentest.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2023 The Authors (see AUTHORS file)
+// Copyright 2024 The Authors (see AUTHORS file)
```
sethvargo, line 1: Same here, don't change this.

templates/commands/goldentest/test_funcs.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2023 The Authors (see AUTHORS file)
+// Copyright 2024 The Authors (see AUTHORS file)
```
sethvargo, line 1: Same here, don't change this.

sethvargo (commented):
LGTM, but I'll let Dave do the final approval

cmd/abc/abc.go
capri-xiyue: done

templates/commands/goldentest/new_test_cli_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	specYaml := `apiVersion: 'cli.abcxyz.dev/v1beta3'
+kind: 'Template'
+
+desc: 'An example template that demonstrates the "print" action'
+
+inputs:
+  - name: 'name'
+    desc: 'the name of the person to greet'
+    default: 'Alice'
```
drevell, line 42: Can we also test the case where an input has no default?

This can be a future PR.

templates/commands/goldentest/new_test_cli_test.go
```diff
@@ -0,0 +1,137 @@
...
+				return
+			}
+
+			gotContents := common.LoadDirWithoutMode(t, filepath.Join(tempDir, "testdata/golden/", tc.newTestName))
+			if diff := cmp.Diff(gotContents, tc.expectedContents); diff != "" {
+				t.Errorf("dest directory contents were not as expected (-got,+want): %s", diff)
+			}
+		})
+	}
+}
```
drevell, line 137: Perhaps we should also add a test for prompting. There's a test https://github.com/abcxyz/abc/blob/76f640dbba1f6519cad9e5aef6ded790851f914b/templates/common/render/render_test.go#L1116 that could be used as an example to follow.

This can be a future PR.

drevell (approved):

templates/commands/goldentest/new_test_cli_test.go
capri-xiyue: Sure. I will add it in another PR.

templates/commands/goldentest/new_test_cli_test.go
capri-xiyue: Will do.

capri-xiyue: I wrote a draft PR. But when I took a look at the original example, https://github.com/abcxyz/abc/blob/76f640dbba1f6519cad9e5aef6ded790851f914b/templates/common/render/render_test.go#L1116 is not testing against the `render` command, it is testing against `Resolve` method https://github.com/abcxyz/abc/blob/76f640dbba1f6519cad9e5aef6ded790851f914b/templates/common/render/render_test.go#L1408. Since `Resolve` has its own UT, I think we don't need to write another test the `Resolve` again in `new-test` as it has already been coverred?

drevell: Ideally I'd like for there to be an integration test that makes sure that new-test is integrating with Resolve correctly. As long as it's not a huge effort. I may have forgotten to test this with Render() though 🫤

capri-xiyue: I tried it but got some error of failing to read the prompt msg https://github.com/abcxyz/abc/blob/452374ecc1637863ff9ba5ef1899e503d6467cc2/templates/common/render/render_test.go#L1541 when I tested against the new-test interface. Let me try to debug what happened.

PR #382, opened 2024-01-25
feat: added builtin vars support

capri-xiyue:
1. added builtin vars support
2. added test case with no default spec yaml

### DIFF
--- a/templates/commands/goldentest/new_test_cli.go
+++ b/templates/commands/goldentest/new_test_cli.go
@@ -27,6 +27,7 @@ import (
 	"gopkg.in/yaml.v3"
 
 	"github.com/abcxyz/abc/templates/common"
+	"github.com/abcxyz/abc/templates/common/builtinvar"
 	"github.com/abcxyz/abc/templates/common/input"
 	"github.com/abcxyz/abc/templates/common/specutil"
 	"github.com/abcxyz/abc/templates/model/decode"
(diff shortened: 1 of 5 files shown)


templates/common/builtinvar/builtin_vars.go
```diff
@@ -68,3 +68,15 @@ func NamesInScope(f features.Features) []string {
 
 	return out
 }
+
+// ValidateBuiltinVars returns error if any of the attemptedNames are not valid builtin
+// var names.
+func ValidateBuiltinVars(attemptedNames []string) error {
```
drevell, line 74: The set of builtin vars depends on the api_version, so we can't hardcode it like this. But the good news is that the function that we need already exists, it's the `Validate()` function on line 49 of this file.

drevell (changes requested):

templates/common/builtinvar/builtin_vars.go
capri-xiyue: done

templates/commands/goldentest/new_test_cli.go
```diff
@@ -99,7 +100,16 @@ func (c *NewTestCommand) Run(ctx context.Context, args []string) (rErr error) {
 		return err //nolint:wrapcheck
 	}
 
-	buf, err := marshalTestCase(resolvedInputs)
+	builtinVarsKeys := make([]string, 0, len(c.flags.BuiltinVars))
+	for k := range c.flags.BuiltinVars {
+		builtinVarsKeys = append(builtinVarsKeys, k)
+	}
+	err = builtinvar.Validate(spec.Features, builtinVarsKeys)
+	if err != nil {
```
sethvargo, line 108: Nit: cuddle errors.

capri-xiyue: fixed by https://github.com/abcxyz/abc/pull/388

PR #383, opened 2024-01-25
feat: made location as a default positional argument

capri-xiyue:
made location as a default positional argument

### DIFF
--- a/templates/commands/goldentest/new_test_cli.go
+++ b/templates/commands/goldentest/new_test_cli.go
@@ -48,7 +48,9 @@ func (c *NewTestCommand) Desc() string {
 
 func (c *NewTestCommand) Help() string {
 	return `
-Usage: {{ COMMAND }} [options] <test_name>
+Usage: {{ COMMAND }} [options] <test_name> <location>
+
+If no "<location>" is given, default to current directory.
 
 The {{ COMMAND }} create a new golden test.
 
--- a/templates/commands/goldentest/new_test_cli.go
+++ b/templates/commands/goldentest/new_test_cli.go
@@ -48,7 +48,7 @@ func (c *NewTestCommand) Desc() string {
 
 func (c *NewTestCommand) Help() string {
 	return `
-Usage: {{ COMMAND }} [options] <test_name> <location>
+Usage: {{ COMMAND }} [options] <test_name> [<location>]
 
 The {{ COMMAND }} create a new golden test.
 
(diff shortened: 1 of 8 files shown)


templates/commands/goldentest/new_test_cli.go
```diff
@@ -48,11 +48,13 @@ func (c *NewTestCommand) Desc() string {
 
 func (c *NewTestCommand) Help() string {
 	return `
-Usage: {{ COMMAND }} [options] <test_name>
+Usage: {{ COMMAND }} [options] <test_name> <location>
```
drevell, line 51: I think the convention would be to add square brackets, like `[<location>]`, for things that are optional. We'd need to do this for `record` and `verify` too.

templates/commands/goldentest/new_test_flags.go
```diff
@@ -112,4 +100,16 @@ func (r *NewTestFlags) Register(set *cli.FlagSet) {
 		}
 		return nil
 	})
+
+	// Default Location to the second CLI argument, if given
+	// If not given, default to current directory.
+	set.AfterParse(func(existingErr error) error {
+		r.Location = strings.TrimSpace(set.Arg(1))
+
+		if r.Location == "" {
```
drevell, line 109: Is there a way to test this? Some examples of flag testing are in abc_test.go and commands/render/render_test.go.

drevell (changes requested):

templates/commands/goldentest/new_test_cli.go
capri-xiyue: done

templates/commands/goldentest/new_test_flags.go
capri-xiyue: done

drevell (approved):

templates/commands/goldentest/new_test_cli_test.go
```diff
@@ -238,3 +239,74 @@ builtin_vars:
...
+			var cmd NewTestCommand
+			cmd.SetLookupEnv(cli.MapLookuper(nil))
+
+			err := cmd.Flags().Parse(tc.args)
+			if err != nil || tc.wantErr != "" {
+				if diff := testutil.DiffErrString(err, tc.wantErr); diff != "" {
+					t.Fatal(diff)
+				}
+				return
+			}
```
sethvargo, line 306: Do we want the nil check here? I think we might want to call `DiffErrString` without the surrounding conditional.

capri-xiyue: Thanks for catching it! I was blindly copy the existing test structure in abc goldentest. I sent another PR to fix all of them https://github.com/abcxyz/abc/pull/387

PR #384, opened 2024-01-25
fix: template with no output should create an empty output dir

drevell:
The former behavior was that the output directory would not be created
in the edge case where the template produced no output files. This was
problematic for two reasons:

 - It broke `golden-test record`; it expected to find an output directory
 - Empty output is semantically valid. Suppose there's a template that
   has a bunch of optional steps that are all skipped.

This requires changes to template rendering, golden-test recording, and golden-test verification.

I also decided to use the `.abc` directory name for both manifests and golden-test recording metadata. It's simple and consistent to have just one metadata directory name that we use everywhere.


templates/commands/goldentest/test_funcs.go
```diff
@@ -156,10 +156,6 @@ func renderTestCases(ctx context.Context, testCases []*TestCase, location string
 func renderTestCase(ctx context.Context, templateDir, outputDir string, tc *TestCase) error {
 	testDir := filepath.Join(outputDir, goldenTestDir, tc.TestName, testDataDir)
 
-	if err := os.RemoveAll(testDir); err != nil {
```
capri-xiyue, line 159: Is this the wrong logic that we want to remove?

templates/common/paths.go
```diff
@@ -0,0 +1,36 @@
...
+package common
+
+import (
+	"path/filepath"
+	"strings"
+)
+
+// ABCInternalDir is the name for internal directories that have things like
+// manifests and .gitkeep files. These may be destination directories or
+// golden-test data/ directories.
```
capri-xiyue, line 24: nits is this a typo for `golden-test data/`

templates/commands/goldentest/test_funcs.go
drevell: This was supposed to be deleted in a previous PR but seems to have been lost in a merge conflict, so I'm deleting it again

templates/common/paths.go
drevell: I think it's good as written

templates/commands/goldentest/record.go
```diff
@@ -112,6 +112,16 @@ func (c *RecordCommand) Run(ctx context.Context, args []string) error {
...
 		}
 		merr = errors.Join(merr, common.CopyRecursive(ctx, nil, params))
+
+		abcInternal := filepath.Join(testDir, common.ABCInternalDir)
+		if err := os.MkdirAll(abcInternal, common.OwnerRWXPerms); err != nil {
+			return fmt.Errorf("failed to create dir %q: %w", abcInternal, err)
+		}
+		// git won't commit an empty directory, so add a placeholder file.
+		gitKeep := filepath.Join(abcInternal, ".gitkeep")
+		if err := os.WriteFile(gitKeep, []byte{}, common.OwnerRWPerms); err != nil {
```
sethvargo, line 122: I believe you can pass `nil` in for the content.

sethvargo (approved):

templates/commands/goldentest/record.go
drevell: Merging now to avoid the velocity hit for another round of review, will fix in a separate PR.

PR #385, opened 2024-01-25
New integration test for record+verify

drevell:
We had tests for golden-test recording and verification separately, but not together. We need to verify that the directory created by the `record` command can be correctly handled by the `verify` command. I found a bug in my code by running CLI commands by hand that probably should have been covered by a test suite. This closes the gap.


templates/commands/goldentest/record_verify_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+// TestRecordVerify tests that the output of the record command can be verified by the verify
+// command.
+func TestRecordVerify(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		name            string
+		testNames       []string
+		templateContent map[string]string
+		messWith        func(*testing.T, string)
```
capri-xiyue, line 35: what's the meaning of messWith?

templates/commands/goldentest/record_verify_test.go
```diff
@@ -0,0 +1,148 @@
+// Copyright 2023 The Authors (see AUTHORS file)
```
capri-xiyue, line 1: nits: 2024

capri-xiyue (dismissed):

templates/commands/goldentest/record_verify_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/abcxyz/abc/templates/common"
+	"github.com/abcxyz/pkg/logging"
+	"github.com/abcxyz/pkg/testutil"
+)
+
+// TestRecordVerify tests that the output of the record command can be verified by the verify
+// command.
+func TestRecordVerify(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
```
sethvargo, line 31:
```suggestion
	t.Parallel()

	cases := []struct {
```

templates/commands/goldentest/record_verify_test.go
drevell: Added a comment.

templates/commands/goldentest/record_verify_test.go
drevell: Done.

capri-xiyue (approved):

PR #386, opened 2024-01-26
fix: fixed empty case and overwrite behavior in goldentest new-test

capri-xiyue:
1. fixed empty case when there is no inputs or built in vars
2. fixed overwrite behavior, previously it can't handle truncate case.

### DIFF
--- a/templates/model/goldentest/v1beta3/goldentest.go
+++ b/templates/model/goldentest/v1beta3/goldentest.go
@@ -56,8 +56,8 @@ type Test struct {
 
 // TestCase represents a test.yaml describing test configs.
 type TestCase struct {
-	APIVersion  string      `yaml:"api_version,omitempty"`
-	Kind        string      `yaml:"kind,omitempty"`
+	APIVersion  string      `yaml:"api_version"`
+	Kind        string      `yaml:"kind"`
 	Inputs      []*VarValue `yaml:"inputs,omitempty"`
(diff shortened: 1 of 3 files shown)


templates/model/goldentest/v1beta3/goldentest.go
```diff
@@ -54,6 +54,14 @@ type Test struct {
 	BuiltinVars []*VarValue `yaml:"builtin_vars,omitempty"`
 }
 
+// TestCase represents a test.yaml describing test configs.
```
drevell, line 57: It's not immediately obvious to me why we need both Test and TestCase

drevell (changes requested):

templates/model/goldentest/v1beta3/goldentest.go
capri-xiyue: If we don't have `TestCase` here and use the `Test`, as the the inputs and built in vars of `Test` can be empty, yaml marshal will emit a string as `{}`. Therefore I use a `TestCase` here with the api version and kind that won't be empty. Another approach would be use the `Test` but I need to check whether it is a empty struct or not and then use yaml marshal, I feel it is kind of less clean code.

drevell: I sent you https://github.com/abcxyz/abc/pull/390 as an alternative approach to addressing this, let me know what you think

PR #389, opened 2024-01-26
Test: prompt  UT in goldentest new-test

capri-xiyue:
Add prompt UT in goldentest new-test

### DIFF
--- a/templates/commands/goldentest/new_test_cli.go
+++ b/templates/commands/goldentest/new_test_cli.go
@@ -92,11 +92,12 @@ func (c *NewTestCommand) Run(ctx context.Context, args []string) (rErr error) {
 	logger.DebugContext(ctx, "resolving inputs")
 
 	resolvedInputs, err := input.Resolve(ctx, &input.ResolveParams{
-		FS:       fs,
-		Inputs:   c.flags.Inputs,
-		Prompt:   c.flags.Prompt,
-		Prompter: c,
-		Spec:     spec,
+		FS:                 fs,
+		Inputs:             c.flags.Inputs,
+		Prompt:             c.flags.Prompt,
+		Prompter:           c,
+		Spec:               spec,
+		SkipPromptTTYCheck: true,
 	})
 	if err != nil {
 		return err //nolint:wrapcheck
--- a/templates/commands/goldentest/new_test_cli.go
+++ b/templates/commands/goldentest/new_test_cli.go
@@ -40,6 +40,9 @@ type NewTestCommand struct {
 	cli.BaseCommand
 
 	flags NewTestFlags
+
+	// used in prompt UT.
+	skipPromptTTYCheck bool
 }
 
 func (c *NewTestCommand) Desc() string {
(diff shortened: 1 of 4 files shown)


templates/commands/goldentest/new_test_cli.go
```diff
@@ -92,11 +92,12 @@ func (c *NewTestCommand) Run(ctx context.Context, args []string) (rErr error) {
...
-		Inputs:   c.flags.Inputs,
-		Prompt:   c.flags.Prompt,
-		Prompter: c,
-		Spec:     spec,
+		FS:                 fs,
+		Inputs:             c.flags.Inputs,
+		Prompt:             c.flags.Prompt,
+		Prompter:           c,
+		Spec:               spec,
+		SkipPromptTTYCheck: true,
```
capri-xiyue, line 100: Not sure whether I can set it to true blindly. I tested the cli, it still works. Let me know if we should only set it to True just in UT.

templates/commands/goldentest/new_test_cli_test.go
```diff
@@ -307,3 +310,127 @@ func TestNewTestFlags_Parse(t *testing.T) {
 		})
 	}
 }
+
+func TestNewTestPrompt(t *testing.T) {
```
capri-xiyue, line 314: I feel it is cleaner to separate the prompt test case to its own test instead of combining them with other test cases that won't need some golang channel.

templates/commands/goldentest/new_test_cli.go
drevell: I don't think we should set it blindly. There's benefit to doing the TTY check for real, which is why we implemented it. We can add a field that's only set in tests that is plumbed through here. LIke:

```
type NewTestCommand struct {
	cli.BaseCommand

    skipPromptTTYCheck bool // <-- here

	flags NewTestFlags
}

...
resolvedInputs, err := input.Resolve(ctx, &input.ResolveParams{
  SkipPromptTTYCheck: c.SkipPromptTTYCheck,.
  ...
```

templates/commands/goldentest/new_test_cli_test.go
drevell: Yeah, a separate test makes sense to me, like you've done it here.

templates/commands/goldentest/new_test_cli_test.go
```diff
@@ -307,3 +310,127 @@ func TestNewTestFlags_Parse(t *testing.T) {
...
+  - desc: 'Print a personalized message'
+    action: 'print'
+    params:
+      message: 'Hello, {{.name}}!'
+`
+
+	cases := []struct {
+		name               string
+		newTestName        string
+		flagInputs         map[string]string
```
drevell, line 336: Some of these fields aren't used and can be removed

drevell (changes requested):

templates/commands/goldentest/new_test_cli.go
capri-xiyue: done

drevell (approved):

PR #390, opened 2024-01-26
Cleaner marshaling of headers when writing YAML

drevell:
An attempted principled approach to marshaling output YAML files with header fields without using maps or copy-pasting struct definitions.

This was prompted by an alternative approach in #386 .


capri-xiyue:
Curious to learn your thoughts regarding the pros and cons of this and https://github.com/abcxyz/abc/pull/386

drevell:
> Curious to learn your thoughts regarding the pros and cons of this and #386. This PR will get rid of the duplicate of header definition across different abc models?

Yeah, I see the benefits as:

 - Avoids having two copies of the GoldenTest, Manifest, etc structs. This is the main benefit in my mind. Having multiple copy-pasted copies of the same structs with slight differences (two added fields) is absolute poison for codebase quality and maintainability.
 - This solution avoids the awkwardness of calling yaml.Marshal twice
 - This solution avoids the awkwardness of using an ad-hoc map for the yaml header

@capri-xiyue

capri-xiyue:
> > Curious to learn your thoughts regarding the pros and cons of this and #386. This PR will get rid of the duplicate of header definition across different abc models?
> 
> Yeah, I see the benefits as:
> 
> * Avoids having two copies of the GoldenTest, Manifest, etc structs. This is the main benefit in my mind. Having multiple copy-pasted copies of the same structs with slight differences (two added fields) is absolute poison for codebase quality and maintainability.
> * This solution avoids the awkwardness of calling yaml.Marshal twice
> * This solution avoids the awkwardness of using an ad-hoc map for the yaml header
> 
> @capri-xiyue

Let's use yours. I will close my PR.

capri-xiyue (dismissed):

templates/commands/goldentest/new_test_cli.go
```diff
@@ -142,24 +144,20 @@ func (c *NewTestCommand) Run(ctx context.Context, args []string) (rErr error) {
 }
 
 func marshalTestCase(inputs, builtinVars map[string]string) ([]byte, error) {
```
capri-xiyue, line 146: Can you add a test case to make sure the simple header works, see example here  https://github.com/abcxyz/abc/pull/386/files#diff-a3e6090ddcab7dc561a353a6ec6d5df1847d1452887fffda4c00c29cf93c815fR121

templates/common/render/manifest.go
```diff
@@ -112,19 +113,9 @@ func writeManifest(ctx context.Context, p *writeManifestParams) (rErr error) {
...
-	headerYAML, err := yaml.Marshal(header)
-	if err != nil {
-		return fmt.Errorf("internal error: failed marshaling api_version: %w", err)
-	}
-
 	buf = append(
 		[]byte("# Generated by the \"abc templates\" command. Do not modify.\n"),
-		append(headerYAML,
-			buf...)...)
+		buf...)
```
capri-xiyue, line 118: Is it possible to have a manifest just with header? I guess no? I'm asking this for whether we need to add a test case with just header avaiailable?

templates/commands/goldentest/new_test_cli.go
drevell: Copied yours, thanks.

templates/common/render/manifest.go
drevell: I don't think so, there are several required fields in the manifest.

capri-xiyue (approved):

sethvargo (approved):

PR #395, opened 2024-01-27
Require that all templates have golden tests

drevell:
In #394 we added golden tests for every template. This check will ensure that all templates going forward will have golden tests.

UPDATE: added a blocklist of templates that are untestable due to #399.


require_golden_tests.sh
```diff
@@ -0,0 +1,18 @@
...
+
+exit_status=0
+for spec_path in $(find . -name spec.yaml) ; do
+     template_dir=$(dirname $spec_path)
+     if [ ! -d "$template_dir/testdata/golden" ] ; then
+        echo "::error title=Missing golden test for template::$template_dir"
+        exit_status=1
+     fi
+done
+exit $exit_status
```
sethvargo, line 18: Nit: Newline EOF

require_golden_tests.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
sethvargo, line 1:
```suggestion
#!/usr/bin/env bash
```

sailorlqh (approved):

Issue #311, opened 2023-12-13
Support running all golden tests for all templates under a directory


drevell:
Currently, the `abc templates golden-test` command tests only one template at a time. It would be great if we could run all the tests for all the templates in a repo/directory with a single command. This command would crawl the directory tree looking for templates, and whenever one is found, all goldentests for that template would be executed.

EDIT: we should definitely support `verify`, but maybe also `record`, for every template that's found?

cc @chloechien

capri-xiyue:
@drevell  I'm thinking about what the UX would look like?

- Option 1: `abc templates golden-test record --recursive examples/templates` and  `abc templates golden-test verify --recursive examples/templates` Here the --location is no longer the the location of the template, just a repo/directory that could includes multiple locations of templates. And `--test-name` no longer make sense here when we have multiple templates, we need to enforce that `--test-name` has to be empty when `--recursive` flag is used.
- Option 2: Have a new subcommand like `abc templates golden-test record-all <location>` and `abc templates golden-test verify-all <location>`

I kind of feel option 2 is better as 
- option 1 forces users to remember different valid option combination when they set location to the location of the template or just a repo/directory that could include multiple templates. For example, `--test-name` has to be empty when `--recursive` flag is used.
- in option 1, the  meaning of `location` is confusing.
What do you think of it?

drevell:
I agree, I like your option 2.

capri-xiyue:
> * just a repo/directory that could includes multiple locations of templates

@drevell raised another option to  extend the functionality to handle multiple templates just in `record` and `verify`

in `record` and `verify`, `location` means the the repo that could include one or more templates(could be a template location or it could be a repo that includes multiple templates. 
The pros is that users will have a single command that always does what they want. The cons is that the meaning of `location` is not consistent between `record`, `verify` and `new-test`.
In `new-test`, location has to be template location
In `record` and `verify`, location could be a template location or a repo that includes multiple templates.

Seems like people prefer one command over consistency meaning of `location` here.
I will implement with the new option mentioned in this thread.

PR #396, opened 2024-01-27
Move the golden-test runner from the GH action yaml to a .sh file

drevell:
Before, you couldn't run this script manually, because it's embedded in the GitHub action YAML file. Now you can run it at will on your local machine.


run_golden_tests.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# This script is intended to be invoked in a github action. It will run all the
+# the golden tests that it can find.
+
+set -eEuo pipefail
+
+exit_status=0
+for template_dir in $(find . -name 'testdata' | grep -v '/.git/' | xargs dirname) ; do
+    echo "Running golden tests for $template_dir"
+    go run cmd/abc/abc.go templates golden-test verify $template_dir
```
sethvargo, line 11: I wonder if this is an indication that we should have a built-in flag like `-recurse` and have the `golden-test` command handle the nesting?

drevell: I agree. We have #311 for that, coming soon.

sethvargo (approved):
LGTM, but CI is unhap.

sethvargo (approved):

mattrandallbecker (approved):
+100 to shift left for CI failures

PR #406, opened 2024-01-31
chore: adjust error message for golden-test verify

### LINKED ISSUE #393: Adjust error messages for `golden-test verify`
### TL;DR

We can make this a little more user-friendly.

As a user, when I run the `golden-test verify` command, if the test.yaml exists but the `data` directory doesn't exist, then we should suggest running the `record` command to create the `data` directory. Something like `there's no recorded golden output for this test, you might need to run the "record" command to create it."

As a user, when I run the `golden-test verify` command, if the verify command fails because the actual output doesn't match the expected output, then we should suggest that they should consider running the `record` command if the new output working as intended and should replace the previous golden output. Something like `if the new template output is correct, you might consider running the "record" command to capture it as the new expected output.`

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

capri-xiyue:
fixed #393

### DIFF
--- a/templates/commands/goldentest/verify.go
+++ b/templates/commands/goldentest/verify.go
@@ -164,15 +164,8 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
 			}
 		}
 
-		if goldenContentFilesNotExist {
-			failureText := red(fmt.Sprintf("golden test [%s] not recorded in test data, you might "+
-				"need to run 'record' command to create it", tc.TestName))
-			err := fmt.Errorf(failureText)
-			tcErr = errors.Join(tcErr, err)
-		}
-
-		if filesMismatch {
-			failureText := red(fmt.Sprintf("golden test [%s] file content mismatch, you might "+
+		if goldenContentFilesNotExist || filesMismatch {
+			failureText := red(fmt.Sprintf("golden test [%s] didn't match actual output, you might "+
 				"need to run 'record' command to capture it as the new expected output", tc.TestName))
 			err := fmt.Errorf(failureText)
 			tcErr = errors.Join(tcErr, err)
--- a/templates/commands/goldentest/verify.go
+++ b/templates/commands/goldentest/verify.go
@@ -124,6 +124,7 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
 
 		var tcErr error
 		goldenContentFilesNotExist := false
+		fileMismatch := false
 		for _, relPath := range relPaths {
 			goldenFile := filepath.Join(goldenDataDir, relPath)
 			tempFile := filepath.Join(tempDataDir, relPath)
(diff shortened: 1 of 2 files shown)


templates/commands/goldentest/verify.go
```diff
@@ -130,7 +130,8 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
 			goldenContent, err := os.ReadFile(goldenFile)
 			if err != nil {
 				if errors.Is(err, os.ErrNotExist) {
-					failureText := red(fmt.Sprintf("-- [%s] generated, however not recorded in test data", goldenFile))
+					failureText := red(fmt.Sprintf("-- [%s] generated, however not recorded in test data, you might "+
```
drevell, line 133: It's probably more user-friendly to print this once per failing test, rather than once per file.

drevell (changes requested):

templates/commands/goldentest/verify.go
capri-xiyue: updated. I still keep this failureText to let users know which file is missing recorded test data in case they don't use record subcommand and write the test data by themselves.

templates/commands/goldentest/verify.go
```diff
@@ -157,9 +160,24 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
 				failureText := red(fmt.Sprintf("-- [%s] file content mismatch", goldenFile))
 				err := fmt.Errorf("%s:\n%s", failureText, dmp.DiffPrettyText(diffs))
 				tcErr = errors.Join(tcErr, err)
+				filesMismatch = true
 			}
 		}
 
+		if goldenContentFilesNotExist {
```
drevell, line 167: Can we combine these into a single message? Like `Golden test contents didn't match actual output, you might need to ...`

Then we only need one bool and one `if` block.

drevell (changes requested):

templates/commands/goldentest/verify.go
capri-xiyue: updated. I combined them but I think we still need two "bool"? How can it work with just one bool as file not exist and file not match are two cases.

templates/commands/goldentest/verify.go
```diff
@@ -123,6 +123,8 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
 		dmp := diffmatchpatch.New()
 
 		var tcErr error
+		goldenContentFilesNotExist := false
+		filesMismatch := false
```
drevell, line 127: I think these two bools can be combined into one (goldenContentFilesNotExist and filesMismatch)

drevell (changes requested):

templates/commands/goldentest/verify.go
capri-xiyue: done

drevell (approved):

Issue #402, opened 2024-01-30
Integrate abc repo with github.com/google-github-actions/send-google-chat-webhook


drevell:
We'd like to have notifications in the dev chat space whenever:
 - A release is successfully released
 - A GitHub issue is opened
 - Maybe other stuff?

PR #409, opened 2024-02-01
feat: create notify on issue and release

sailorlqh:
please merge after https://github.com/google-github-actions/send-google-chat-webhook/pull/16 is merged and new release (v0.0.2) is created.

fixes: https://github.com/abcxyz/abc/issues/402

### DIFF
--- a/.github/workflows/notify-on-issue-and-release.yml
+++ b/.github/workflows/notify-on-issue-and-release.yml
@@ -2,18 +2,17 @@ name: 'notify-on-issue-and-release'
 
 on:
   issues:
-    types: [opened]
+    types: ['opened']
   release:
-    types: [released]
+    types: ['released']
   workflow_dispatch:
-     
 
 jobs:
   call_action:
     runs-on: 'ubuntu-latest'
     permissions:
-        contents: 'read'
-        id-token: 'write'
+      contents: 'read'
+      id-token: 'write'
     steps:
       - name: 'Checkout'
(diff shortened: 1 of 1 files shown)


.github/workflows/notify-on-issue-and-release.yml
```diff
@@ -0,0 +1,22 @@
...
+  call_action:
+    runs-on: 'ubuntu-latest'
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    steps:
+      - name: 'Checkout'
+        uses: 'actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3' # ratchet:actions/checkout@v3
+      - name: 'call action'
+        uses: 'google-github-actions/send-google-chat-webhook@v0.0.2'
```
drevell, line 20: Should we pin this to a specific commit, similarly to the checkout action at line 18? That would also save us from waiting for a release of the google-github-actions repo.

drevell (changes requested):

.github/workflows/notify-on-issue-and-release.yml
sailorlqh: The pin doesn't work, cause the action downloads go binary base on the release tag and run that binary, currently v0.0.1 doesn't support issue and release, and v0.0.2 is not released yet.

drevell: Ah, I see.

drevell (approved):

capri-xiyue (approved):

PR #410, opened 2024-02-02
feat: verify stdout in goldentest

### LINKED ISSUE #356: Golden tests should record and verify standard output
### TL;DR

Currently, the `golden-test` command just asserts that the output *files* of a template match the expected files. It could also assert that the *stdout* of the template also matches the expected value.

This would involve:
 - Changing the `golden-test record` command to save the stdout from template rendering
 - Changing the `golden-test verify` command to compare the actual vs expected stdout

cc @chloechien @capri-xiyue

capri-xiyue:
fixed #356

### DIFF
--- a/templates/commands/goldentest/verify.go
+++ b/templates/commands/goldentest/verify.go
@@ -204,14 +204,14 @@ func addTestFiles(fileSet map[string]struct{}, testDataDir string) error {
 			return fmt.Errorf("filepath.Rel(%s,%s): %w", testDataDir, path, err)
 		}
 
-		// Don't assert the contents of ".abc". As of this writing, the .abc
-		// dir contains things that are specific to recorded tests and not part
+		// Don't assert the contents of ".abc" except ".abc/.stdout".
+		// As of this writing, the .abc dir except ".abc/.stdout" contains things that are specific to recorded tests and not part
 		// of the expected template output.
 		if common.IsReservedInDest(relToSrc) {
-			if de.IsDir() {
-				return fs.SkipDir
+			// skip files under ".abc" directory except ".abc/.stdout".
+			if !common.IsReservedStdout(relToSrc) {
+				return nil
 			}
-			return nil
 		}
 		if de.IsDir() {
 			return nil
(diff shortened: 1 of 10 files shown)


yolocs (approved):

capri-xiyue:
@drevell Merged for now as you are OOO. Feel free to review it when you come back. I can create other PRs to fix the comments if there is any.

templates/common/paths.go
```diff
@@ -22,7 +22,10 @@ import (
 // ABCInternalDir is the name for internal directories that have things like
 // manifests and .gitkeep files. These may be destination directories or
 // golden-test data/ directories.
-const ABCInternalDir = ".abc"
+const (
+	ABCInternalDir    = ".abc"
+	ABCInternalStdout = ".stdout"
```
drevell, line 27: I don't think this name should start with a dot. Typically that's for "hidden" files, but this file is already in a hidden directory so it doesn't need to be hidden. For example, if you look in `~/.vscode` and `~/.config`, most files in there don't start with a dot.

templates/commands/goldentest/verify.go
```diff
@@ -204,14 +204,14 @@ func addTestFiles(fileSet map[string]struct{}, testDataDir string) error {
...
-		// Don't assert the contents of ".abc". As of this writing, the .abc
-		// dir contains things that are specific to recorded tests and not part
+		// Don't assert the contents of ".abc" except ".abc/.stdout".
+		// As of this writing, the .abc dir except ".abc/.stdout" contains things that are specific to recorded tests and not part
 		// of the expected template output.
 		if common.IsReservedInDest(relToSrc) {
-			if de.IsDir() {
-				return fs.SkipDir
+			// skip files under ".abc" directory except ".abc/.stdout".
+			if !common.IsReservedStdout(relToSrc) {
```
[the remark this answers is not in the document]
capri-xiyue: fixed by https://github.com/abcxyz/abc/pull/419

templates/common/paths.go
capri-xiyue: fixed by https://github.com/abcxyz/abc/pull/419

PR #412, opened 2024-02-05
chore: removed windows support

### LINKED ISSUE #405: Remove Windows support
There's a doc circulating internally about removing Windows support due to the moderate cost and minimal benefit. Once that doc is approved, we can remove Windows support.

This will involve:
 - Removing windows from the the CI matrix in lint-test.yml
 - Simplifying all the filepath logic including .FromSlash and .ToSlash 
 - Removing the windows build from .goreleaser.yaml
 - Removing test cases that apply only to windows
 - Ctrl-f for "windows" and start removing stuff.

We should probably also error on startup if `runtime.GOOS=="windows"`, rather than just being silently broken.

capri-xiyue:
fixed #405


pdewilde (approved):

sailorlqh (approved):

templates/common/test_utils.go
```diff
@@ -47,26 +46,8 @@ var (
 	MinimalGitHeadShortSHA = MinimalGitHeadSHA[:7]
 )
 
-// CmpFileMode is a cmp option that handles the conflict between Unix and
-// Windows systems file permissions.
+// CmpFileMode is a cmp option that to compare the file mode.
 var CmpFileMode = cmp.Comparer(func(a, b fs.FileMode) bool {
```
drevell, line 50: I think CmpFileMode can be removed. It was only needed because of Windows.

capri-xiyue: fixed by https://github.com/abcxyz/abc/pull/418

PR #414, opened 2024-02-06
feat: renamed gitignore file in abc goldentest

### LINKED ISSUE #399: golden-tests can break when recorded output has a .gitignore file
When a template outputs a .gitignore file, it will be saved in in the `testdata/golden/$testname/data` directory. Git will see this and ignore certain files under that directory. This isn't what we want; we need *all* files in the directory to be committed in order for the recorded test output to be accurate.

Possible solutions:
 1. (preferred) rename certain files when saving them into `testdata/golden/$testname/data`. The renaming could look like:`.gitignore` -> `.gitignore.abc_renamed`.
 2. Rather than saving golden test outputs in a directory, we could save all template outputs in a single compressed file. That would hide the .gitignore file where it can't be seen by git.

This problem affects the template `/t/nextjs_react_template` in this repo.

capri-xiyue:
fixed #399


templates/commands/goldentest/test_funcs.go
```diff
@@ -212,3 +213,22 @@ func mapToVarValues(m map[string]string) []*goldentest.VarValue {
 	}
 	return out
 }
+
+func renameGitignoreFiles(dir string) error {
```
drevell, line 217: I think we should support more than just `.gitignore` here. Maybe we should handle any file or directory beginning with `.git*`? That would support `.gitattributes` and it would support outputting a `.git` directory if anybody wants to do that.

templates/commands/goldentest/verify.go
```diff
@@ -84,6 +84,9 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
 	if err != nil {
 		return fmt.Errorf("failed to render test cases: %w", err)
 	}
+	if err := renameGitignoreFiles(tempDir); err != nil {
```
drevell, line 87: The problem with this approach is that we will show incorrect filenames to the user if there's a difference between the temp dir and the golden dir. The user shouldn't see `.abc_renamed` in that message.

One alternative that I think might work: when reading the contents of the golden dir, remove and ignore the `.abc_renamed` suffix in memory, for files that begin with `.git*`.


I think we also need a test case that looks at the filenames that would catch problems like this.

drevell (changes requested):

templates/commands/goldentest/verify.go
capri-xiyue: I may fail to get the point. If in the `record` subcommand, we renamed the `.gitignore` to `.gitignore.abc_renamed`, I think it makes sense to make user see `.abc_renamed` in that message because that's what we emit in `record` subcommand?

templates/commands/goldentest/test_funcs.go
capri-xiyue: done

templates/commands/goldentest/verify.go
capri-xiyue: Done. And I use file suffix trimming and emphasize on original name.

templates/commands/goldentest/test_funcs_test.go
```diff
@@ -584,7 +584,69 @@ steps:
...
+				"subfolder2/.gitignore": "subfolder2 gitignore contents",
+				"file1.txt":             "file1",
+			},
+			want: map[string]string{
+				"subfolder1/.gitignore.abc_renamed": "subfolder1 gitignore contents",
+				"subfolder2/.gitignore.abc_renamed": "subfolder2 gitignore contents",
+				"file1.txt":                         "file1",
+			},
+		},
+	}
```
drevell, line 631: Perhaps also have a test case where a directory is renamed containing a file that is also renamed?

drevell (changes requested):

templates/commands/goldentest/test_funcs.go
```diff
@@ -212,3 +219,33 @@ func mapToVarValues(m map[string]string) []*goldentest.VarValue {
...
+	// Need to rename files first and then dirs, otherwise you will encounter dir not found error as the original dir has been renamed.
+	err := filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !d.IsDir() && strings.HasPrefix(d.Name(), gitPrefix) {
+			newPath := path + ".abc_renamed"
+			if err := os.Rename(path, newPath); err != nil {
+				return fmt.Errorf("error renaming file %s: %w", path, err)
```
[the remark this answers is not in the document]
capri-xiyue: We can have `.abc_renamed.abc_renamed`
But will it be better that we make`abc_renamed` is a reserved suffix(maybe we can add some check in render later to make sure users won't provide files end with `abc_renamed`)?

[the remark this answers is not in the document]
drevell: Yeah, that does seem simpler. SGTM.

templates/commands/goldentest/test_funcs.go
```diff
@@ -55,6 +56,12 @@ const (
 	// The golden test config file is always located in the test case root dir and
 	// named test.yaml.
 	configName = "test.yaml"
+
+	// The prefix of git related directories and files.
+	gitPrefix = ".git"
+
+	//
```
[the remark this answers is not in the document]
capri-xiyue: done

templates/commands/goldentest/record_test.go
```diff
@@ -197,6 +197,23 @@ steps:
...
 			},
 		},
+		{
+			name: "test_with_gitignore_succeeds",
+			filesContent: map[string]string{
+				"spec.yaml":                      specYaml,
+				"a.txt":                          "file A content",
+				"b.txt":                          "file B content",
+				"testdata/golden/test/test.yaml": testYaml,
+				".gitignore":                     "gitignore contents",
```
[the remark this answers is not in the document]
capri-xiyue: I tested it in local. It has to be a real `.git` directory that can be achieved via `git init`. I can't figure out a easy way to create such a real `.git` directory in the UT. Do you have any suggestions?

templates/commands/goldentest/test_funcs.go
```diff
@@ -212,3 +219,33 @@ func mapToVarValues(m map[string]string) []*goldentest.VarValue {
...
+func renameGitDirsAndFiles(dir string) error {
+	var gitDirPaths []string
+	// Need to rename files first and then dirs, otherwise you will encounter dir not found error as the original dir has been renamed.
+	err := filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !d.IsDir() && strings.HasPrefix(d.Name(), gitPrefix) {
+			newPath := path + ".abc_renamed"
```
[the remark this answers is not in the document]
capri-xiyue: done

templates/commands/goldentest/test_funcs.go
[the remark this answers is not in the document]
capri-xiyue: Will do the check in another PR.

templates/commands/goldentest/test_funcs.go
```diff
@@ -212,3 +219,33 @@ func mapToVarValues(m map[string]string) []*goldentest.VarValue {
 	}
 	return out
 }
+
+func renameGitDirsAndFiles(dir string) error {
+	var gitDirPaths []string
+	// Need to rename files first and then dirs, otherwise you will encounter dir not found error as the original dir has been renamed.
```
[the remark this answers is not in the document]
capri-xiyue: done

templates/commands/goldentest/record_test.go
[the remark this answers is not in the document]
drevell: I think we could create a dir named `.gitfoo` containing just an empty text file.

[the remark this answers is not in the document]
capri-xiyue: done

templates/commands/goldentest/test_funcs.go
capri-xiyue: I made `.github` an exception. I don't think we should rename `.github`

drevell: I think we SHOULD rename .github. If there's a template that outputs a .github directory, then that .github directory should not affect how the golden data directory is handled by github. Or maybe I'm missing your point.

capri-xiyue: Updated.

templates/commands/goldentest/test_funcs.go
```diff
@@ -214,21 +220,32 @@ func mapToVarValues(m map[string]string) []*goldentest.VarValue {
...
 		}
 		return nil
 	})
+	if err != nil {
+		return fmt.Errorf("WalkDir: %w", err) // There was some filesystem error while crawling.
+	}
 
-	return err //nolint:wrapcheck
+	// rename in reverse order, otherwise you will rename directory before you rename the files under the specific directory.
+	for i := len(gitPaths) - 1; i >= 0; i-- {
```
drevell, line 243: Optional: use https://pkg.go.dev/golang.org/x/exp/slices#Reverse to reverse the slice, then iterate over it normally using a `range` expression.

templates/commands/goldentest/test_funcs.go
```diff
@@ -214,21 +220,32 @@ func mapToVarValues(m map[string]string) []*goldentest.VarValue {
 	return out
 }
 
-func renameGitignoreFiles(dir string) error {
+func renameGitDirsAndFiles(dir string) error {
+	// including path of git related directories and files.
+	var gitPaths []string
+	// Need to rename files first and then dirs, otherwise you will encounter dir not found error as the original dir has been renamed.
```
drevell, line 226: This comment is now obsolete, thanks for the cleanup

templates/commands/goldentest/test_funcs.go
```diff
@@ -214,21 +220,32 @@ func mapToVarValues(m map[string]string) []*goldentest.VarValue {
...
 
-		if !d.IsDir() && d.Name() == ".gitignore" {
-			newPath := path + ".abc_renamed"
-
-			if err := os.Rename(path, newPath); err != nil {
-				return fmt.Errorf("error renaming file %s: %w", path, err)
-			}
+		if strings.HasPrefix(d.Name(), gitPrefix) {
+			gitPaths = append(gitPaths, path)
+			return nil
```
drevell, line 234: Arguably this line is unnecessary

drevell (changes requested):

templates/commands/goldentest/test_funcs.go
capri-xiyue: done

drevell (approved):

PR #415, opened 2024-02-07
abc-updater integration

pdewilde:
Integrate abcxyz/abc-updater. This checks a remote server for any new updates and notifies uses in stderr if a new version is available.


sethvargo:
https://github.com/abcxyz/abc/pull/417

pdewilde:
(will merge soon now that updater is public)

cmd/abc/abc.go
```diff
@@ -15,13 +15,17 @@
...
 	"context"
 	"fmt"
 	"os"
 	"os/signal"
 	"runtime"
+	"sync"
 	"syscall"
+	"time"
 
+	"github.com/abcxyz/abc-updater/sdk/go/abc-updater"
```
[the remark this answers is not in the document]
pdewilde: Yeah, that's a good point, I inherited this project, but I assume the reasoning was to make a distinction between the sdk and any backend services, as well as to provide for multiple SDK's in the future.

Do you think just doing something like `github.com/abcxyz/abc-updater/pkg/sdk` would make more sense?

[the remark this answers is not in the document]
sethvargo: Why do we need the "pkg" layout? The name of the folder will be the default name of the package ("sdk" in the example of github.com/abcxyz/abc-updater/pkg/sdk). I guess I'd prefer github.com/abcxyz/abc-updater/pkg/abcupdater as the root package if we need to stick to the pkg setup.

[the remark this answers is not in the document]
pdewilde: I don't care about `pkg`, thought that was the convention for locating non-executables which aren't internal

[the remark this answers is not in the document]
sethvargo: Then I'd go with `github.com/abcxyz/abc-updater/pkg/abcupdater`

cmd/abc/abc.go
```diff
@@ -104,9 +107,56 @@ func setLogEnvVars() {
...
+		default:
+		}
+	}()
+
+	return func() {
+		select {
+		case <-ctx.Done():
+			// Context was cancelled
+		case msg := <-updatesCh:
+			logging.FromContext(ctx).WarnContext(ctx, msg)
```
pdewilde, line 141: I am having an issue where the logger is escaping the output from updater

```
time=2024-03-27T11:44:58.635-07:00 severity=WARNING message="A new version of abc is available! Your current version is 0.4.0. Version 0.7.0 is available at https://github.com/abcxyz/abc.\n\nTo disable notifications for this new version, set ABC_IGNORE_VERSIONS=\"0.7.0\". To disable all version notifications, set ABC_IGNORE_VERSIONS=\"all\".\n"
```

Would we rather just print to stderr? @drevell thoughts?

cmd/abc/abc.go
```diff
@@ -104,9 +107,56 @@ func setLogEnvVars() {
 	}
 }
 
+// checkForUpdates asynchronously checks for updates and prints results to
```
drevell, line 110: This comment seems out of sync with the code.

cmd/abc/abc.go
```diff
@@ -104,9 +107,56 @@ func setLogEnvVars() {
...
+// returned buffer.
+// A sync.WaitGroup is returned so main can wait for completion before exiting.
+func checkForUpdates(ctx context.Context) func() {
+	updatesCh := make(chan string, 1)
+
+	go func() {
```
drevell, line 116: Can abc-updater hide this complexity so that callers don't have to each implement this kind of concurrency?

It's really hard to get the concurrency details right (channels, goroutines, and selects), so better to just implement it once and have everyone call it.

I realize this contradicts the standard Go API advice of letting callers handle concurrency and keeping the library simple. But in this case I think it's worth it, because most people struggle to implement correct concurrent code of this type, and this library is for our internal use where we can dictate a consistent style.

I'm imagining something like:

```
func main() {
  report := abcupdater.Check(appID, version.Version)
  defer report()
}
```

cmd/abc/abc.go
```diff
@@ -104,9 +107,56 @@ func setLogEnvVars() {
...
+		defer close(updatesCh)
+
+		var b bytes.Buffer
+		if err := abcupdater.CheckAppVersion(ctx, &abcupdater.CheckVersionParams{
```
drevell, line 120: Why not have CheckAppVersion return string instead of writing to a writer? Could be simpler.

cmd/abc/abc.go
```diff
@@ -104,9 +107,56 @@ func setLogEnvVars() {
...
+			AppID: version.Name,
+			// Version: version.Version,
+			Version: "0.4.0", // intentionally older than cur version to test
```
drevell, line 123: I wonder what will happen if we're running an abc that was built from HEAD (so there's no semver version)? Should we skip the version check in that case?

cmd/abc/abc.go
```diff
@@ -104,9 +107,56 @@ func setLogEnvVars() {
...
+			Writer:  &b,
+		}); err != nil {
+			logging.FromContext(ctx).WarnContext(ctx, "failed to check for new versions",
+				"error", err)
+		}
+
+		select {
+		case updatesCh <- b.String():
```
drevell, line 131: I think it's impossible for this to block, so I think the select is unnecessary:

```
updatesCh <- b.String()
```

cmd/abc/abc.go
drevell: Well, we could be sneaky/hacky and just remove newlines and double quotes from the logged message. `ABC_IGNORE_VERSIONS=\"0.7.0\"` could be just `ABC_IGNORE_VERSIONS=0.7.0`, and so on. This assumes that the version numbers never contain whitespace or shell special characters.

stderr also seems fine to me.

cmd/abc/abc.go
pdewilde: Yeah, I had inherited that design, not quite sure what the reasoning was, my guess was to make it non-blocking so you could run it in a goroutine (since `go func()` doesn't capture the return of `func()`)

Now that its been refactored, string return would probably make more sense

cmd/abc/abc.go
pdewilde: Assuming `report` is a closure with the channel that will do the select, and `Check` would set up the channel and kick off the goroutine?

cmd/abc/abc.go
sethvargo: I'm not sure we should use the logger for human messages. The logger is for diagnostic messages. This feels like something we want to print to stderr using the cli.

cmd/abc/abc.go
drevell: Yeah, that's what I had in mind.

drevell: Also, the `abcupdater.Check()` function could be a wrapper around a simpler non-concurrent API like `func OtherCheck(appID, version string) (string, error)`. The caller could choose to deal with concurrency or not.

cmd/abc/abc.go
pdewilde: Added a check to skip

cmd/abc/abc.go
```diff
@@ -124,5 +126,19 @@ func realMain(ctx context.Context) error {
 	if runtime.GOOS == "windows" {
 		return fmt.Errorf("windows os is not supported in abc cli")
 	}
+
+	// Only check for updates if not building from HEAD.
```
drevell, line 130: Consider `building` -> `built` ?

drevell (approved):
I think the PR description needs an update

pdewilde:
> I think the PR description needs an update
updated

PR #417, opened 2024-02-07
Demonstrate how Seth would do this

sethvargo:


pdewilde:
Go lint will fail until abc updater is public. Merging to my branch, thanks for the good changes!

PR #419, opened 2024-02-08
chore: golden test improve stdout

capri-xiyue:
golden test improve stdout
fixed https://github.com/abcxyz/abc/pull/410#discussion_r1483490146 and https://github.com/abcxyz/abc/pull/410#discussion_r1483506932


templates/commands/goldentest/verify.go
```diff
@@ -163,6 +165,17 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) error {
 			}
 		}
 
+		stdOutDiff, err := getStdoutDiff(goldenStdoutFile, tempStdoutFile, dmp)
```
drevell, line 168: This function is getting awfully long (135 lines), could you factor out one or more of the `for` loops into separate function(s)? This could be a followup PR.

drevell, line 168: There's some inconsistent capitalization here. I think the `o` in stdout should be consistently lower-case.

templates/commands/goldentest/verify.go
```diff
@@ -235,3 +248,28 @@ func hasDiff(diffs []diffmatchpatch.Diff) bool {
 	}
 	return false
 }
+
+func getStdoutDiff(goldenStdoutPath, tempStdoutPath string, dmp *diffmatchpatch.DiffMatchPatch) ([]diffmatchpatch.Diff, error) {
+	goldenStdout, err := os.ReadFile(goldenStdoutPath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			goldenStdout = []byte("")
+		} else {
```
drevell, line 257: In Go, the usual style is to avoid unnecessary `else`. So this could be:

```
if err != nil {
  if !errors.Is(err, os.ErrNotExist) {
    return nil, fmt.Errorf(........)
  }
  goldenStdout = []byte("")
}
```

Same with `tempStdout` below.

drevell (changes requested):

templates/commands/goldentest/verify.go
capri-xiyue: done

templates/commands/goldentest/verify.go
capri-xiyue: done

templates/commands/goldentest/verify.go
capri-xiyue: Will do it in another PR.

drevell (approved):

PR #421, opened 2024-02-10
Rewrite temp dir management

drevell:
Handling of temp dirs was slightly incorrect and ugly in several ways:

 - In some places, errors were ignored when removing temp dirs
 - One temp dir was created but never used
 - There was a mishmash of styles, all of which were overly verbose and error prone
 - Some temp dirs were not cleaned up on the error path
 - os.TempDir() was being called unnecessarily, MkdirTemp() accepts `""` to default to using `/tmp`

So we renovate `TempTracker` and use it consistently.

### DIFF
--- a/templates/commands/goldentest/test_funcs.go
+++ b/templates/commands/goldentest/test_funcs.go
@@ -28,6 +28,7 @@ import (
 	"github.com/abcxyz/abc/templates/common"
 	"github.com/abcxyz/abc/templates/common/errs"
 	"github.com/abcxyz/abc/templates/common/render"
+	"github.com/abcxyz/abc/templates/common/tempdir"
 	"github.com/abcxyz/abc/templates/model"
 	"github.com/abcxyz/abc/templates/model/decode"
 	goldentest "github.com/abcxyz/abc/templates/model/goldentest/v1beta3"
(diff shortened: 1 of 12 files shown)


templates/commands/goldentest/test_funcs.go
```diff
@@ -134,9 +135,9 @@ func parseTestConfig(ctx context.Context, path string) (*goldentest.Test, error)
 	return out, nil
 }
 
-// renderTestCases render all test cases in a temporary directory.
-func renderTestCases(ctx context.Context, testCases []*TestCase, location string) (string, error) {
-	tempDir, err := os.MkdirTemp("", "abc-test-*")
+// renderTestCases render all test cases into a temporary directory.
+func renderTestCases(ctx context.Context, testCases []*TestCase, location string) (dir string, _ error) {
```
capri-xiyue, line 139: Why do we need to change the return signature here?

templates/common/tempdir/tempremover.go
```diff
@@ -0,0 +1,80 @@
...
+		return "", err //nolint:wrapcheck
+	}
+	t.Track(tempDir)
+	return tempDir, nil
+}
+
+// DeferMaybeRemoveAll should be called in a defer to clean up temp dirs, like this:
+//
+//	defer t.DeferMaybeRemoveAll(ctx, &rErr)
+func (t *DirTracker) DeferMaybeRemoveAll(ctx context.Context, outErr *error) {
```
capri-xiyue, line 67: Curious to learn whether it is a common practise to indicate the user that a error happened by modifying the parameter error instead of returning the error? Though I know currently using this mechanism, we don't have to have defer func in code everywhere.

templates/commands/goldentest/test_funcs.go
drevell: We don't really need to, it just makes it obvious what the meaning of the returned string is. It's purely cosmetic and functionally equivalent to the old code, just a bit more readable IMO.

templates/common/tempdir/tempremover.go
drevell: It's not common, but I think it makes sense here stylistically because it removes some error-prone boilerplate (`defer func() { rErr = errors.Join(rErr, tempRemover.RemoveAll(ctx)) }()`) in the caller, and there are several callers. Some people will likely hate this, but I love it.

capri-xiyue (approved):

PR #422, opened 2024-02-10
Factor out downloading from the `Render()` function

drevell:
Instead of providing a string template location to the Render() function and having Render() do the parsing, we provide an  implementation of the downloader interface. So now Render() is agnostic to the format of the template location string.

This lays more groundwork for the template upgrade feature. When we're upgrading a template, the template location string has a different format than the format provided by the user on the command line. When upgrading, we have a manifest file that provides the canonical template location (no `@version` suffix) and a location type. When a user is running a `render` command, we have the format with `@version`, which needs to be parsed differently.


templates/commands/describe/describe.go
```diff
@@ -97,18 +95,25 @@ func (c *Command) realRun(ctx context.Context, rp *runParams) (rErr error) {
...
 		Source:      c.flags.Source,
 		GitProtocol: c.flags.GitProtocol,
-		CWD:         cwd,
 	})
-	tempTracker.Track(templateDir)
 	if err != nil {
 		return err //nolint:wrapcheck
 	}
 
+	_, err = downloader.Download(ctx, cwd, templateDir)
```
capri-xiyue, line 112: nits: If with a short statement?

capri-xiyue (approved):

capri-xiyue (approved):

PR #426, opened 2024-02-12
chore: added react template back

capri-xiyue:
added react template back since #399 has been fixed.


drevell:
🎉🎉 thanks for following up with the details here

drevell (approved):

PR #427, opened 2024-02-13
Add parser and downloader factory for the upgrade case

drevell:
When we're doing template upgrades, we'll have a canonical location and a location type, and we'll need a `Downloader` to get the new template contents. This code does just that: `f(location_type, location) -> Downloader`


templates/common/templatesource/upgrade_test.go
```diff
@@ -0,0 +1,191 @@
...
+		{
+			name:        "remote_git_https_no_subdir",
+			location:    "github.com/abcxyz/abc",
+			locType:     "remote_git",
+			gitProtocol: "https",
+			wantDownloader: &remoteGitDownloader{
+				canonicalSource: "github.com/abcxyz/abc",
+				cloner:          &realCloner{},
+				remote:          "https://github.com/abcxyz/abc.git",
+				subdir:          "",
```
capri-xiyue, line 52: Nits: Do we need to explicitly set subdir as "". Same below

drevell: Fixed.

templates/common/templatesource/upgrade.go
```diff
@@ -0,0 +1,109 @@
...
+			`$`) // Anchor the end, must match the entire input
+)
+
+type upgradeDownloaderFactory func(_ context.Context, canonicalLocation, gitProtocol, destDir string) (Downloader, error)
+
+// ForUpgrade takes a location type and canonical location from a manifest file,
+// and returns a downloader that will download the latest version of that
+// template.
+func ForUpgrade(ctx context.Context, canonicalLocation, locType, gitProtocol, destDir string) (Downloader, error) {
+	factory, ok := upgradeDownloaderFactories[locType]
```
yolocs, line 55: Definitely fancier than a `swtich` :)

yolocs (approved):

templates/common/templatesource/upgrade.go
drevell: 🎩

Issue #428, opened 2024-02-13
Newer golden tests fail with old abc version


drevell:
EDIT: an old version of this issue incorrectly said that there was a problem with old golden tests running in new binaries. The problem is actually the other way around, with newer golden tests running with older binaries.

This is a blocker for the 0.6.0 release.

PR #429, opened 2024-02-13
chore: added UT to keep .gitkeep

capri-xiyue:
fixed part1 of #428 
add a unit test that makes sure a test without an .abc directory can still execute


capri-xiyue:
@drevell I verified in local and took a look at the code, we already have the logic that `don't assert the contents of the .abc directory, except for stdout which is a special case.`

sailorlqh (approved):

PR #430, opened 2024-02-14
Add the concept of unreleased api_versions

drevell:
Currently, it can be hard to release a new binary version because any unfinished features and api_versions will be exposed in that release. For example, we can't release a new binary version if api_version `v1beta4` is unfinished, because the new binary will put `api_version: v1beta4` in all the yaml files that it creates.

Instead, we want the ability to release a new binary version at any time, with unfinished features hidden behind an api_version that is still hidden from the public.

To do this, we mark each release struct with a boolean indicating whether it is released or not. api_versions that are under construction will not be used by official release binaries. When a release build of abc creates a yaml file, the header will use api_version=$last_officially_released_api_version, not the api_version that's under construction at the time of release.

### DIFF
--- a/templates/model/decode/decode.go
+++ b/templates/model/decode/decode.go
@@ -24,6 +24,7 @@ import (
 	"golang.org/x/exp/slices"
 	"gopkg.in/yaml.v3"
 
+	"github.com/abcxyz/abc/internal/version"
 	"github.com/abcxyz/abc/templates/model"
 	goldentestv1alpha1 "github.com/abcxyz/abc/templates/model/goldentest/v1alpha1"
 	goldentestv1beta3 "github.com/abcxyz/abc/templates/model/goldentest/v1beta3"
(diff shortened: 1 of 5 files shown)


sailorlqh (approved):

templates/model/decode/decode.go
```diff
@@ -228,3 +236,27 @@ func decodeFromVersionKind(filename, apiVersion, kind string, buf []byte) (model
...
+func LatestSupportedAPIVersion(isReleaseBuild bool) string {
+	// Release builds (like "I am the official release of version 1.2.3") will
+	// read and write only officially released, finalized api_versions. Other
+	// builds (e.g. CI builds, local dev builds, devs running "go test" on
+	// workstations) are more permissive and will read and write the most recent
+	// unreleased work-in-progress api_version.
+	if !isReleaseBuild {
+		return apiVersions[len(apiVersions)-1].apiVersion
+	}
+	for i := len(apiVersions) - 1; i > 0; i-- {
```
capri-xiyue, line 254: optional: https://pkg.go.dev/golang.org/x/exp/slices#Reverse?

capri-xiyue (approved):

templates/model/decode/decode.go
drevell: I considered it, but it's a little awkward here because we'd have to make a copy, then reverse the copy, then iterate.

PR #431, opened 2024-02-14
Create api_version v1beta5, same as v1beta4

drevell:
This is a simple hack for a very complex reason.

In #428, there's a problem where abc CLI version 0.5.0 *thinks* that it supports api_version v1beta4, but it really doesn't. It just supports the in-progress subset of v1beta4 that was implemented at the time that 0.5.0 was released. This causes a problem because when abc v0.5.0 sees a test.yaml that has api_version v1beta4, it *should* say "I don't understand this version, please upgrade your abc CLI" but what it will actually say is a weird test failure involving the `.gitkeep` file.

By bumping the soon-to-be-released api_version to v1beta5, that will be beyond what abc v0.5.0 thinks it can support, so it will correctly ask the user to upgrade.

The underlying root cause (abc CLI thinks it supports api_versions that it really doesn't) is fixed in #430 , but obviously we can't fix v0.5.0 which is already out in the wild.


capri-xiyue (approved):

capri-xiyue:
Not sure whether we want to do patch release for v0.5.0, for example having a v0.5.1 instead of bumping it to v1beta5?
One advantage of it could be we can have a quick fix before we release v0.6.0.

drevell:
> Not sure whether we want to do patch release for v0.5.0, for example having a v0.5.1 instead of bumping api version to v1beta5? One advantage of it could be we can have a quick fix before we release v0.6.0.

Discussed offline. For historical purposes: I think v0.6.0 is coming soon enough that there's not much benefit in creating a v0.5.1, and would incur some extra work in backporting a patch.

PR #432, opened 2024-02-14
feat: upgrade goldentest to v1beta4

capri-xiyue:
fixed part of #428 
1. Upgraded goldentest to v1beta4
2. Added UT to guard unreleased version.

### DIFF
--- a/templates/commands/goldentest/test_funcs_test.go
+++ b/templates/commands/goldentest/test_funcs_test.go
@@ -24,7 +24,7 @@ import (
 	"github.com/google/go-cmp/cmp/cmpopts"
 
 	"github.com/abcxyz/abc/templates/model"
-	goldentest "github.com/abcxyz/abc/templates/model/goldentest/v1beta3"
+	goldentest "github.com/abcxyz/abc/templates/model/goldentest/v1beta4"
 	abctestutil "github.com/abcxyz/abc/templates/testutil"
 	"github.com/abcxyz/pkg/testutil"
 )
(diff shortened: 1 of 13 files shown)


templates/commands/goldentest/test_funcs_test.go
```diff
@@ -24,7 +24,7 @@ import (
 	"github.com/google/go-cmp/cmp/cmpopts"
 
 	"github.com/abcxyz/abc/templates/model"
-	goldentest "github.com/abcxyz/abc/templates/model/goldentest/v1beta3"
+	goldentest "github.com/abcxyz/abc/templates/model/goldentest/v1beta4"
```
sailorlqh, line 27: is this `v1beta4` expected?

capri-xiyue: This is expected

templates/model/goldentest/v1beta3/upgrade.go
```diff
@@ -16,15 +16,21 @@ package goldentest
...
+	v1beta4 "github.com/abcxyz/abc/templates/model/goldentest/v1beta4"
 )
 
 // Upgrade implements model.ValidatorUpgrader.
 func (t *Test) Upgrade(ctx context.Context) (model.ValidatorUpgrader, error) {
-	logger := logging.FromContext(ctx).With("logger", "Upgrade")
-	logger.DebugContext(ctx, "finished upgrading goldentest model, this is the most recent version")
+	var out v1beta4.Test
+
+	if err := copier.Copy(&out, t); err != nil {
```
drevell, line 31: I think we need to do something like `SkipStdout` here to make sure that a new binary doesn't attempt to verify the stdout of old tests. Though probably you have this planned for a separate PR.

drevell (approved):

templates/model/goldentest/v1beta3/upgrade.go
capri-xiyue: I will do it in another PR. This PR focuses on gitkeep ut and goldentest upgrade

PR #435, opened 2024-02-14
Update dependencies in go.mod

drevell:
Command was `go get -u ./... ; go mod tidy`. The main thing we want is the 1.0.0 release of github.com/abcxyz/pkg, but the other upgrades look fine also.


sethvargo (approved):
I was going to wait until Go 1.22 landed in Homebrew and then do updates across the board, but you're ahead of me :)

PR #436, opened 2024-02-14
fix: enable stdout verification only when goldentest spec is >=v1beta4

capri-xiyue:
fixed part of #428 
1. Added logic that stdout verification is only enabled in goldentest spec v1beta4


templates/model/spec/features/features.go
```diff
@@ -14,15 +14,15 @@
 
 package features
 
-// Features controls which code paths are enabled by a given api_version.
+// Features controls which code paths are enabled in abc goldentest by a given api_version.
```
drevell, line 17: goldentest -> spec

templates/model/goldentest/v1beta4/goldentest.go
```diff
@@ -53,6 +54,8 @@ type Test struct {
 
 	Inputs      []*VarValue `yaml:"inputs,omitempty"`
 	BuiltinVars []*VarValue `yaml:"builtin_vars,omitempty"`
+	// Features configures which features to use depending on goldentest version.
```
drevell, line 57: Maybe say "API version" rather than just "version"?

drevell (approved):

drevell (approved):

PR #439, opened 2024-02-15
Dynamically compute version information

sethvargo:


lock14 (approved):

verbanicm (approved):

.goreleaser.yaml
```diff
@@ -33,9 +33,9 @@ builds:
     ldflags:
       - '-s'
       - '-w'
-      - '-X={{ .ModulePath }}/internal/version.Name=abc'
-      - '-X={{ .ModulePath }}/internal/version.Version={{ .Version }}'
-      - '-X={{ .ModulePath }}/internal/version.Commit={{ .Commit }}'
+      - '-X={{ .ModulePath }}/internal/version.name=abc'
+      - '-X={{ .ModulePath }}/internal/version.version={{ .Version }}'
```
capri-xiyue, line 37: Curious to learn why it should be `version.version` instead of `version.Version`, I saw other examples are `version.Version`, see https://github.com/hashicorp/packer-plugin-qemu/blob/1984bb1d2168c671886d40c8a87b341192db5944/.goreleaser.yml#L23. And the official example is goreleaser is `main.version={{.Version}}`, see https://goreleaser.com/customization/builds/#builds

.goreleaser.yaml
```diff
@@ -33,9 +33,9 @@ builds:
     ldflags:
       - '-s'
       - '-w'
-      - '-X={{ .ModulePath }}/internal/version.Name=abc'
-      - '-X={{ .ModulePath }}/internal/version.Version={{ .Version }}'
-      - '-X={{ .ModulePath }}/internal/version.Commit={{ .Commit }}'
+      - '-X={{ .ModulePath }}/internal/version.name=abc'
```
capri-xiyue, line 36: Curious to learn why we don't use the default flags in ` -s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}} -X main.builtBy=goreleaser` in go releaser, see https://goreleaser.com/customization/builds/, I thought with the default flag, it will just override the buildinfo.Version() in `internal/version/version.go`.

capri-xiyue: Never mind. I made an experiment. It won't override the buildinfo.Version.

PR #440, opened 2024-02-15
fix: fixed release build info

### LINKED ISSUE #438: IsReleaseBuild() returns incorrect info.
### TL;DR

Ideally,  `IsReleaseBuild` returns true if this binary was built by goreleaser as part of
// the official release process (as opposed to a user just running "go build",
// or running in a CI environment, or something else). But somehow when I install abc cli via `go install ./...`, or https://github.com/abcxyz/abc/blob/23068d1417191358257a9663cc3bf10dffcffe1b/cmd/abc/abc_test.go#L66, `IsReleaseBuild` also returns true

### Expected behavior

_No response_

### Observed behavior

_No response_

### Environment Details

```markdown
N/A
```

### Additional information

_No response_

capri-xiyue:
fixed #438

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -15,8 +15,9 @@
 package version
 
 import (
-	"github.com/abcxyz/pkg/buildinfo"
 	"golang.org/x/mod/semver"
+
+	"github.com/abcxyz/pkg/buildinfo"
 )
 
 var (
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -49,7 +49,11 @@ var (
 // or running in a CI environment, or something else).
 func IsReleaseBuild() bool {
 	// Binary from release build should follow semver format.
-	return semver.IsValid("v" + Version)
+	// When installing via `go install github.com/abcxyz/abc/cmd/abc@vXXX`,
+	// `Version` is `vXXX` for example `v0.1.0` with `v` prefix.
+	// When installing via downloading artifacts in GitHub directly,
+	// `Version` is `XXX` for example `0.1.0` without `v` prefix.
+	return semver.IsValid("v"+Version) || semver.IsValid(Version)
 }
 
 func valueOrFallback(val string, fn func() string) string {
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -49,7 +49,7 @@ var (
 // or running in a CI environment, or something else).
 func IsReleaseBuild() bool {
 	// Binary from release build should follow semver format.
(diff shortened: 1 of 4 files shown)


internal/version/version.go
```diff
@@ -44,11 +48,8 @@ var (
 // the official release process (as opposed to a user just running "go build",
 // or running in a CI environment, or something else).
 func IsReleaseBuild() bool {
-	// "source" is a sentinel value returned from buildinfo.Version() when the
-	// build metadata is not present. The build metadata is filled in with the
-	// "-X" flag during release builds, so if it's not present, this binary is
-	// not a release build.
-	return Version != "source"
+	// Binary from release build should follow semver format.
```
capri-xiyue, line 51: Previously logic doesn't work because for `go install ./...` and code in `main` package, the Version will be `(devel)` instead of "source"

internal/version/version.go
```diff
@@ -44,11 +48,8 @@ var (
...
+	return semver.IsValid(Version)
```
capri-xiyue, line 52: I checked the code, the Version here should starts with "v". When we released v0.0.6-rc, we'd better double check it.

internal/version/version.go
```diff
@@ -44,11 +48,8 @@ var (
 // the official release process (as opposed to a user just running "go build",
 // or running in a CI environment, or something else).
 func IsReleaseBuild() bool {
-	// "source" is a sentinel value returned from buildinfo.Version() when the
-	// build metadata is not present. The build metadata is filled in with the
-	// "-X" flag during release builds, so if it's not present, this binary is
-	// not a release build.
-	return Version != "source"
+	// Binary from release build should follow semver format.
+	return semver.IsValid("v" + Version)
```
capri-xiyue, line 52: I googled it. Looks like the Version should not have `v` prefix by default. Let me know if I'm wrong

capri-xiyue: But I saw some conflicts, when I run `abc templates -version` on v0.0.5, it shows `abc v0.5.0 (HEAD, darwin/amd64)` from `HumanVersion = Name + " " + Version + " (" + Commit + ", " + OSArch + ")"`, looks like `Version` has v prefix.

But for https://github.com/abcxyz/abc/blob/3d99e9bae57b99bc551904ab1b916ecd3c9c4118/.goreleaser.yaml#L50, and the 
Screenshot 2024-02-14 at 9 25 54 PM, the `Version` doesn't have v prefix

drevell (approved):

PR #444, opened 2024-02-15
fix: enable abc rename only in goldentest version >=v1beta4

### LINKED ISSUE #443: enable .abc_renamed only when goldentest spec is >=v1beta4
### TL;DR

 user 1 creates a test.yaml with api_version v1beta3, and runs golden-test record using an old version of abc. Then user 2 runs golden-test verify with a new version of abc, and there's failure because the new version of abc expects to see files with .abc_renamed, but the testdata directory was never renamed with .abc_renamed

### Expected behavior

_No response_

### Observed behavior

_No response_

### Environment Details

```markdown
N/A
```

### Additional information

_No response_

capri-xiyue:
part of #428 
fixed #443 
enable abc rename only in goldentest version >=v1beta4


drevell (approved):
Thanks for the very quick fix on this

PR #457, opened 2024-02-23
Support node protos

shankiyani:


drevell (approved):

t/proto_template/version.yaml
verbanicm, line 1: As discussed, drop the yaml and use a string in a `VERSION` file

shankiyani: ack will follow up

PR #467, opened 2024-03-04
Add the concept of "ultimate" destination to rendering logic

drevell:
Before now, the Render() function passed through the output directory to the Download() function, so the downloader could determine if the template source location was canonical (input and output are in the same git workspace). But now we need more info, because of the subleties of upgrading.

When doing `abc templates upgrade`, we'll use the Render() function as a building block in a way that's a little different than how it's used by `abc templates render`. The `abc templates upgrade` doesn't use the same directory for the render output as for the final upgrade destination, which means we need to now provide both directories to Render(). Hence the addition of the DestDirUltimate field.

### DIFF
--- a/templates/common/render/render_test.go
+++ b/templates/common/render/render_test.go
@@ -1159,8 +1159,8 @@ steps:
 			t.Parallel()
 
 			tempDir := t.TempDir()
-			dest := filepath.Join(tempDir, "dest")
-			abctestutil.WriteAllDefaultMode(t, dest, tc.existingDestContents)
+			outDir := filepath.Join(tempDir, "out_dir")
+			abctestutil.WriteAllDefaultMode(t, outDir, tc.existingDestContents)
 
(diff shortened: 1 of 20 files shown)


templates/common/render/render_test.go
```diff
@@ -1178,7 +1178,7 @@ steps:
 				Backups:             true,
 				BackupDir:           backupDir,
 				Clock:               clk,
-				DestDir:             dest,
+				OutDir:              dest,
```
capri-xiyue, line 1181: nits: Do we need to rename `dest`to `outputDir`

capri-xiyue (approved):

templates/common/render/render_test.go
drevell: Yeah, we should. Done.

capri-xiyue (approved):

PR #470, opened 2024-03-05
Upgrade to Go 1.22.1

drevell:
Some Go CVEs were just announced, and they're fixed in 1.22.1, so let's upgrade.

I dunno if our GitHub Actions will support 1.22.1 yet, but let's try.


drevell:
Looks like we can't use Go 1.22 yet because CodeQL doesn't support it (https://github.com/github/roadmap/issues/918). Converting this PR to "draft" state for now.

sethvargo:
@drevell mind if I take this over? I'm doing it in other repos and I know what needs to change to make CodeQL happy.

drevell:
> @drevell mind if I take this over? I'm doing it in other repos and I know what needs to change to make CodeQL happy.

That would be very welcome, thanks.

sethvargo:
@drevell should be good now. I did everything in a single commit so you can see the diff more easily.

drevell:
@sethvargo LGTM. I can't approve though because it's my PR :)

sethvargo (approved):

sethvargo:
There's definitely a fun bug where I can approve my own changes that I added to your branch, but... approved 🚢 🚀

PR #483, opened 2024-03-14
Make run_golden_tests.sh 10x faster

drevell:
Instead of running `go run` repeatedly, which takes a while even when there's nothing to rebuild, we just build the binary once and run it directly.


sethvargo (approved):
Genius

drevell:
> Genius

💥💥💥
💥🤯💥
💥💥💥

PR #485, opened 2024-03-14
Add individual file hashing function

drevell:
This is part of the template upgrades feature. I'm committing in pieces because the entire feature is too big for one PR.

Why are there no tests? I've chosen to test this at the level of the exported API. There are tests already written that will be sent out soon that cover all of the code paths in this PR.

### DIFF
--- a/templates/common/upgrade/hash.go
+++ b/templates/common/upgrade/hash.go
@@ -27,8 +27,8 @@ import (
 	"github.com/abcxyz/abc/templates/common"
 )
 
-// TODO(upgrade): delete this once the rest of the upgrade code exists. The
-// compiler is complaning about unused variables.
+// ExportToAvoidWarnings avoids compiler warnings complaning about unused
+// variables. TODO(upgrade): remove this when no longer necessary.
 var ExportToAvoidWarnings = []any{hashAndCompare, notComputed}
 
 // hashResult is the result from hashAndCompare().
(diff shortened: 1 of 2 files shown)


templates/common/upgrade/hash.go
```diff
@@ -0,0 +1,95 @@
...
+	tokens := strings.SplitN(wantHash, ":", 2)
+	if len(tokens) != 2 {
+		return "", fmt.Errorf("malformed hash, expected it to begin with hash name followed by colon: %q", wantHash)
+	}
+
+	var hasher hash.Hash
+	var wantHashUnmarshaled []byte
+	switch tokens[0] {
+	case "h1":
+		hasher = sha256.New()
```
sethvargo, line 68: From an extensibility perspective, it might be better if we let the user supply a hashing function (as a string) and have a default on `""`. We can have a `switch` on the string among common hashing functions (md5, sha128, sha256, sha384, sha512, etc).

drevell: I'm doing this for consistency with dirhash, which we already use [here](https://github.com/abcxyz/abc/blob/ca35eed9cc7b1fa23801a84b877d2738c32a58e8/templates/common/dirhash/dirhash.go#L40). This way, we can have "h1" mean "use sha256 for dirhashing and also for individual file hashes." I don't think we want to give the user any control here, we want to limit choice for the sake of simplicity.

Here's an example of a manifest:

```
# Generated by the "abc templates" command. Do not modify.
api_version: cli.abcxyz.dev/v1beta6
kind: Manifest
creation_time: 2023-12-08T23:59:02.000000013Z
modification_time: 2023-12-08T23:59:02.000000013Z
template_location: "github.com/foo/bar"
location_type: "remote_git"
template_version: ""
template_dirhash: h1:uh/nUYc3HpipWEon9kYOsvSrEadfu8Q9TdfBuHcnF3o=
inputs: []
output_hashes:
    - file: a.txt
      hash: h1:ZmFrZV9vdXRwdXRfaGFzaF8zMl9ieXRlc19zaGEyNTY=
```

I think the there's nice conceptual simplicity in using the same "hash framing" scheme for both dirhashes and file hashes.

sethvargo: Ah - so you don't plan to expose this as a template func?

sethvargo (approved):

templates/common/upgrade/hash.go
drevell: Right, yeah, this is something totally else. That would be a neat template func though.

PR #486, opened 2024-03-14
Add the SkipGlob option to LoadDir

drevell:
This uses the "functional options" pattern, and leaves open the possibility of future options flags to LoadDir().

This will be used in upcoming template upgrade tests.

### DIFF
--- a/templates/testutil/fs.go
+++ b/templates/testutil/fs.go
@@ -88,8 +88,10 @@ func WriteAllMode(tb testing.TB, root string, files map[string]ModeAndContents) 
 // LoadDirMode reads all the files recursively under "dir", returning their contents as a
 // map[filename]->contents. Returns nil if dir doesn't exist. Keys use slash separators, not
 // native.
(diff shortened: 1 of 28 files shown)


templates/testutil/fs.go
```diff
@@ -102,7 +104,18 @@ func LoadDirMode(tb testing.TB, dir string) map[string]ModeAndContents {
 		if err != nil {
 			return err
 		}
+		relPath, err := filepath.Rel(dir, path)
+		if err != nil {
+			tb.Fatal(err)
+		}
+		skip := skippable(tb, relPath, opt.skipGlobs)
 		if d.IsDir() {
+			if skip {
```
capri-xiyue, line 113: This part looks a little bit confusing to me? Can we `if skip` as the outlayer and in the inner layer, check whether `d.IsDir()`.

capri-xiyue (approved):

templates/testutil/fs.go
drevell: Yeah, that is better, thanks. Done.

PR #487, opened 2024-03-15
Implement part of the upgrade "merge" algorithm

drevell:
The "merge" phase of the upgrade operation is where we take (1) the output of the upgraded template and (2) the user's already-rendered template output directory, and intelligently incorporate the upgraded template files.

This is tricky because there are so many cases. What if the user edited a file that was output by a previous template version, and now the new template wants to delete that file? We don't want to clobber the user's edits, so we ask them to manually resolve the conflict. There are several tricky cases like that (see `decideMerge()`).

Exhaustive tests are coming in a future PR.


yolocs (approved):
Fancy!

PR #490, opened 2024-03-15
proto workflows should exit on failure

shankiyani:


t/proto_template/scripts/generate.sh
```diff
@@ -19,17 +20,14 @@ function pre_gen_go {
...
 }
 
 function pre_gen_node {
   echo "running pre_gen_node"
 
-  proto_version_key='proto_version:'
-  proto_version_file='version.yaml'
-
-  PROTO_VERSION=${PROTO_VERSION:-$(grep $proto_version_key $proto_version_file | cut -c $((${#proto_version_key} + 1))- | xargs)}  
+  PROTO_VERSION=${PROTO_VERSION:-$(grep 'proto_version:' 'version.yaml' | cut -d':' -f2 | xargs echo)}
```
shankiyani, line 30: this will use the env variable if its set or fallback to the one in the version file

t/proto_template/buf.gen.yaml
```diff
@@ -4,7 +4,7 @@ version: 'v1'
 managed:
   enabled: true
   go_package_prefix:
-    default: 'github.com/REPLACE_GITHUB_ORG_NAME/REPLACE_GITHUB_REPO_NAME/gen/go'
+    default: 'github.com/REPLACE_GITHUB_ORG_NAME/REPLACE_GITHUB_REPO_NAME'
```
shankiyani, line 7: This was a bug. We upload a go module _within_ the `gen/go` path. Therefore when importing protos as a consuming application, this suffix will not be a part of the path.

t/proto_template/.github/workflows/ci.yml
```diff
@@ -73,12 +73,8 @@ jobs:
       - name: 'Run Build'
         shell: 'bash'
         run: |
-          set +e
-          BUILD_CHECK=$(buf build)
-          if ! [[ -z "${BUILD_CHECK}" ]]; then
-            echo "::error::$BUILD_CHECK"
-            exit 1
-          fi
+          set -eEuo pipefail
```
verbanicm, line 76: Since you are using `shell: 'bash'` this shouldnt be needed in this workflow. The generate script you are calling however was creating a new shell which needed those set.

More info: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

t/proto_template/.github/workflows/ci.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: 'Capture proposed version'
         id: 'capture'
         run: |
-          echo "PROPOSED_PROTO_VERSION=$(yq eval '.proto_version' version.yaml)" >> $GITHUB_OUTPUT
+          echo "PROPOSED_PROTO_VERSION=$(yq eval '.proto_version' version.yaml)" >> ${GITHUB_OUTPUT}
```
verbanicm, line 37: nit: I still think yaml is useless here, it should just be a text file and you can cat the entire file instead of relying on yq...

t/proto_template/.github/workflows/ci.yml
verbanicm: Throughout....

t/proto_template/buf.gen.yaml
verbanicm: Easy fix

verbanicm (approved):

PR #499, opened 2024-03-19
feat: detect previous unmerged files and abort upgrade

drevell:
This is similar to how git won't let you proceed until you resolve merge conflicts. When an upgrade operation begins, we look for any `*.abcmerge_*` files that would indicate that there was a previous upgrade operation that still needs to be manually resolved. If we were to ignore this and upgrade anyway, the user would likely be incredibly confused, and that would be bad.

### DIFF
--- a/templates/common/upgrade/merge.go
+++ b/templates/common/upgrade/merge.go
@@ -251,11 +251,15 @@ func mergeAll(ctx context.Context, p *commitParams, dryRun bool) error {
 }
 
 const (
+	// These are appended to files that need manual merge conflict resolution.
 	suffixLocallyAdded                  = ".abcmerge_locally_added"
 	suffixLocallyEdited                 = ".abcmerge_locally_edited"
 	suffixFromNewTemplate               = ".abcmerge_from_new_template"
 	suffixFromNewTemplateLocallyDeleted = ".abcmerge_locally_deleted_vs_new_template_version"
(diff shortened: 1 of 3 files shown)


templates/common/upgrade/merge.go
```diff
@@ -251,11 +251,15 @@ func mergeAll(ctx context.Context, p *commitParams, dryRun bool) error {
...
 const (
+	// These are appended to files that need manual merge conflict resolution.
 	suffixLocallyAdded                  = ".abcmerge_locally_added"
 	suffixLocallyEdited                 = ".abcmerge_locally_edited"
 	suffixFromNewTemplate               = ".abcmerge_from_new_template"
 	suffixFromNewTemplateLocallyDeleted = ".abcmerge_locally_deleted_vs_new_template_version"
 	suffixWantToDelete                  = ".abcmerge_template_wants_to_delete"
+
+	// This is the beginning of all the above suffixes.
+	conflictSuffixBegins = ".abcmerge_"
```
capri-xiyue, line 262: nits: use such suffix for the above variables like `suffixLocallyAdded`, `suffixLocallyEdited` and so on?

capri-xiyue (approved):

templates/common/upgrade/merge.go
drevell: Hm I think I want to keep the name different, because that makes it more obvious that this isn't just another suffix.

PR #501, opened 2024-03-21
Return merge conflicts and explanations from Upgrade()

drevell:
The Upgrade() function now returns a `Result` struct, which explains what happened.

If the Upgrade was a noop because there was no new template version, then `Result.AlreadyUpToDate` is set.

If there were any merge conflicts, then the files that need to be resolved are in `Result.Conflicts`, along with explanations. All other non-conflicting files are in `Result.NonConflicts`.


templates/common/upgrade/upgrade.go
```diff
@@ -96,28 +111,86 @@ type Params struct {
...
+}
+
 // Upgrade takes a directory containing previously rendered template output and
 // updates it using the newest version of the template, which is pointed to by
 // the manifest file.
 //
 // Returns true if the upgrade occurred, or false if the upgrade was skipped
 // because we're already on the latest version of the template.
-func Upgrade(ctx context.Context, p *Params) (_ bool, rErr error) {
+func Upgrade(ctx context.Context, p *Params) (_ Result, rErr error) {
```
yolocs, line 178: Would it be easier if we return a pointer here?

templates/common/upgrade/upgrade.go
```diff
@@ -205,26 +277,38 @@ func Upgrade(ctx context.Context, p *Params) (_ bool, rErr error) {
...
-		if err := commit(ctx, p, dryRun); err != nil {
-			return err
+		var err error
+		actionsTaken, err = commit(ctx, p, dryRun)
+		if err != nil {
+			return nil, err
 		}
+		sort.Slice(actionsTaken, func(i, j int) bool {
+			return actionsTaken[i].Path < actionsTaken[j].Path
+		})
```
yolocs, line 308: Shouldn't this be outside of the for look?

yolocs (approved):

templates/common/upgrade/upgrade.go
drevell: Yeah, done.

templates/common/upgrade/upgrade.go
drevell: Yep, done.

PR #504, opened 2024-03-21
Fix output formatting in upgrade output re merge conflicts

drevell:
There weren't enough newlines, causing everything to be a mishmash


capri-xiyue (approved):

templates/commands/upgrade/upgrade.go
```diff
@@ -149,15 +149,16 @@ func (c *Command) Run(ctx context.Context, args []string) error {
 	//
 	// TODO(upgrade):
 	//  - suggest diff / meld / vim commands?
-	fmt.Fprint(c.Stdout(), mergeInstructions+"\n")
+	fmt.Fprint(c.Stdout(), mergeInstructions+"\n\n")
```
capri-xiyue, line 152: Is there a UT to check our stdout?

drevell: Not yet, but there's a TODO for that a couple of lines above this.

sethvargo (approved):

PR #510, opened 2024-03-26
fix: template_location was mistakenly made into a required field

drevell:
template_location is required for upgrading (so we add a new error case to verify that), but it should not be required when unmarshaling a manifest model. That unmarshaling validation has now been removed. It's valid for a template to be rendered from a non-canonical template location, in which case template_location will be empty string.

### DIFF
--- a/templates/common/upgrade/upgrade_test.go
+++ b/templates/common/upgrade/upgrade_test.go
@@ -24,11 +24,14 @@ import (
 	"github.com/benbjohnson/clock"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"gopkg.in/yaml.v3"
 
 	"github.com/abcxyz/abc/templates/common"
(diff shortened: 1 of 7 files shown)


templates/common/upgrade/upgrade_test.go
```diff
@@ -760,6 +763,51 @@ steps:
...
+			TemplateDirhash:  model.String{Val: "asdfasdfasdfasdfasdf"},
+		},
+	}
+
+	buf, err := yaml.Marshal(m)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	manifestFullPath := filepath.Join(templateDir, "/manifest.yaml")
```
sailorlqh, line 793: Is the `/` in `/manifest` necessary? I think filepath.Join() will add the `/`

sailorlqh (approved):

sailorlqh (approved):

PR #511, opened 2024-03-26
Add tests for the upgrade command (not the library)

drevell:
These tests make sure that the upgrade command is invoking the upgrade library correctly. It especially verifies that the merge conflicts are printed in readable way, and verifies that prompting for inputs is working.


templates/commands/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,443 @@
+// Copyright 2023 The Authors (see AUTHORS file)
```
capri-xiyue, line 1: nits: 2024

templates/testutil/prompt.go
```diff
@@ -104,7 +104,7 @@ func DialogTest(ctx context.Context, tb testing.TB, steps []DialogStep, cmd cli.
 	}()
 
 	select {
-	case <-time.After(time.Second): // time is arbitrary
+	case <-time.After(5 * time.Second): // time is arbitrary
```
capri-xiyue, line 107: Wondering should we make such time second value a parameter instead of a value that should satisfy all callers?

drevell: I think that's unnecessary. The callers don't need to be different, and that just adds complexity to using this feature, whose whole purpose is convenience. If we were exposing this as a public API outside of the abc project, I think this would be worth doing, but not as it currently is.

Also, this timeout only happens on test failures, not on successes, so it's not really wasting anyone's time.

templates/commands/upgrade/upgrade_test.go
drevell: Fixed, thanks.

PR #515, opened 2024-03-27
infer branch on checkout step for protos template ci workflow

shankiyani:


t/proto_template/scripts/generate.sh
```diff
@@ -61,6 +61,9 @@ function post_gen_node {
 }
 
 function main {
+  # delete the gen folder and recreate from scratch
+  rm -rf gen
```
shankiyani, line 65: Addresses an unrelated bug to ensure deleted protos are also deleted in the gen folder.

sethvargo (approved):

PR #525, opened 2024-04-04
Factor out a wrapper function around exec'ing diff

drevell:
The golden-test verify code uses /bin/diff, and I'm about to add another use case for template upgrades, so it makes sense to factor this out into a library function with a clean interface and good unit tests.

Also, factor out an Overwrite() function that was duplicated in a couple different test files.

GitHub Actions uses macos 12, which doesn't support `diff --color`, so we have a workaround to skip diff coloring when not supported.


templates/common/run/diff_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+func TestDiff(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name        string
+		dirContents map[string]string
+		color       bool
+		// file1Contents string
+		// file2Contents string
```
sailorlqh: Curious if these two will have some usecase in the future.

drevell: Just fixed :)

sailorlqh (approved):

PR #530, opened 2024-04-04
Include patches in manifest file when rendering

drevell:
In the case where a file is included with `from: destination`, which is used to modify existing files in place, we'll save a diff when overwriting the destination file when commiting the rendered template output. This allows us to reverse the change later, which is important for the template upgrade process.


templates/common/render/render_test.go
```diff
@@ -1287,11 +1347,13 @@ steps:
 				t.Errorf("scratch directory contents were not as expected (-got,+want): %s", diff)
 			}
 
-			gotDestContents := abctestutil.LoadDir(t, outDir)
+			gotDestContents := abctestutil.LoadDir(t, outDir, abctestutil.SkipGlob(".abc/manifest*"))
```
capri-xiyue, line 1350: nits: Add another comment to explain that manifest will be evaluated separately for the reason XXXX?

capri-xiyue (approved):

templates/common/render/render_test.go
drevell: Done.

PR #531, opened 2024-04-06
Redesign the "run" package interface

drevell:
Since we're accumulating more use cases for running external commands,
it's time to clean up this interface. I'm about to add some new code
that runs the "patch" command, which requires passing standard input,
which we haven't done before.

So it's time to add a nice "functional options" interface for the Run()
function that allows options to be mixed-and-matched.


templates/common/run/run.go
verbanicm, line 1: @sethvargo @drevell abc and guardian both shell out to external commands, should we centralize this into pkg?

drevell: https://github.com/abcxyz/guardian/blob/b01fa398bb7ca21a4c0402ee1be902af9a7d5a45/pkg/child/child.go#L56 does look kinda similar. I'd say yes we should factor out into pkg, but it's not a high priority. I don't feel strongly about it thoughj.

verbanicm: Yea, no rush, can be factored out later.

templates/common/ptr.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2024 The Authors (see AUTHORS file)
```
sethvargo, line 1: Minor, but https://github.com/abcxyz/pkg/blob/main/pointer/pointer.go

templates/common/run/run.go
sethvargo: I don't have strong opinions other than maybe it shouldn't be part of _cli_. I'm fine it it's in pkg, but it's sufficiently different from _cli_.

templates/common/run/run.go
```diff
@@ -28,72 +29,76 @@ import (
...
-	return stdout, stderr, err
-}
-
-// RunAllowNonzero is like Run, except that if the command has a nonzero exit
-// code, that doesn't cause an error to be returned.
-func RunAllowNonzero(ctx context.Context, args ...string) (stdout, stderr string, exitCode int, _ error) {
-	return run(ctx, true, args...)
+func Simple(ctx context.Context, args ...string) (stdout, stderr string, _ error) {
+	stdoutBuf := &bytes.Buffer{}
+	stderrBuf := &bytes.Buffer{}
```
sethvargo, line 43: Nit, it'd be more idiomatic to do:

```go
var stdoutBuf, stderrBuf bytes.Buffer
```

templates/common/ptr.go
drevell: Removed.

templates/common/run/run.go
```diff
@@ -28,72 +29,75 @@ import (
...
-func Run(ctx context.Context, args ...string) (stdout, stderr string, _ error) {
-	stdout, stderr, _, err := run(ctx, false, args...)
-	return stdout, stderr, err
-}
-
-// RunAllowNonzero is like Run, except that if the command has a nonzero exit
-// code, that doesn't cause an error to be returned.
-func RunAllowNonzero(ctx context.Context, args ...string) (stdout, stderr string, exitCode int, _ error) {
-	return run(ctx, true, args...)
+func Simple(ctx context.Context, args ...string) (stdout, stderr string, _ error) {
```
capri-xiyue, line 41: nits: Without reading the comments, it is difficult to know what `simple` does here by reading the function name. Wondering whether `RunWithStd` is better here?

capri-xiyue (approved):

templates/common/run/run.go
drevell: This name follows the guideline in https://google.github.io/styleguide/go/decisions#package-vs-exported-symbol-name to think about the package name when choosing the function name. Callers see this as "run.Simple", which I think is a good name.

capri-xiyue (approved):

templates/common/run/run.go
drevell: Filed https://github.com/abcxyz/pkg/issues/323

PR #532, opened 2024-04-11
feat: added record all & verify all

### LINKED ISSUE #311: Support running all golden tests for all templates under a directory

capri-xiyue:
fixed #311

### DIFF
--- a/templates/commands/goldentest/record.go
+++ b/templates/commands/goldentest/record.go
@@ -15,7 +15,7 @@
 // Package goldentest implements golden test related subcommands.
 package goldentest
 
-// This file implements the "templates golden-test recordTestCases" subcommand.
+// This file implements the "templates golden-test record" subcommand.
 
 import (
 	"context"
--- a/templates/commands/goldentest/record.go
+++ b/templates/commands/goldentest/record.go
@@ -72,7 +72,7 @@ func (c *RecordCommand) Run(ctx context.Context, args []string) (rErr error) {
 		return fmt.Errorf("failed to parse flags: %w", err)
 	}
 
-	templateLocations, err := parseTemplateLocations(c.flags.Location)
+	templateLocations, err := crawlTemplateLocations(c.flags.Location)
 	if err != nil {
 		return fmt.Errorf("failed to parse template locations")
 	}
@@ -88,7 +88,7 @@ func record(ctx context.Context, templateLocation string, testNames []string) (r
 	logger.InfoContext(ctx, "recording test for template location", "template_location", templateLocation)
 	testCases, err := parseTestCases(ctx, templateLocation, testNames)
 	if err != nil {
-		return fmt.Errorf("failed to parse golden test: %w", err)
+		return fmt.Errorf("failed to parse golden test for template location %v: %w", templateLocation, err)
 	}
 
 	rfs := &common.RealFS{}
@@ -109,6 +109,7 @@ func record(ctx context.Context, templateLocation string, testNames []string) (r
 
 	// Recursively copy files from tempDir to template golden test directory.
 	for _, tc := range testCases {
+		logger.InfoContext(ctx, "recording test for test name", "testname", tc.TestName)
 		testDir := filepath.Join(templateLocation, goldenTestDir, tc.TestName, testDataDir)
 		if err := os.RemoveAll(testDir); err != nil {
 			return fmt.Errorf("failed to clear test directory: %w", err)
--- a/templates/commands/goldentest/record.go
+++ b/templates/commands/goldentest/record.go
@@ -15,7 +15,7 @@
 // Package goldentest implements golden test related subcommands.
 package goldentest
 
-// This file implements the "templates golden-test record" subcommand.
+// This file implements the "templates golden-test recordTestCases" subcommand.
 
 import (
 	"context"
@@ -38,7 +38,7 @@ type RecordCommand struct {
 }
 
 func (c *RecordCommand) Desc() string {
-	return "record the template rendering result to golden tests " +
+	return "recordTestCases the template rendering result to golden tests " +
 		"(capture the anticipated outcome akin to expected output in unit test)"
 }
 
@@ -78,12 +78,12 @@ func (c *RecordCommand) Run(ctx context.Context, args []string) (rErr error) {
 	}
 	var merr error
 	for _, templateLocation := range templateLocations {
-		merr = errors.Join(merr, record(ctx, templateLocation, c.flags.TestNames))
+		merr = errors.Join(merr, recordTestCases(ctx, templateLocation, c.flags.TestNames))
 	}
 	return merr
 }
 
-func record(ctx context.Context, templateLocation string, testNames []string) (rErr error) {
+func recordTestCases(ctx context.Context, templateLocation string, testNames []string) (rErr error) {
 	logger := logging.FromContext(ctx)
 	logger.InfoContext(ctx, "recording test for template location", "template_location", templateLocation)
 	testCases, err := parseTestCases(ctx, templateLocation, testNames)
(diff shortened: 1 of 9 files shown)


templates/commands/goldentest/test_funcs.go
```diff
@@ -252,3 +253,24 @@ func renameGitDirsAndFiles(dir string) error {
...
 	return nil
 }
+
+func parseTemplateLocations(dir string) ([]string, error) {
+	var out []string
+	err := filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() {
```
drevell, line 263: Can we invert this conditional and save a level of nesting?

```
if !d.IsDir() {
    continue
}
```

templates/commands/goldentest/test_funcs.go
```diff
@@ -252,3 +253,24 @@ func renameGitDirsAndFiles(dir string) error {
...
+			if _, err := os.Stat(filepath.Join(path, "spec.yaml")); err == nil {
+				out = append(out, path)
+			}
```
drevell, line 266: Let's also exclude golden-test directories from being crawled. Some templates, like gcp-org-terraform template, have *output files* named spec.yaml. Therefore, these spec.yaml files will appear in golden test testdata directories.

So we can do something like `if d is a golden test testdata directory, then return filepath.SkipDir`.

templates/commands/goldentest/test_funcs.go
```diff
@@ -252,3 +253,24 @@ func renameGitDirsAndFiles(dir string) error {
 	}
 	return nil
 }
+
+func parseTemplateLocations(dir string) ([]string, error) {
```
drevell, line 257: I think we should remove "parse" from this function name. It's not really parsing, it's crawling.

templates/commands/goldentest/record_test.go
```diff
@@ -303,6 +303,148 @@ kind: 'GoldenTest'`,
...
+				"template2/spec.yaml":                               specYaml,
+				"template2/a.txt":                                   "template 2 file A content",
+				"template2/b.txt":                                   "template 2 file B content",
+				"template2/testdata/golden/test/test.yaml":          testYaml,
+				"template2/testdata/golden/test/data/.abc/.gitkeep": "",
+				"template2/testdata/golden/test/data/a.txt":         "template 2 file A content",
+				"template2/testdata/golden/test/data/b.txt":         "template 2 file B content",
+			},
+			wantErr: "failed to parse golden test",
+		},
```
drevell, line 412: Could we add a test case for a template that has an output file named spec.yaml in its testdata?

templates/commands/goldentest/test_funcs.go
```diff
@@ -79,12 +79,13 @@ func parseTestCases(ctx context.Context, location string, testNames []string) ([
 
 	if len(testNames) > 0 {
 		for _, testName := range testNames {
-			testCase, err := buildTestCase(ctx, testDir, testName)
-			if err != nil {
-				return nil, err
+			if _, err := os.Stat(filepath.Join(testDir, testName)); err == nil {
```
drevell, line 82: This code ignores the error returned from os.Stat(). We should be checking whether the error is os.ErrNotExist (which is fine) or some other random error (which isn't fine and should be returned).

templates/commands/goldentest/record.go
```diff
@@ -85,18 +99,17 @@ func (c *RecordCommand) Run(ctx context.Context, args []string) (rErr error) {
 	// Create a temporary directory to validate golden tests rendered with no
 	// error. If any test fails, no data should be written to file system
 	// for atomicity purpose.
-	tempDir, err := renderTestCases(ctx, testCases, c.flags.Location)
+	tempDir, err := renderTestCases(ctx, testCases, templateLocation)
 	if err != nil {
 		return fmt.Errorf("failed to render test cases: %w", err)
```
drevell, line 104: We need some way of identifying which test failed.

Here's what it looks like when spec.yaml parsing fails (notice how it doesn't say which file):

```
$ go run cmd/abc/abc.go templates golden-test record
failed to render test cases: failed to render golden tests: error reading template spec file: error parsing YAML file spec.yaml: at line 26 column 7: unknown field name "foo"; valid choices are [as from on_conflict paths skip]
exit status 1
```

templates/commands/goldentest/verify.go
```diff
@@ -86,7 +102,7 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) (rErr error) {
 	defer tempTracker.DeferMaybeRemoveAll(ctx, &rErr)
 
 	// Create a temporary directory to render golden tests
-	tempDir, err := renderTestCases(ctx, testCases, c.flags.Location)
+	tempDir, err := renderTestCases(ctx, testCases, templateLocation)
```
drevell, line 105: Same here, we need some way for the user to identify which tests failed to run. The error message doesn't say:

```
failed to render test cases: failed to render golden tests: error reading template spec file: error parsing YAML file spec.yaml: at line 26 column 7: unknown field name "foo"; valid choices are [as from on_conflict paths skip]
exit status 1
```

In general, I think it would be a good idea to pretend that you're a user and execute the CLI for the common failure cases.

drevell (changes requested):
Could you add docs to README.md for this feature?

capri-xiyue:
> Could you add docs to README.md for this feature?

Yep. I will do it in a separate PR.

templates/commands/goldentest/record.go
capri-xiyue: I changed the error msg here.
At the same time, I added the logging info to indicate which test case or which template location we are using.

templates/commands/goldentest/test_funcs.go
```diff
@@ -154,7 +161,10 @@ func renderTestCases(ctx context.Context, testCases []*TestCase, location string
 
 	var merr error
 	for _, tc := range testCases {
-		merr = errors.Join(merr, renderTestCase(ctx, location, tempDir, tc))
+		renderErr := renderTestCase(ctx, location, tempDir, tc)
+		if renderErr != nil {
+			merr = errors.Join(merr, fmt.Errorf("failed to render test case [%s] for template location [%s]: %w", tc.TestName, location, err))
```
drevell, line 166: `err` isn't the right variable to use in this message, should be `renderErr`.

You probably would have noticed this if you had tried running it on the command line, because it looks like this (note the `%!w(<nil>)`):

```
failed to render test cases: failed to render golden tests: failed to render test case [example_test] for template location [examples/templates/render/hello_jupiter]: %!w(<nil>)
```

Here's a more conventional way to write this, which avoids creating the renderErr variable and therefore avoids this kind of bug:

```
if err := renderTestCase(ctx, location, tempDir, tc); err != nil {
  merr  = errors.Join(merr, fmt.Errorf(..., err))
}
```

templates/commands/goldentest/record_test.go
```diff
@@ -127,6 +127,31 @@ kind: 'GoldenTest'`
 				"test/data/a.txt":         "file A content",
 			},
 		},
+		{
+			name: "failed_to_render_test_case",
+			filesContent: map[string]string{
+				"spec.yaml": `apiVersion: 'cli.abcxyz.dev/v1beta5'
+kind: 'Template'
+
+
```
drevell, line 136: This might be overly complex. Can we just have a spec.yaml whose contents are just `this is not a valid spec file and will fail to parse`?

templates/commands/goldentest/test_funcs.go
```diff
@@ -254,16 +264,20 @@ func renameGitDirsAndFiles(dir string) error {
...
 			return err
 		}
-		if d.IsDir() {
-			if _, err := os.Stat(filepath.Join(path, "spec.yaml")); err == nil {
-				out = append(out, path)
-			}
+		if !d.IsDir() {
+			return nil
+		}
+		if strings.HasSuffix(path, goldenTestDir) && d.IsDir() {
```
drevell, line 276: The call to `d.IsDir()` isn't necessary because the `if` three lines above already guarantees that this is a directory.

templates/commands/goldentest/verify_test.go
```diff
@@ -112,6 +112,30 @@ kind: 'GoldenTest'`
...
+  - desc: 'Print input values'
+    action: 'print'
+    params:
+      message: |
+        The variable values are:
+          my_input_without_default={{.my_input_without_default}}`,
+				"testdata/golden/test/test.yaml": testYaml,
+			},
+			wantErrs: []string{
+				"failed to render test case [test] for template location",
```
drevell, line 136: Perhaps we should check more of the error message. That might have caught the bug where the wrong error variable was printed.

drevell (changes requested):

templates/commands/goldentest/verify_test.go
capri-xiyue: I can't check more of the error message is because I won't know the value of template location in advance.

drevell: Ah, I see, thanks

templates/commands/goldentest/record_test.go
drevell: Got it, thanks

templates/commands/goldentest/verify.go
```diff
@@ -75,7 +77,21 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) (rErr error) {
 		return fmt.Errorf("failed to parse flags: %w", err)
 	}
 
-	testCases, err := parseTestCases(ctx, c.flags.Location, c.flags.TestNames)
+	templateLocations, err := crawlTemplateLocations(c.flags.Location)
+	if err != nil {
+		return fmt.Errorf("failed to parse template locations")
```
drevell, line 82: Two issues here:
 - `parse` -> `crawl`
 - The call to fmt.Errorf should include `err` with `%w`, so the actual error is shown to the user

templates/commands/goldentest/verify.go
```diff
@@ -124,7 +140,7 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) (rErr error) {
 		if err != nil {
 			if common.IsStatNotExistErr(err) {
 				return fmt.Errorf("no recorded test data in %q, "+
-					"please run `record` command to record the template rendering result to golden tests", goldenDataDir)
+					"please run `recordTestCases` command to recordTestCases the template rendering result to golden tests", goldenDataDir)
```
drevell, line 143: I think this might be a find-and-replace error. I don't think the `recordTestCases` command exists. So `recordTestCases` should be changed back to `record`.

templates/commands/goldentest/verify.go
```diff
@@ -204,18 +220,18 @@ func (c *VerifyCommand) Run(ctx context.Context, args []string) (rErr error) {
 
 		if outputMismatch {
 			failureText := red(fmt.Sprintf("golden test [%s] didn't match actual output, you might "+
-				"need to run 'record' command to capture it as the new expected output", tc.TestName))
+				"need to run 'record' command to capture it as the new expected output for template location %s", tc.TestName, templateLocation))
```
drevell, line 223: I think this would be a little nicer putting the template location at the beginning of the string next to the test name. Like:

`fmt.Sprintf("template [%s] golden test [%s] didn't match.......`

Here and elsewhere in this PR.

templates/commands/goldentest/verify_test.go
```diff
@@ -376,7 +388,7 @@ kind: 'GoldenTest'`,
 kind: 'GoldenTest'`,
 			},
 			wantErrs: []string{
-				"please run `record` command to record the template rendering result to golden tests",
+				"please run `recordTestCases` command to recordTestCases the template rendering result to golden tests",
```
drevell, line 391: Same here, `recordTestCases` -> `record`

drevell (changes requested):

drevell (approved):

PR #539, opened 2024-04-16
Remember source directory for "include from destination"

drevell:
In this edge case:

 - We're doing an upgrade
 - The template has an include action with `from: destination`
 - Some upgrade-reversal patches were successfully applied automatically by abc
 - Other upgrade-reversal patches had merge conflicts and were resolved manually by the user
 
... then there's some ambiguity about which directory is the "destination" directory (in the sense of "include from destination"). There are two directories that could be considered the be the source directory of the "include from destination" action. The first one is the real user-visible directory where the template is installed, and the other is the temp directory where the upgrade command saves its successful upgrade-reversal-patched files.

Therefore, whenever we do an "include from destination" action, we save the directory that the file came from. This will be needed later in upgrade.go when we merge the new template output with the installed template directory.


templates/common/render/render.go
```diff
@@ -389,19 +389,18 @@ type stepParams struct {
...
+	// location of the file in the scratch directory, and the map values are the
+	// directory from which the file was taken (sometimes this gets complicated
+	// there are multiple "destination" directories. We want to track these
+	// because they are treated specially in the final phase of rendering. When
+	// we commit the template output from the scratch directory into the
+	// destination directory, these paths are always allowed to be overwritten.
+	// For other files not in this list, it's an error to try to write to an
+	// existing file. This whole scheme supports the feature of modifying files
+	// that already exist in the destination.
+	includedFromDest map[string]string
```
yolocs, line 403: Just a question.

I haven't used this feature before. Say I have existing files in the dest folder that I'd like to keep, what will happen if I don't specify them via `includedFromDest`? Put it differently, when would I want to `includedFromDest`

yolocs (approved):

templates/common/render/render.go
drevell: As a template developer, the reason you'd want to use "include from destination" is because you want the template to modify a file in place. The canonical use case is `outputs.tf`: in the spec file, you'd "include-from-destination" the preexisting outputs.tf file, do a string replacement or append operation to add a new terraform output, and then the modified file would overwrite the old file.

PR #542, opened 2024-04-17
Clean up the places where we check existence of files

drevell:
1. Rename IsStatNotExistErr -> IsNotExistErr to reflect that it can be used for more than just Stat().
2. Use common.Exists() consistently, rather than calling `_, err := os.Stat()`.


templates/commands/goldentest/test_funcs.go
```diff
@@ -80,11 +80,9 @@ func parseTestCases(ctx context.Context, location string, testNames []string) ([
...
 		for _, testName := range testNames {
 			testPath := filepath.Join(testDir, testName)
-			_, err := os.Stat(testPath)
-			if err != nil {
-				if !common.IsStatNotExistErr(err) {
-					return nil, fmt.Errorf("failed to stat %q, %w", testPath, err)
-				}
+			if exists, err := common.Exists(testPath); err != nil {
+				return nil, fmt.Errorf("failed to stat %q, %w", testPath, err)
+			} else if !exists {
```
capri-xiyue, line 85: Can we avoid else if here?
Can we just use 

```
if !exists {
  continue
}

Same for others.

drevell: Sure, done.

capri-xiyue (approved):

PR #543, opened 2024-04-17
Improve the Result returned from the upgrade operation

drevell:
- Improve the Result returned by the Upgrade() library function by:
    - Changing the boolean status to an enum status because there are four distinct success/failure cases
    - Add a field for patch reversal conflicts, coming in a future PR
 - Add the ability for a command to return an exit code other than 0 or 1. This is important because when an upgrade fails, I want to indicate why in the status code. This allows automation to detect what went wrong without resorting to parsing error messages.
    - Use this feature to return 2 for merge conflicts and 3 for patch  reversal conflicts.


templates/commands/upgrade/upgrade.go
```diff
@@ -140,29 +140,32 @@ func (c *Command) Run(ctx context.Context, args []string) error {
...
+			if cf.IncomingTemplatePath != "" {
+				fmt.Fprintf(c.Stdout(), "incoming file: %s\n", cf.IncomingTemplatePath)
+			}
+			fmt.Fprintf(c.Stdout(), "--\n")
 		}
-		fmt.Fprintf(c.Stdout(), "--\n")
+		return &common.ExitCodeError{Code: 2}
+	case upgrade.PatchReversalConflict:
+		// TODO(upgrade): include instructions for resolving these conflicts
+		return &common.ExitCodeError{Code: 3}
```
yolocs, line 168: Fancy. Just checked even git doesn't differentiate these exit codes :)

I wonder if this should be specific error types. Exit codes are for CLI but not super library friendly. Non blocking though.

templates/common/render/action_include.go
```diff
@@ -138,17 +137,50 @@ func isGlob(matchedPaths []model.String, originalPath, matchedPath string) bool
...
+		// if sp.rp.IncludeFromDestExtraDir != "" {
+		// For complicated reasons related to upgrading, we sometimes add
+		// another include source directory that contains files after having
+		// had their include-from-destination patch reversed as part of the
+		// upgrade process.
+		//	fromDirs = append(fromDirs, sp.rp.IncludeFromDestExtraDir)
+		// }
+	}
+
+	anyMatches := false
```
yolocs, line 156: Feels like I've seen this change in the other PR you have.

yolocs (approved):

templates/commands/upgrade/upgrade.go
drevell: This file is just for the CLI command, not the library function. The library function for Upgrade() returns a status enum, which kinda does what you suggest.

templates/common/render/action_include.go
drevell: Oops, I didn't set the base branch on this PR properly, sorry. This should resolve itself when I merge the parent PR.

PR #549, opened 2024-04-18
Run GH actions on macos 14 instead of 12

drevell:
Why this is uncontroversial: macos 14 is about to become the default for the github runner label `macos-latest`, according to GitHub's blog. We're just changing it early.

Why do this now?
 - macos 12 is causing repeated problems by having weird versions of `diff` and `patch` that format their output differently.
 - we get test failures on macos 14 (the most common macos version) that don't show up in CI.

### DIFF
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -37,7 +37,7 @@ jobs:
       fail-fast: false
       matrix:
         runner:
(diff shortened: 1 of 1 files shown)


drevell:
I'm going to abandon this for now and try to use `git diff` and `git apply` instead to smooth over the differences between macos 12 and 14.

PR #552, opened 2024-04-18
Template rest_server: switch from chi to std library.

pdewilde:
Now that the std library supports routing via method: https://go.dev/blog/routing-enhancements the main reason for using Chi is no longer needed.

I removed the walk function as it's not necessary and not easily supported by the standard library.


yolocs (approved):
Nice!

PR #553, opened 2024-04-19
Use `git diff` instead of plain `diff`

drevell:
We've had a couple of issues now where the differences in git behavior between platforms causes problems. `diff` behaves differently on macos 12 vs macos 14 when coloring diff output, for example. By using `git diff` instead, we get more uniform behavior and fewer bugs.

This comes at a cost, though. The git diff command is missing some important flags that exist on regular diff, like:

 - missing -N flag to treat nonexistent files as empty
 - missing --label flag to override the filenames that are part of the diff output
 - inability to dereference symlinks

The absence of these flags forces us into a huge hack: to diff two files and output a diff where the files seem to be named p1 and p2, we have to *create a temp directory and copy the two files into that temp dir as a/p1 and b/p2*. This incurs filesystem overhead and is hacky as heck.

So given this is super hacky, why are we still doing it?
 - The temp dir is likely to be on tmpfs in RAM, so the FS operations should be fast. The non-tmpfs filesystem operations don't change (there's still one read of of the files to be diffed).
 - The complexity cost is worth having a uniform diff API across systems.

A similar problem exists for the `patch` command. If this PR is approved, I'll replace `patch` with `git patch` next. `patch` also has inconsistencies across macos versions.


templates/common/run/diff.go
```diff
@@ -35,101 +37,128 @@ import (
...
+// These metadata lines aren't presented in regular non-diff git output, and
+// they're just clutter. Also, the "index" line leaks metadata that we don't
+// want to leak.
+func trimMetadata(diffOutput string) string {
+	const linesToCheck = 2
+	splits := strings.SplitN(diffOutput, "\n", linesToCheck+1)
+	out := make([]string, 0, 1)
+	linesToIgnorePrefixes := []string{
+		"diff --git",
+		"index ",
```
sailorlqh, line 147: Is the extra space needed? Seems the below code still work without the extra. ` `

sailorlqh (approved):

templates/common/run/diff.go
drevell: It's not needed, but it is helpful. These lines look like `index foo..bar` and we want to match the largest possible prefix to try to ensure we're matching the right line.

PR #565, opened 2024-04-26
Add environment variables for some common flags

drevell:
It would be annoying for the user to have to add --git-protocol=ssh to every command line if they always use ssh.


drevell:
cc @bharathsabc

capri-xiyue (approved):

PR #566, opened 2024-04-26
Add the "ignore unknown inputs" feature

drevell:
This has two uses:

 - Most importantly, it's used by the upgrade logic. When we're
   upgrading a manifest that has some flag x, and the most recent
   version of the template removed that flag, then there would be an
   error unless the upgrader passed IgnoreUnknownInputs to the
   renderer.
 - It lets CLI users be less picky about which flags they copy-paste to
   which templates. I can just say `--input=service-account=foo`, and if
   the template doesn't want the service account, it can just ignore it,
   if --ignore-unknown-inputs was provided.

This flag defaults to false to provide greater safety and predictability by default.


templates/commands/render/flags.go
```diff
@@ -97,6 +101,13 @@ func (r *RenderFlags) Register(set *cli.FlagSet) {
 		Usage:   "If an output file already exists in the destination, overwrite it instead of failing.",
 	})
 
+	f.BoolVar(&cli.BoolVar{
+		Name:    "ignore-unknown-inputs",
+		Target:  &r.IgnoreUnknownInputs,
+		Default: false,
+		Usage:   "If a user-provided input name isn't recoognized by the template, ignore that input value instead of failing.",
```
yolocs, line 108:
```suggestion
		Usage:   "If a user-provided input name isn't recognized by the template, ignore that input value instead of failing.",
```

yolocs (approved):

PR #575, opened 2024-05-22
Add topological sort

drevell:
We'll use this in an upcoming PR for sorting the dependency graph of template installations, which matters when a template outputs another template.

Ideally we wouldn't write this ourselves, but the main graph library seems to be `gonum` and it just defines interfaces that we have to implement ourselves, at the cost of lots of boilerplate ([example](https://pkg.go.dev/gonum.org/v1/gonum@v0.15.0/graph#example-package-Implicit)). We can do the algorithm ourselves easily.

### DIFF
--- a/templates/common/graph/graph.go
+++ b/templates/common/graph/graph.go
@@ -15,135 +15,73 @@
 package graph
 
 import (
-	"cmp"
 	"fmt"
-	"slices"
-
-	"golang.org/x/exp/maps"
 )
 
-// A DAG is a directed graph. It is list of nodes, identified by their slice
-// index, each having 0 or more outgoing edges to other node indices. So
-// [][]int{{1}, {}} is a graph with two nodes, where node 0 has an outgoing edge
-// to node 1, and node 1 has no outgoing edges.
+// Graph represents a directed graph.
+type Graph[T comparable] struct {
+	edges map[T][]T
+}
+
+// NewGraph creates a new graph.
+func NewGraph[T comparable]() *Graph[T] {
+	return &Graph[T]{
+		edges: make(map[T][]T),
+	}
+}
+
+// AddEdge adds a directed edge from source to destination. This should be
+// interpreted as "$source depends on $destination," not "$source comes before
+// "$destination." In the topologically sorted output, $destination will come
+// before $source.
+func (g *Graph[T]) AddEdge(source, destination T) {
+	g.edges[source] = append(g.edges[source], destination)
+}
+
+// TopologicalSort performs a topological sort. For all edges a->b, the output
+// will have b before a.
 //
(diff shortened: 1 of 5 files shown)


templates/common/graph/graph.go
```diff
@@ -0,0 +1,149 @@
...
+
+// A DAG is a directed graph. It is list of nodes, identified by their slice
+// index, each having 0 or more outgoing edges to other node indices. So
+// [][]int{{1}, {}} is a graph with two nodes, where node 0 has an outgoing edge
+// to node 1, and node 1 has no outgoing edges.
+//
+// The edge directions are interpreted as "node index N depends on node indices
+// [X,Y,Z]". So the returned topological sort will place X, Y, and Z somewhere
+// before N.
+type DAG [][]int
```
sethvargo, line 33: I'm always hesitant returning a direct data structure like this. It might be better to have a struct with a private field instead.

```go
type DAG struct {
  fields [][]int
}
```

and then some public functions for accessing things.

Also, should this be `int64`?

templates/common/graph/graph.go
```diff
@@ -0,0 +1,149 @@
...
+//
+// If any outbound edge index in any of the slices is out of range
+// 0<=edgeIndex<len(d) then it will panic.
+//
+// Sorting an empty graph returns an empty list.
+//
+// Note for maintainers: why didn't we import a third-party lib for this?
+// They're all either sketchy, unmaintainted, or require implementing lots of
+// cumbersome interfaces. And this entire algorithm is tiny.
+func TopoSort(d DAG) ([]int, error) {
```
sethvargo, line 63: I'm a little confused about the decision to have a generic implementation and an int-based one? It seems a little odd. Can you explain?

sethvargo (approved):
This LGTM and leaving it here for now makes sense. However, we might want to look at upleveling this into `pkg` at some point.

templates/common/graph/graph.go
drevell: Re hiding the [][]int : sure, will do.

Re int64: the values in the slice are slice indices, which are ints, so no they shouldn't be int64.

templates/common/graph/graph.go
yolocs: Wouldn't it be equally straightforward if you directly implement with the generic?

drevell: Yeah, true. Maybe easier to read also. A generic sort using maps rather than slices would be bigger in memory and slower due to using generics and map lookups, but I suppose that doesn't really matter for this use case. I doubt anyone will install enough templates in a single repo for the difference in runtime between a int-slice and generic-map to matter (right?)

drevell: Rewrote.

yolocs (approved):

PR #580, opened 2024-05-24
Add "upgrade all" feature

drevell:
With this PR, users can just run "abc upgrade" and we'll find all the template installations under the current directory and upgrade all of them. The user no longer has to point to a manifest file, or even know what a manifest file is.

It's still missing some tests, which I'll add in a future PR. This one is already huge.

This is alpha quality at best. It probably has many critical bugs. But I've been working on this for months and it's time to get something committed.

### DIFF
--- a/templates/common/upgrade/upgrade.go
+++ b/templates/common/upgrade/upgrade.go
@@ -44,10 +44,12 @@ import (
 	"github.com/abcxyz/abc/templates/model/header"
 	manifest "github.com/abcxyz/abc/templates/model/manifest/v1alpha1"
 	"github.com/abcxyz/pkg/logging"
+	"github.com/abcxyz/pkg/sets"
 )
 
 // TODO(upgrade):
 //   - add "abort if conflict" feature
+//   - rethink overly complex manifest file name?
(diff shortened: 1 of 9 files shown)


templates/common/upgrade/upgrade.go
```diff
@@ -125,42 +129,90 @@ type Params struct {
...
-type ResultType string
+// ResultType is the outcome of attempting an upgrade. It may succeed, may be
+// unnecessary, or may have some sort of merge conflict.
+type ResultType int
 
 const (
 	// This is the value of Result.Type when the upgrade was vacuously
 	// successful because the template is already on the latest version.
-	AlreadyUpToDate ResultType = "already_up_to_date"
+	AlreadyUpToDate ResultType = iota
```
capri-xiyue, line 139: Curious, what's usgae of iota here?

capri-xiyue (approved):

templates/common/upgrade/upgrade.go
drevell: It's a Go feature, https://go.dev/ref/spec#Iota

PR #582, opened 2024-05-29
Add the --manifest-only flag to output just the manifest file

drevell:
This is useful when we're backfilling manifest files for already-rendered templates. We *do* want to re-render the template to get the manifest file and the hashes of all the files that were produced by the template, but we *don't* want to actually overwrite the files in the user's repo.


templates/commands/render/flags.go
```diff
@@ -115,7 +119,17 @@ func (r *RenderFlags) Register(set *cli.FlagSet) {
...
+	})
+
+	f.BoolVar(&cli.BoolVar{
+		Name:    "manifest-only",
+		Target:  &r.ManifestOnly,
+		Default: false,
+		EnvVar:  "ABC_MANIFEST_ONLY",
+		// TODO(upgrade): remove "(experimental)"
+		Usage: "(experimental) write only a manifest file and no other files; implicitly sets --manifest=true",
 	})
```
sethvargo, line 133: Do you want to hide these if they are experimental? `Hidden: true`?

templates/common/render/render.go
```diff
@@ -258,6 +266,7 @@ func RenderAlreadyDownloaded(ctx context.Context, dlMeta *templatesource.Downloa
 		rp:               p,
 		scope:            scope,
 		scratchDir:       scratchDir,
+		suppressPrint:    p.ManifestOnly, // if --manifest-only was given, then the user doesn't want printed output.
```
sethvargo, line 269: Nit: it feels like "silent mode" is a feature that's valuable on it's own. Should we couple it to `manifest-only` like this?

sethvargo (approved):

templates/commands/render/flags.go
drevell: I did consider that, and I think I want to leave it exposed. If someone wants to be future-proof and generate a manifest file to be ready for the upgrade feature once it exists, I'd like to allow that. And having allowed it, it should appear in the help text (with the `experimental` warning to avoid guaranteeing that the file format won't change).

templates/common/render/render.go
drevell: I could see that being valuable. But I think we can wait til somebody asks for it. I think it will be easy to uncouple in the future.

PR #584, opened 2024-05-30
fix: "upgrade all" should halt on merge conflict

drevell:
When a template fails to upgrade due to conflicts between the template and local edits, we should stop upgrading any further templates until the merge conflicts are resolved.

It used to work this way, but broke in a refactor by accident. I added a test to prevent regressing on this behavior again.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -33,6 +33,18 @@
                 "${workspaceFolder}/examples/templates/render/hello_jupiter",
                 "example_test"
             ]
-        }
+        },
+
+        {
+            "name": "Debug abc upgrade",
+            "type": "go",
+            "request": "launch",
+            "mode": "debug",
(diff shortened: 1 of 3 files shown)


.vscode/launch.json
```diff
@@ -33,6 +33,18 @@
...
+
+        {
+            "name": "Debug abc upgrade",
+            "type": "go",
+            "request": "launch",
+            "mode": "debug",
+            "program": "cmd/abc/abc.go",
+            "args": [
+                "upgrade",
+                "/usr/local/google/home/revell/git/abcxyz/infra-gcp"
```
capri-xiyue, line 46: This is not common for use?

drevell: It's a good starting point for other users to use, if they wish.

capri-xiyue (approved):

PR #585, opened 2024-06-13
Adjust the suggested commands to continue an upgrade

drevell:
When the upgrade operation is interrupted due to a merge conflict that needs to be manually resolved, we instruct the user how to continue the upgrade by running a command. Before this PR, we attempted to print the exact command to run to continue the upgrade. But I realized that this approach was flawed because we need to print all the flag values that might be needed (like --version). These flags may or may not have been provided by the user in their upgrade command.

That would have been too complex. So the new solution in this PR is to tell the user to re-run their previous upgrade command with a certain flag added, without attempting to provide the full set of flags.

### DIFF
--- a/templates/commands/upgrade/upgrade.go
+++ b/templates/commands/upgrade/upgrade.go
@@ -231,10 +231,9 @@ func summarizeResult(r *upgrade.ManifestResult, location string) string {
 		}
 		fmt.Fprintf(&out, `
 
-After manually resolving the merge conflict, run this command to continue
-upgrading other template installations that may exist:
-
-  abc upgrade %s`, location)
+After manually resolving the merge conflict, re-run the upgrade command to
+upgrade any other rendered templates in this location that may still need
+upgrading.`)
 
 		return out.String()
 	case upgrade.PatchReversalConflict:
@@ -259,10 +258,9 @@ upgrading other template installations that may exist:
 			resumeFrom = fmt.Sprintf(" --resume-from=%s", r.ManifestPath)
 		}
 		fmt.Fprintf(&out, `
-After manually applying the rejected hunks, run this command to continue:
-
-  abc upgrade %s --already-resolved=%s%s`,
-			shellescape.Quote(location), strings.Join(relPaths, ","), resumeFrom)
+After manually applying the rejected hunks, re-run the upgrade command with
+--already-resolved=%s%s`,
+			strings.Join(relPaths, ","), resumeFrom)
 		return out.String()
 	}
 	panic("unreachable") // the go lint exhaustive check prevents this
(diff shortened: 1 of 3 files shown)


pdewilde:
Couldn't you find flags the user provided by just using `os.Args`?

pdewilde (approved):

drevell:
> Couldn't you find flags the user provided by just using `os.Args`?

Yeah... we'd have to:
 - find and replace any previously-provided value of --resume-from and --already-resolved
 - shell-escape everything

That's not so hard, maybe I will do that... I'll consider.

pdewilde:
> > Couldn't you find flags the user provided by just using `os.Args`?
> 
> Yeah... we'd have to:
> 
> * find and replace any previously-provided value of --resume-from and --already-resolved
> * shell-escape everything
> 
> That's not so hard, maybe I will do that... I'll consider.

Though if someone used --prompt, just copy-pasting their previous command wouldn't work

drevell:
> > > Couldn't you find flags the user provided by just using `os.Args`?
> > 
> > 
> > Yeah... we'd have to:
> > 
> > * find and replace any previously-provided value of --resume-from and --already-resolved
> > * shell-escape everything
> > 
> > That's not so hard, maybe I will do that... I'll consider.
> 
> Though if someone used --prompt, just copy-pasting their previous command wouldn't work

On further thought, we'd also have to adjust --input and --input-file... it's getting complicated. Let's hold off for now.

pdewilde:
> > > > Couldn't you find flags the user provided by just using `os.Args`?
> > > 
> > > 
> > > Yeah... we'd have to:
> > > 
> > > * find and replace any previously-provided value of --resume-from and --already-resolved
> > > * shell-escape everything
> > > 
> > > That's not so hard, maybe I will do that... I'll consider.
> > 
> > 
> > Though if someone used --prompt, just copy-pasting their previous command wouldn't work
> 
> On further thought, we'd also have to adjust --input and --input-file... it's getting complicated. Let's hold off for now.

I wonder if would make more sense to searilize the process state in a temp file and have some special command to resume.

```
After manually applying the rejected hunks, run the command `abc resume jc8ji`
```

Where jc8ji would be some randomly generated ID for the temp file where state was stored.

drevell:
> > > > > Couldn't you find flags the user provided by just using `os.Args`?
> > > > 
> > > > 
> > > > Yeah... we'd have to:
> > > > 
> > > > * find and replace any previously-provided value of --resume-from and --already-resolved
> > > > * shell-escape everything
> > > > 
> > > > That's not so hard, maybe I will do that... I'll consider.
> > > 
> > > 
> > > Though if someone used --prompt, just copy-pasting their previous command wouldn't work
> > 
> > 
> > On further thought, we'd also have to adjust --input and --input-file... it's getting complicated. Let's hold off for now.
> 
> I wonder if would make more sense to searilize the process state in a temp file and have some special command to resume.
> 
> ```
> After manually applying the rejected hunks, run the command `abc resume jc8ji`
> ```
> 
> Where jc8ji would be some randomly generated ID for the temp file where state was stored.

Yeah, we'll probably get there someday, if this project continues going strong and users find themselves resolving a lot of conflicts. Like `git merge --continue`.

yolocs (approved):

PR #587, opened 2024-06-18
Clone entire repos instead of only the requested version

### LINKED ISSUE #324: Avoid extra git operation to look up "latest" template version
### TL;DR

When installing a template from github with version `@latest`, we do two git operations internally: (1) `git ls-remote` to get the tags and (2) `git clone --depth 1 --branch v1.2.3` to download the template contents. For people like me who use a security key for MFA, this requires two touches of the security key.

A possible alternative: just blindly clone the whole repo, including all branches and tags, rather than trying to optimally clone just the needed tag.

Benefits:
 - Fewer security key touches
 - Probably faster, there's only one connection setup and auth, rather than two.

Downsides:
 - Possibly more downloaded data

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

drevell:
Fixes #324 .

Before this, downloading the `latest` template version from a git repo involved two steps: querying the list of tags in the remote repo using `git ls-remote`, then running `git clone --branch` to get a particular template without downloading unnecessary history. This design was intended to avoid unnecessarily large downloads. However, there were some issues:

 - Opening a connection and authenticating twice is slow for SSH connections.
 - For users of physical security keys, two touches are needed.

So now we just do a plain `git clone` (no `--branch`), and then `git checkout` the requested version.

Also, we remove `findSymlinks()` and instead implement anti-symlink protection as a feature on `CopyRecursive` (this is the `ForbidSymlink` change).

A kevya and drevell pair-programming joint.

### DIFF
--- a/templates/common/fs.go
+++ b/templates/common/fs.go
@@ -124,7 +124,7 @@ type CopyParams struct {
 	// source, to allow customization of the copy operation on a per-file basis.
 	Visitor CopyVisitor
 
-	// Return [ErrSymlinkForbidden] if any symlinks are found when copying.
+	// Return [SymlinkForbiddenError] if any symlinks are found when copying.
 	//
 	// Why are symlinks risky in a template?
 	//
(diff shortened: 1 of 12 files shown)


templates/common/fs.go
```diff
@@ -124,6 +124,23 @@ type CopyParams struct {
...
+	//  - Users might be surprised if a template outputs a symlink, which
+	//    could point to any location outside the render destination dir.
+	//
+	// In general, symlinks could cause all kinds of surprising behavior. We're
+	// trying to provide a bit of protection for template users against
+	// malicious or overzealous template authors.
+	//
+	// We could be more permissive in the future and allow symlinks to exist in
+	// the repo as long as they're not involved with the template.
+	ForbidSymlinks bool
```
capri-xiyue, line 142: Will we default to allow symlinks?

capri-xiyue (approved):

templates/common/fs.go
drevell: Hmm good point, I'll invert `ForbidSymlinks` to `AllowSymlinks` so the default is more safe.

drevell: I ended up just hardcoding the symlink check. I don't think we ever want to allow symlinks, at least for now. We can make it configurable later, if needed.

PR #590, opened 2024-06-21
Fix golden tests broken by dependabot

drevell:
Recovers from #588 being merged prematurely. It had failing tests.

Separately, I'll work on updating the github config to require passing tests.


pdewilde (approved):
It's charitable to say they were broken by dependabot :stuck_out_tongue:

PR #592, opened 2024-06-24
Rename upgrade_track manifest field to upgrade_channel

drevell:
Thanks to Brad for pointing out the conventional terminology for this.


templates/common/render/manifest.go
```diff
@@ -208,7 +208,7 @@ func buildManifest(p *writeManifestParams) (*manifest.WithHeader, error) {
 			LocationType:     model.String{Val: locType},                  // may be empty string if location isn't canonical
 			TemplateDirhash:  model.String{Val: templateDirhash},
 			TemplateVersion:  model.String{Val: p.dlMeta.Version},
-			UpgradeTrack:     model.String{Val: p.dlMeta.UpgradeTrack},
+			UpgradeChannel:   model.String{Val: p.dlMeta.UpgradeTrack},
```
capri-xiyue, line 211: nits: Do you need to update `p.dlMeta.UpgradeTrack` to `p.dlMeta.UpgradeChannel`?

capri-xiyue (approved):

templates/common/render/manifest.go
drevell: Oops, yes, thanks for catching that.

PR #596, opened 2024-06-27
Finish implementing --resume-from/--already-resolved

drevell:
This code runs when recovering from merge conflicts. After a merge conflict is resolved, the user runs `abc upgrade <directory> --resume-from=some-manifest.yaml --already-resolved=<list-of-files>`. In the case where there are multiple template installations to be upgraded, the `--already-resolved` flag should only apply to the first one, and later upgrades should be attempted from scratch.

The implementation is just one line, and the rest is just test updates and reorganization.

### DIFF
--- a/templates/common/upgrade/upgrade_test.go
+++ b/templates/common/upgrade/upgrade_test.go
@@ -1393,13 +1393,12 @@ func TestPatchReversalManualResolution(t *testing.T) {
 	if err != nil {
 		t.Fatalf("time.LoadLocation(): %v", err)
 	}
-	beforeUpgradeTime := time.Date(2024, 3, 1, 4, 5, 6, 7, loc)
+	render1Time := time.Date(2024, 3, 1, 4, 5, 6, 7, loc)
 
 	tempBase := t.TempDir()
 	// Make tempBase into a valid git repo.
 	abctestutil.WriteAll(t, tempBase, abctestutil.WithGitRepoAt("", nil))
 
-	destDir := filepath.Join(tempBase, "dest_dir")
 	templateDir := filepath.Join(tempBase, "template_dir")
 
 	origTemplateDirContents := map[string]string{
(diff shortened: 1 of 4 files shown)


templates/common/upgrade/upgrade_test.go
```diff
@@ -1393,13 +1393,12 @@ func TestPatchReversalManualResolution(t *testing.T) {
 	if err != nil {
 		t.Fatalf("time.LoadLocation(): %v", err)
 	}
-	beforeUpgradeTime := time.Date(2024, 3, 1, 4, 5, 6, 7, loc)
+	render1Time := time.Date(2024, 3, 1, 4, 5, 6, 7, loc)
```
capri-xiyue, line 1396: nits: `renderTime1` instead of `render1Time`? For the consistency of `destDir1` and `destDir2`?

templates/common/upgrade/upgrade_test.go
```diff
@@ -1427,17 +1426,25 @@ steps:
...
+
+	abctestutil.WriteAll(t, destDir1, origDestContents)
+	abctestutil.WriteAll(t, destDir2, origDestContents)
 
 	ctx := context.Background()
 	clk := clock.NewMock()
-	clk.Set(beforeUpgradeTime)
-	renderResult := mustRender(t, ctx, clk, nil, tempBase, templateDir, destDir)
+	clk.Set(render1Time)
+	render1Result := mustRender(t, ctx, clk, nil, tempBase, templateDir, destDir1)
```
capri-xiyue, line 1442: renderResult1?

capri-xiyue (approved):

templates/common/upgrade/upgrade_test.go
drevell: Done

templates/common/upgrade/upgrade_test.go
drevell: Done

PR #601, opened 2024-07-03
Add the --upgrade-channel flag

drevell:
This lets the user specify what branch they want to track for upgrades. The default is still the old behavior, which is to make an educated guess (if the user installed from `mybranch` then track that, otherwise upgrade from the latest semver release).

### DIFF
--- a/templates/commands/render/flags.go
+++ b/templates/commands/render/flags.go
@@ -78,6 +78,10 @@ type RenderFlags struct {
 	// Whether to *only* create a manifest file without outputting any other
 	// files from the template.
 	ManifestOnly bool
+
+	// Overrides the `upgrade_channel` field in the output manifest. Can be
+	// either a branch name or the special string "latest".
+	UpgradeChannel string
 }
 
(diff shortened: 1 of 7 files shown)


templates/commands/render/flags.go
```diff
@@ -123,6 +127,14 @@ func (r *RenderFlags) Register(set *cli.FlagSet) {
 		Usage: "(experimental) write a manifest file containing metadata that will allow future template upgrades.",
 	})
 
+	f.StringVar(&cli.StringVar{
+		Name:    "upgrade-channel",
+		Target:  &r.UpgradeChannel,
+		Default: "",
```
capri-xiyue, line 133: Curious What does default "" means in `upgrade-channel`?

capri-xiyue (approved):

templates/commands/render/flags.go
drevell: That's explain in the usage text below: `The default is to upgrade from the branch that the template was originally rendered from if rendered from a branch, or in any other case to use the value "latest" to upgrade to the latest release tag by semver order.`

PR #603, opened 2024-07-09
Test prompting during upgrade when a new input is added with a default

drevell:
This tests an edge case where:

 1. A template is installed that has no inputs.
 2. The template is updated to add an input that has a default value
 3. We run an upgrade
 4. The user should be prompted for a value for the new input; the default value should not silently be used.

I thought I saw this last week in my manual testing, but the test passes without changing the logic, so I may have been mistaken.

This change involves some refactoring and test util updates because this is the first time that we've tested prompting logic at the library layer (`common/*`), rather than the Command layer (`commands/*`) which wraps the library.

### DIFF
--- a/templates/testutil/prompt/prompt.go
+++ b/templates/testutil/prompt/prompt.go
@@ -27,9 +27,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
+
 	"github.com/abcxyz/abc/templates/common/input"
 	"github.com/abcxyz/pkg/cli"
-	"github.com/google/go-cmp/cmp"
 )
 
 const (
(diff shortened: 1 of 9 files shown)


templates/testutil/prompt/prompt.go
```diff
@@ -131,6 +136,71 @@ func DialogTest(ctx context.Context, tb testing.TB, steps []DialogStep, cmd cli.
...
+type FakePrompter struct {
+	steps     []DialogStep // Immutable.
+	stepIndex int          // Mutable. Incremented after each back-and-forth
+}
+
+func (f *FakePrompter) Prompt(ctx context.Context, msg string, args ...any) (string, error) {
+	// Note for maintainers: we call panic() rather than t.Fatal() in this
+	// function because panic() will stop all goroutines while t.Fatal() won't.
+	// If the other goroutines are left running, they will produce misleading
+	// failure messages that will obscure the true cause of the problem.
```
yolocs, line 148: Maybe worth clarifying what the _other_ goroutines are? Eyeballing this PR, we only spawn one goroutine in `DialogTestFunc` per each subtest. I wonder if `t.Fatal` + return an error would just work?

Using `panic` would terminate other tests, wouldn't it?

yolocs (approved):

templates/testutil/prompt/prompt.go
drevell: This message is written from the point of view of the `prompt` library; it doesn't really know who's calling or what goroutines may be running, but it's safe to assume that one or more other goroutines are doing a test and are will soon do assertions that we hope to avoid (like timing out).

Yeah, `panic` will disrupt other tests, which kinda sucks. But I think it's worth it to achieve clarity of failure messages in the rare event of a test failure.

PR #608, opened 2024-07-15
Speed up `golden-test verify` by 2.5x using concurrency

drevell:
The `golden-test verify` subcommand is becoming weirdly slow (1.9 seconds when run on the abc repo). Profiling shows that the cause of slowness is mostly dealing with temp files, especially in our `diff` implementation which creates a copy of the files being diffed.

This PR just adds some coarse-grained concurrency at the top level of the verify command. There's still room for improvement, but this is a decent start.

Ways to improve performance further in the future:
 - Fix our diff algorithm. To smooth over the differences between Linux diff and MacOS/BSD diff, we have a gross hack where we invoke `git diff`. To do that we have to copy the two files being diffed into a temp directory before running git diff. This wastes a LOT of time. There seems to be no good unified diff library for Go, which is unfortunate.
 - More concurrency:
   - Concurrently render the test cases, not just the the diffing
   - Concurrently run diffs for the test cases within a single test. Currently all the test cases for a single template run in teh same worker thread.
   - parallelize temp dir cleanup inside tempdir.DirTracker. This might suck if we have to reimplement os.RemoveAll() 😟 .


templates/commands/goldentest/verify.go
```diff
@@ -124,40 +148,50 @@ func verify(ctx context.Context, templateLocation string, testNames []string, st
...
-		}
-		if err := runTestCase(ctx, p, tc); err != nil {
-			result := red(fmt.Sprintf("[x] template location [%s] golden test [%s] fails", templateLocation, tc.TestName))
-			tcErr := fmt.Errorf("%s:\n %w", result, err)
-			merr = errors.Join(merr, tcErr)
-			resultReport.WriteString(result)
-		} else {
-			resultReport.WriteString(green(fmt.Sprintf("[✓] template location [%s] golden test [%s] succeeds", templateLocation, tc.TestName)))
-		}
+			tempDir, err := tempTracker.MkdirTempTracked("", "verify-testcase-"+tc.TestName+"-")
```
sailorlqh, line 158: just curious why we need a `-` at the end?

sailorlqh (approved):

templates/commands/goldentest/verify.go
drevell: It will be appended with a random cookie, like `verify-testcase-foo-123123`. It's nicer to have the dash to keep the parts separate.

https://pkg.go.dev/os#MkdirTemp

lock14 (approved):

templates/commands/goldentest/verify.go
sailorlqh: Thank you! Learned another new thing from you :)

PR #614, opened 2024-07-22
Add the --even-if-up-to-date flag to `upgrade`

drevell:
This flag forces the upgrade command to proceed even if the the installed template version is already the latest version. This is useful for:

 - Forcing the manifest to be rewritten, perhaps to change the --template-location or --upgrade-channel values
 - Re-rendering a template with different inputs using `--input=foo=bar`

### DIFF
--- a/templates/commands/upgrade/flags.go
+++ b/templates/commands/upgrade/flags.go
@@ -51,6 +51,12 @@ type Flags struct {
 	// See common/flags.DebugStepDiffs().
 	DebugStepDiffs bool
 
+	// Continue upgrading even if the dirhash matches between the
+	// already-installed template version and the to-be-installed template
+	// version. This is useful to for the manifest to be rewritten with a new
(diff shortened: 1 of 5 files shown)


templates/commands/upgrade/flags.go
```diff
@@ -101,6 +107,11 @@ func (f *Flags) Register(set *cli.FlagSet) {
 		Target:  &f.ResumeFrom,
 		Usage:   "begin or resume the upgrade starting at this manifest file",
 	})
+	u.BoolVar(&cli.BoolVar{
+		Name:   "even-if-up-to-date",
```
yolocs, line 111: The name is a bit hard to read. Would something like `--force-upgrade` or `--force-refresh` better?

drevell: `--continue-if-current` is :chefkiss: . Thanks! Changed.

verbanicm (approved):

PR #631, opened 2024-08-09
fix merge conflict instructions

drevell:
This logic was changed a few weeks ago but the instructions werent updated accordingly.

### DIFF
--- a/templates/commands/upgrade/upgrade.go
+++ b/templates/commands/upgrade/upgrade.go
@@ -78,11 +78,15 @@ Background on conflict types:
 
  - editEditConflict: you made some local edits to this file that was installed
    by the template, which conflicts with the new version of the template which
-   wants to edit the file. Both versions of the file are left in your output
-   directory, named "yourfile.abcmerge_locally_edited" and
-   "yourfile.abcmerge_from_new_template". Please resolve the conflict by either
-   (1) renaming one of the files to "yourfile" and deleting the other, or (2)
-   merging the two files into "yourfile".
+   wants to edit the file. Your locally edited file is unchanged, and the
+   incoming file from the template has the additional extension 
+   ".abcmerge_from_new_template". Please resolve the conflict by either
+     - selectively incorporating some of the changes from the 
+       .abcmerge_from_new_template file into your local file
+	 - rejecting this incoming change by removing the 
+	   .abcmerge_from_new_template file
+	 - overwriting your locally edit file with the .abcmerge_from_new_template
+	   incoming file. 
 
  - editDeleteConflict: you made an edit to this file that was installed by the
    template, which conflicts with the new version of the template, which wants


templates/commands/upgrade/upgrade.go
```diff
@@ -78,11 +78,15 @@ Background on conflict types:
 
  - editEditConflict: you made some local edits to this file that was installed
    by the template, which conflicts with the new version of the template which
-   wants to edit the file. Both versions of the file are left in your output
-   directory, named "yourfile.abcmerge_locally_edited" and
-   "yourfile.abcmerge_from_new_template". Please resolve the conflict by either
-   (1) renaming one of the files to "yourfile" and deleting the other, or (2)
-   merging the two files into "yourfile".
+   wants to edit the file. Your locally edited file is unchanged, and the
```
verbanicm, line 81: The language in this entire blob of text is a little wordy. I am suggesting a change for the entire `mergeInstructions` var.


```go
	mergeInstructions = `
Some manual conflict resolution is required because of a conflict between your
local edits and the new version of the template. Please look at all files ending
in .abcmerge_* and either edit, delete, or rename them to reflect your decision.

Background on conflict types:

 - editEditConflict: Local file edits conflict with an updated template file. 
   Your locally edited file is unchanged and the incoming file from the template 
   has the additional extension ".abcmerge_from_new_template".
   
   Please resolve the conflict by:

      (1) selectively incorporating some of the changes from the
          .abcmerge_from_new_template file into your local file
          
	    (2) rejecting this incoming change by removing the
          .abcmerge_from_new_template file

	    (3) overwriting your locally edit file with the .abcmerge_from_new_template
  	      incoming file.

 - editDeleteConflict: Local file exists but was deleted from the template. Your 
   version has been renamed to "yourfile.abcmerge_template_wants_to_delete". 
   Please resolve the conflict by renaming it back to "yourfile" or deleting it.

 - deleteEditConflict: Local file has been deleted, but exists in the template. 
   The new version from the template is named "yourfile.abcmerge_locally_deleted_vs_new_template_version".
   Please resolve the conflict by renaming it to "yourfile" or deleting it.

 - addAddConflict: Local file created with same name as a new template file. Your 
   version of the file has been renamed to "yourfile.abcmerge_locally_added", 
   and the version of the template is named "yourfile.abcmerge_from_new_template". 

   Please resolve the conflict by either: 

      (1) renaming one of these files to "yourfile" and deleting the other

      (2) merging the two files into "yourfile".
`
```

drevell: I think I have to disagree with these edits. I think the preexisting version is more clear. If someone is reading this, they don't know what's going on, so a sentence like "Local file edits conflict with an updated template file" may not be helpful enough.

Concretely: in "Local file edits conflict with an updated template file":
 - Grammatical ambiguity: it's not clear with "edits" or "conflict" which one is the noun and which one is a verb. Is "edits" a present-tense verb? Or is it a plural noun? If you don't understand abc's data model it might not be obvious.
 - Assumed domain knowledge: the user may not understand what "updated template file" means. I can imagine them asking "what's a template file? Is that a file that is a template?" They may not know that we mean "a file which was installed by the template."

verbanicm: I have no strong opinions.

verbanicm (approved):

PR #633, opened 2024-08-09
Reuse already-loaded manifest instead of re-reading

drevell:
Before this, we were unnecessarily re-reading the manifest from the filesystem. It's already loaded in UpgradeAll(), so now we just pass it to upgrade() instead of passing the path and re-parsing the manifest.

### DIFF
--- a/templates/common/upgrade/upgradeall.go
+++ b/templates/common/upgrade/upgradeall.go
@@ -88,7 +88,7 @@ func UpgradeAll(ctx context.Context, p *Params) *Result {
 		return &Result{Err: err}
 	}
 
-	_, sorted, depGraph, err := manifestsToUpgrade(ctx, p)
+	manifests, sorted, depGraph, err := manifestsToUpgrade(ctx, p)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


templates/common/upgrade/upgradeall.go
```diff
@@ -104,7 +104,8 @@ func UpgradeAll(ctx context.Context, p *Params) *Result {
 		}
 		logger.InfoContext(ctx, "beginning upgrade of manifest",
 			"manifest", absManifestPath)
-		result, err := upgrade(ctx, p, absManifestPath)
+		manifest := manifests[manifestPath]
```
verbanicm, line 107: should we check for non-existence here or does that happen later?

verbanicm (approved):

templates/common/upgrade/upgradeall.go
drevell: I just added a comment to make this more clear. We're guaranteed that `manifests` and `sorted` are one-to-one, so this lookup is guaranteed to succeed.

PR #651, opened 2024-10-14
Metrics fixes

pdewilde:
Copied from changes in guaridan metrics PR.

### DIFF
--- a/cmd/abc/abc.go
+++ b/cmd/abc/abc.go
@@ -47,8 +47,9 @@ const (
 	defaultLogFormat = logging.FormatText
 	// Long since only runs once every 24 hours.
 	updateTimeout = time.Second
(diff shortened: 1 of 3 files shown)


cmd/abc/abc.go
```diff
@@ -176,7 +177,9 @@ func realMain(ctx context.Context) error {
 
 	mClient, err := metrics.New(ctx, version.Name, version.Version)
 	if err != nil {
-		fmt.Printf("metric client creation failed: %v\n", err)
+		if err != nil {
```
sailorlqh, line 180: Is this if redundent?

pdewilde: oops, good catch.

sailorlqh (approved):

sailorlqh (approved):

